PR #242, opened 2025-12-09
feat: add early exit for concurrency

yaythomas:
*Issue #, if available:*

*Description of changes:*

- Add OrphanedChildException (BaseException) to terminate orphaned children when parent completes early
- Modify ThreadPoolExecutor to shutdown without waiting (wait=False) when completion criteria met
- Raise exception when orphaned children attempt to checkpoint, preventing subsequent operations from executing
- Update state.py to reject orphaned child checkpoints with exception instead of silent return
- Add comprehensive tests for early exit behavior and orphaned child handling

When min_successful or error threshold is reached in parallel/map operations, the parent now returns immediately without waiting for remaining branches to complete. Orphaned branches are terminated on their next checkpoint attempt, preventing wasted work and ensuring correct semantics for completion criteria.


PartiallyUntyped (commented):
How does this work with at most once semantics?

PR #335, opened 2026-04-02
feat: expand RetryPresets with fixed, linear, and slow strategies

### LINKED ISSUE #328: [Feature]: expand RetryPresets
### What would you like?

Expand RetryPresets to include more handy presets. 

To this in conjunction with JS to offer same features across SDKs:
https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/retry/retry-presets/retry-presets.ts

### Possible Implementation

- FixedWait
- LinearBackOff
- Slow

No

### Does this require an RFC?

No

### Additional Context

_No response_

AdityaAudi:
Adds linear and fixed retry strategies aligned with the JS SDK, as requested in #328.
                                                                                                                                                                
                                                                                                                                                                
* `create_linear_retry_strategy()`                                                                                                                                                                                                                                                                                
  * Top-level creator mirroring the JS SDK's `createLinearRetryStrategy`.                                                                                       
  * Uses a linearly increasing delay: `initial + increment * (attempts_made - 1)`.                                                                            
  * Defaults to `max_attempts=6` with retry delays of `1s, 2s, 3s, 4s, 5s`.                                                                                     
  * No jitter or `max_delay` cap; callers requiring those behaviors can compose a custom strategy via `create_retry_strategy()`.                              
* `RetryPresets.linear()`                                                                                                                                                                
  * Thin wrapper around `create_linear_retry_strategy()` with JS-aligned defaults.
* `RetryPresets.fixed(interval=None)`                                                                                                                                                                                                                                                                          
  * Uses a constant delay between retries (`5s` by default, no jitter).                                                                                         
  * Useful for polling-style workloads where consistent retry spacing is preferred.                                                                             
                                                                                                                                                                `create_retry_strategy()` continues to provide exponential backoff unchanged, resulting in the same retry-strategy creator shape as the JS SDK.                 
                                                                                                                                                                
## Test Plan                                    
                                                                                                                                                                
* [x] `hatch run dev-core:test` -> 1262 tests pass (0 regressions)                                                                                             
* [x] `hatch fmt --check` -> all lint/format checks pass                                                                                                         
* [x] `hatch run dev-core:typecheck` -> mypy clean, no issues in 63 source files
                                                                                                                                                                
Resolves #328

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/retries.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/retries.py
@@ -125,6 +125,34 @@ def create_retry_strategy(
     return retry_strategy
 
 
+def create_linear_retry_strategy(
+    max_attempts: int = 6,
+    initial_delay: Duration | None = None,
+    increment: Duration | None = None,
+) -> Callable[[Exception, int], RetryDecision]:
+    """Linearly increasing delay between retries: initial + increment * (attempts_made - 1).
+
+    Mirrors the JS SDK's ``createLinearRetryStrategy``. With the defaults this
+    yields delays of 1s, 2s, 3s, 4s, 5s. No jitter is applied and there is no
+    upper cap on the delay; callers who need either can build their own
+    strategy via ``create_retry_strategy``.
+    """
+    initial: Duration = (
+        initial_delay if initial_delay is not None else Duration.from_seconds(1)
+    )
+    step: Duration = increment if increment is not None else Duration.from_seconds(1)
+
+    def linear_retry_strategy(_error: Exception, attempts_made: int) -> RetryDecision:
+        if attempts_made >= max_attempts:
+            return RetryDecision.no_retry()
+        delay_seconds: int = initial.to_seconds() + step.to_seconds() * (
+            attempts_made - 1
+        )
+        return RetryDecision.retry(Duration(seconds=delay_seconds))
+
+    return linear_retry_strategy
+
+
 class RetryPresets:
     """Default retry presets."""
 
@@ -176,6 +204,31 @@ class RetryPresets:
                 initial_delay=Duration.from_seconds(1),
                 max_delay=Duration.from_minutes(1),
                 backoff_rate=1.5,
+                jitter_strategy=JitterStrategy.NONE,
+            )
+        )
+
+    @classmethod
+    def linear(cls) -> Callable[[Exception, int], RetryDecision]:
+        """Linearly increasing delay between retries: 1s, 2s, 3s, 4s, 5s."""
+        return create_linear_retry_strategy(
+            max_attempts=6,
+            initial_delay=Duration.from_seconds(1),
+            increment=Duration.from_seconds(1),
+        )
+
+    @classmethod
+    def fixed(
+        cls, interval: Duration | None = None
+    ) -> Callable[[Exception, int], RetryDecision]:
+        """Constant delay between retries (5s by default, no jitter)."""
+        delay: Duration = interval if interval is not None else Duration.from_seconds(5)
+        return create_retry_strategy(
+            RetryStrategyConfig(
+                max_attempts=5,
+                initial_delay=delay,
+                max_delay=delay,
+                backoff_rate=1,
                 jitter_strategy=JitterStrategy.NONE,
             )
         )
--- a/packages/aws-durable-execution-sdk-python/tests/retries_test.py
+++ b/packages/aws-durable-execution-sdk-python/tests/retries_test.py
@@ -11,6 +11,7 @@ from aws_durable_execution_sdk_python.retries import (
     RetryDecision,
     RetryPresets,
     RetryStrategyConfig,
+    create_linear_retry_strategy,
     create_retry_strategy,
 )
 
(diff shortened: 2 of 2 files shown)


AdityaAudi:
Just checking in. Happy to implement either approach once you've aligned on the direction. No rush!

src/aws_durable_execution_sdk_python/retries.py
```diff
@@ -172,3 +179,42 @@ def critical(cls) -> Callable[[Exception, int], RetryDecision]:
                 jitter_strategy=JitterStrategy.NONE,
             )
         )
+
+    @classmethod
+    def fixed_wait(cls) -> Callable[[Exception, int], RetryDecision]:
```
zhongkechen: Why don't these factory methods take parameters to customize the retry strategies, like `max_attempts` and etc? If they are just some constants, why don't we put them as constants like `retries.FIXED_WAIT`

AdityaAudi: @zhongkechen Good call-out on the pattern. The reason I kept it as parameterless classmethods is that default(), transient(), critical() already use that shape. I didn't want to introduce a second style just for the new ones. If a parameterized or constants-based API is the direction we want to go, I think it's worth doing as a follow-up that converts all of them together.

If I had to pick between the two: I'd lean toward parameterized factories over module-level constants, mostly because the RetryPresets. namespace is where users discover these via autocomplete. But genuinely happy to defer to your preference here.

AdityaAudi:
> Thanks for the review @yaythomas, and good callout @ParidelPooya. Apologies for the incorrect JS SDK parity claim.
> 
> I'd advocate for the enum approach here. Here's why:
> 
> **The sentinel pattern creates a maintenance trap.** `backoff_rate=0` meaning "linear" is a hidden contract if someone later adds validation like `backoff_rate must be > 0`, it silently breaks linear. It also means `backoff_rate` has two different semantics depending on its value, which makes the config harder to reason about.
> 
> **The codebase already has the pattern.** `JitterStrategy` is a `StrEnum` on `RetryStrategyConfig` that controls _how_ a calculation is applied. A `BackoffStrategy` enum would be the exact same pattern controls _how_ delay grows. Developers familiar with one will immediately understand the other.
> 
> **Discoverability compounds.** As @ParidelPooya noted, IDEs, LLMs, and docs all benefit from an explicit type. This matters more as the SDK grows, every future preset or custom config gets clearer for free.
> 
> The concrete proposal would be a lightweight StrEnum with calculation logic staying centralized in `create_retry_strategy`:
> 
> ```
> class BackoffStrategy(StrEnum):
>   """How retry delay grows between attempts."""
>   EXPONENTIAL = "EXPONENTIAL" # initial_delay * backoff_rate ^ (attempts - 1)
>   LINEAR = "LINEAR"      # initial_delay * attempts
>   FIXED = "FIXED"       # initial_delay (constant)
> ```
> 
> RetryStrategyConfig defaults to EXPONENTIAL, so fully backward-compatible. The fixed() and linear() presets work exactly as @yaythomas suggested just using the enum instead of a sentinel.
> 
> Thoughts? Happy to push whichever you land on.
> 
> > Worth noting: repurposing backoff_rate=0 for linear would also be a silent behavioral change for anyone currently using backoff_rate=0 (today it gives [5s, 1s, 1s, 1s, ...], the proposed sentinel would change it to [5s, 10s, 15s, ...]). An explicit enum avoids that ambiguity entirely.


@yaythomas @ParidelPooya Just checking in. Happy to implement either approach once you've aligned on the direction.

yaythomas:
Thank you for your patience @AdityaAudi, and for the really thoughtful design discussion! The points you raised about discoverability and the maintenance trap are spot on :-) 

The reference JS implementation has implemented linear retry as a dedicated helper rather than a backoff mode on the config, please see [createLinearRetryStrategy](https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/retry/linear-retry-strategy/linear-retry-strategy.ts).

So we'd like to take some of your ideas here and reshape the PR to align with that structure:

### linear creator
Add a linear creator mirroring JS (createLinearRetryStrategy(maxAttempts=6, initialDelay=1, increment=1)):

```python
def create_linear_retry_strategy(
    max_attempts: int = 6,
    initial_delay: Duration | None = None,
    increment: Duration | None = None,
) -> Callable[[Exception, int], RetryDecision]:
    """Linearly increasing delay: initial + increment * (attempts_made - 1)."""
    ...
```

with the additive formula initial + increment*(attempts_made - 1) → 1s, 2s, 3s, 4s, 5s.

One important note: the current PR's linear is initial * attempts (5, 10, 15…), which isn't the same as JS's initial + increment*(n-1) (1, 2, 3, 4, 5). The JS version has a separate increment knob, so we'd want to follow that shape.

### RetryPresets
Plus a thin RetryPresets.linear() that calls it with (6, 1s, 1s), matching JS's retryPresets.linear.

Fixed convenience: we can still offer this without the enum or a new creator, since fixed is just exponential with backoff_rate=1. It lives as a preset (this is essentially what was sketched earlier):

```python
@classmethod
def fixed(cls, interval: Duration | None = None) -> Callable[[Exception, int], RetryDecision]:
    """Constant delay between retries."""
    delay = interval or Duration.from_seconds(5)
    return create_retry_strategy(RetryStrategyConfig(
        max_attempts=5,
        initial_delay=delay,
        max_delay=delay,
        backoff_rate=1,
        jitter_strategy=JitterStrategy.NONE,
    ))
```

Since Exponential already lives in create_retry_strategy (≡ JS createRetryStrategy), we can drop the BackoffStrategy enum, the backoff_strategy field, linear_backoff, and slow, because with the linear helper we have the same two-creator shape as JS. slow is just exponential with longer delays, so it doesn't need a dedicated preset.

Last thing, sorry, since we started talking the repo has changed to a monorepo with a restructure and so the PR is currently conflicting, so it'll need a rebase onto main (paths moved under packages/aws-durable-execution-sdk-python/...).

Happy to help with this if anything's unclear, and thanks again for your great work! 🙇

AdityaAudi:
> Thank you for your patience @AdityaAudi, and for the really thoughtful design discussion! The points you raised about discoverability and the maintenance trap are spot on :-)
> 
> The reference JS implementation has implemented linear retry as a dedicated helper rather than a backoff mode on the config, please see [createLinearRetryStrategy](https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/retry/linear-retry-strategy/linear-retry-strategy.ts).
> 
> So we'd like to take some of your ideas here and reshape the PR to align with that structure:
> 
> ### linear creator
> Add a linear creator mirroring JS (createLinearRetryStrategy(maxAttempts=6, initialDelay=1, increment=1)):
> 
> ```python
> def create_linear_retry_strategy(
>     max_attempts: int = 6,
>     initial_delay: Duration | None = None,
>     increment: Duration | None = None,
> ) -> Callable[[Exception, int], RetryDecision]:
>     """Linearly increasing delay: initial + increment * (attempts_made - 1)."""
>     ...
> ```
> 
> with the additive formula initial + increment*(attempts_made - 1) → 1s, 2s, 3s, 4s, 5s.
> 
> One important note: the current PR's linear is initial * attempts (5, 10, 15…), which isn't the same as JS's initial + increment*(n-1) (1, 2, 3, 4, 5). The JS version has a separate increment knob, so we'd want to follow that shape.
> 
> ### RetryPresets
> Plus a thin RetryPresets.linear() that calls it with (6, 1s, 1s), matching JS's retryPresets.linear.
> 
> Fixed convenience: we can still offer this without the enum or a new creator, since fixed is just exponential with backoff_rate=1. It lives as a preset (this is essentially what was sketched earlier):
> 
> ```python
> @classmethod
> def fixed(cls, interval: Duration | None = None) -> Callable[[Exception, int], RetryDecision]:
>     """Constant delay between retries."""
>     delay = interval or Duration.from_seconds(5)
>     return create_retry_strategy(RetryStrategyConfig(
>         max_attempts=5,
>         initial_delay=delay,
>         max_delay=delay,
>         backoff_rate=1,
>         jitter_strategy=JitterStrategy.NONE,
>     ))
> ```
> 
> Since Exponential already lives in create_retry_strategy (≡ JS createRetryStrategy), we can drop the BackoffStrategy enum, the backoff_strategy field, linear_backoff, and slow, because with the linear helper we have the same two-creator shape as JS. slow is just exponential with longer delays, so it doesn't need a dedicated preset.
> 
> Last thing, sorry, since we started talking the repo has changed to a monorepo with a restructure and so the PR is currently conflicting, so it'll need a rebase onto main (paths moved under packages/aws-durable-execution-sdk-python/...).
> 
> Happy to help with this if anything's unclear, and thanks again for your great work! 🙇


Thanks @yaythomas, this works. create_linear_retry_strategy covers what I was after with the enum, so good to drop it. You're right the linear formula in the current PR is off too, the additive version matches JS. Will reshape along your sketch and rebase for the monorepo paths.              

Two things before pushing:
1. For create_linear_retry_strategy, planning JitterStrategy.FULL and 5min max_delay. Shall I use different defaults if JS does?                               
2. RetryPresets.fixed(interval=None) takes a param where the other presets don't. Assuming intentional, please flag if not.

yaythomas:
## 1. which defaults for linear
idea is to drop the extras (no jitter, no cap), in line with JS implementation, so max_attempts=6, initial_delay=1s, increment=1s. `RetryPresets.linear()` is just that with no extras.
```python
def create_linear_retry_strategy(
    max_attempts: int = 6,
    initial_delay: Duration | None = None,
    increment: Duration | None = None,
) -> Callable[[Exception, int], RetryDecision]:
    """Linearly increasing delay: initial + increment * (attempts_made - 1)."""
    initial: Duration = initial_delay if initial_delay is not None else Duration.from_seconds(1)
    step: Duration = increment if increment is not None else Duration.from_seconds(1)

    def linear_retry_strategy(_error: Exception, attempts_made: int) -> RetryDecision:
        if attempts_made >= max_attempts:
            return RetryDecision.no_retry()
        delay_seconds: int = initial.to_seconds() + step.to_seconds() * (attempts_made - 1)
        return RetryDecision.retry(Duration(seconds=delay_seconds))

    return linear_retry_strategy
```

## 2. param on fixed
yes it takes a param, and no the others don't. in this particular case, it's prob not that useful just to have a fixed with a arbitrary interval of 5s? As it is this is an extra convenience helper which the JS implementation doesn't have, so it might be that either we a) remove it altogether, or b) we keep it and introduce the arg as a convenience layered on top of the aligned base functionality, since it's effectively a "third" type of backoff in addition to exponential or linear?

By comparison,  to get fixed in JS (Without jitter), you'd need: `createRetryStrategy({ backoffRate: 1, jitter: JitterStrategy.NONE })` Just doing this `createRetryStrategy({ backoffRate: 1 })`, will also apply jitter to the fixed backoffrate.

In your initial PR you suggested a default of `(5s, no backoff, no jitter)`, so the quick sample I sketched out here with `fixed()` take its cues from you there :-)

since you're already done the work, I'm thinking b) we add in the fixed with the arg :-) What do you think, happy to go with your preference?

AdityaAudi:
@yaythomas  I've pushed the reshape and updated the description to match.                                                                                                        

Quick recap:                                                                                                                                                    
- added `create_linear_retry_strategy`, `RetryPresets.linear()`, `RetryPresets.fixed(interval=None)`                                                          
- dropped `BackoffStrategy`, `backoff_strategy`, `linear_backoff`, `slow`                                                                                       
- rebased onto main for the monorepo restructure                                                                                                              

Let me know if anything else needs tweaking.

yaythomas (approved):
:shipit: 🙇

Issue #345, opened 2026-04-15
[Feature]: add ci check for lint commits


yaythomas:
### What would you like?

the commit msg linting is not enabled on CI

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #348, opened 2026-04-17
ci: add commit message linting to ci workflow

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/345

*Description of changes:* 
- Added `--range` flag to `lintcommit.py` for CI to validate commits in a git revision range (e.g. `origin/main..<head sha>`)
- Added `lint-commits` CI job that runs on `pull_request` events
- Refactored commit parsing to use null char separator (`-z`) and `--no-merges` instead of magic string delimiter and merge skip logic
- Extracted `lint_range()` returning `LintResult`/`CommitResult` dataclasses so tests assert on returned data instead of stdout/stderr

### DIFF
--- a/ops/lintcommit.py
+++ b/ops/lintcommit.py
@@ -124,29 +124,33 @@ def validate_message(message: str) -> tuple[str | None, list[str]]:
     return (error, warnings)
 
 
-def run_local() -> None:
-    """Validate local commit messages ahead of origin/main.
+def run_range(git_range: str, *, skip_dirty_check: bool = False) -> None:
+    """Validate commit messages in a git range (e.g. 'origin/main..HEAD').
 
-    If there are uncommitted changes, prints a warning and skips validation.
+    Args:
+        git_range: A git revision range like 'origin/main..HEAD'.
+        skip_dirty_check: When True, skip the uncommitted changes check
+            (useful in CI where the worktree may be clean by definition).
     """
     import subprocess
 
-    # Check for uncommitted changes
-    status: subprocess.CompletedProcess[str] = subprocess.run(
-        ["git", "status", "--porcelain"],
-        capture_output=True,
-        text=True,
-    )
-    if status.stdout.strip():
-        print(
-            "WARNING: uncommitted changes detected, skipping commit message validation.\n"
-            "Commit your changes and re-run to validate."
+    if not skip_dirty_check:
+        # Check for uncommitted changes
+        status: subprocess.CompletedProcess[str] = subprocess.run(
+            ["git", "status", "--porcelain"],
+            capture_output=True,
+            text=True,
         )
-        return
+        if status.stdout.strip():
+            print(
+                "WARNING: uncommitted changes detected, skipping commit message validation.\n"
+                "Commit your changes and re-run to validate."
+            )
+            return
 
-    # Get all commit messages ahead of origin/main
+    # Get all commit messages in the range
     result: subprocess.CompletedProcess[str] = subprocess.run(
-        ["git", "log", "origin/main..HEAD", "--format=%H%n%B%n---END---"],
+        ["git", "log", git_range, "--format=%H%n%B%n---END---"],
         capture_output=True,
         text=True,
     )
@@ -156,7 +160,7 @@ def run_local() -> None:
 
     raw: str = result.stdout.strip()
     if not raw:
-        print("No local commits ahead of origin/main")
+        print(f"No commits in range {git_range}")
         return
 
     blocks: list[str] = raw.split("---END---")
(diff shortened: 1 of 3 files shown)


ayushiahjolia:
`lint-pr-title` job is skipped for this PR, because it was introduced in this PR. It will run on subsequent PRs.

ops/lintcommit.py
```diff
@@ -124,29 +124,33 @@ def validate_message(message: str) -> tuple[str | None, list[str]]:
...
+def run_range(git_range: str, *, skip_dirty_check: bool = False) -> None:
+    """Validate commit messages in a git range (e.g. 'origin/main..HEAD').
 
-    If there are uncommitted changes, prints a warning and skips validation.
+    Args:
+        git_range: A git revision range like 'origin/main..HEAD'.
+        skip_dirty_check: When True, skip the uncommitted changes check
+            (useful in CI where the worktree may be clean by definition).
     """
     import subprocess
```
yaythomas: avoid inline imports (import on top)

ops/tests/test_lintcommit.py
```diff
@@ -151,3 +157,117 @@ def test_empty_message() -> None:
...
+@patch("subprocess.run")
+def test_run_range_warnings_printed(mock_run, capsys) -> None:
+    log_output = _make_git_log_output(
+        "feat: add thing\n\n" + "x" * 80,
+    )
+    mock_run.return_value = _completed(stdout=log_output)
+
+    run_range("origin/main..HEAD", skip_dirty_check=True)
+
+    out = capsys.readouterr().out
```
yaythomas: this would test a lot easier if you refactor the logic into a function that returns an errs and/or warnings collection. 

then you dont have to capture stdout/stderr.

You could maybe have a separate `write_output(info: list, errs: list)` method?

ops/lintcommit.py
```diff
@@ -174,15 +178,21 @@ def run_local() -> None:
         if not message:
             continue
 
+        subject_line: str = message.splitlines()[0]
+
+        # Skip merge commits (auto-generated by git)
+        if subject_line.startswith("Merge "):
+            print(f"SKIP {sha}: {subject_line} (merge commit)")
+            continue
```
yaythomas: we don't want merge commits. 

so maybe just remove this section. 

gh has "Require linear history - Prevent merge commits from being pushed to matching refs." enabled..

can prob  pass --no-merges.

ops/lintcommit.py
```diff
@@ -124,29 +124,33 @@ def validate_message(message: str) -> tuple[str | None, list[str]]:
...
+                "WARNING: uncommitted changes detected, skipping commit message validation.\n"
+                "Commit your changes and re-run to validate."
+            )
+            return
 
-    # Get all commit messages ahead of origin/main
+    # Get all commit messages in the range
     result: subprocess.CompletedProcess[str] = subprocess.run(
-        ["git", "log", "origin/main..HEAD", "--format=%H%n%B%n---END---"],
+        ["git", "log", git_range, "--format=%H%n%B%n---END---"],
```
yaythomas: rather than magic string to split, use null char. (i.e it's possible a commit msg has a legit ---END--- in it)

```
 result = subprocess.run(
     ["git", "log", "--no-merges", git_range, "-z", "--format=%H%n%B"],
     capture_output=True, text=True,
 )
 if result.returncode != 0:
     print(f"git log failed: {result.stderr}", file=sys.stderr)
     sys.exit(1)

 commits: list[tuple[str, str]] = []
 for r in result.stdout.split("\0"):
     if not r.strip():
         continue
     sha, _, message = r.partition("\n")
     commits.append((sha, message.strip()))
```

yaythomas (changes requested):

ops/lintcommit.py
ayushiahjolia: yes that's better. Personally, I click update branch on github UI all the times, which adds a commit in history, and that fails the check.

ops/lintcommit.py
zhongkechen, line 7: The package name `ops` is misleading IMO, makes me think this is a SDK operation tool at runtime. We could put these github workflow scripts under `.github`

zhongkechen: And BTW, `ci-checks.sh` script is broken due to `lintcommit.py`

```
# commit message validation
python ops/lintcommit.py
```

This script should run in hatch's virtualenv.

ayushiahjolia: 1. Good point on the naming. I'd prefer to keep this as a separate change though, it's pre-existing - moving `ops/` would touch the multiple files.
2. lintcommit.py only uses stdlib so it runs fine with bare python - no hatch env needed. I just verified `ci-checks.sh` runs end-to-end successfully. What's broken?

ops/lintcommit.py
```diff
@@ -124,75 +127,146 @@ def validate_message(message: str) -> tuple[str | None, list[str]]:
...
+        dest="git_range",
+        help="Validate all commits in a git revision range (e.g. 'origin/main..HEAD'). "
+        "Skips the uncommitted-changes check (useful in CI).",
+    )
+    args = parser.parse_args()
+
+    if args.git_range is not None:
+        run_range(args.git_range, skip_dirty_check=True)
+    else:
+        run_local()
```
yaythomas: nit: the naming here doesn't make a lot of sense anymore. couple of refactors down the line, why run_local? is there a run_server? why does run_local_ just to run_range?

suggestions: just run_range, run_local can go away

yaythomas (changes requested):

ops/lintcommit.py
ayushiahjolia: It was mainly to differentiate between running locally and running against PR commits.

ops/lintcommit.py
zhongkechen: I don't have a valid `python` command out of the hatch env.

ayushiahjolia:
Build failing due to unrelated parallel unit tests, will debug.

ayushiahjolia:
Failing test should be fixed by https://github.com/aws/aws-durable-execution-sdk-python/pull/367

ops/lintcommit.py
yaythomas: maybe update comment to include `ops/tests/test_lintcommit.py`

if you're doing a rename subsequent of ops, can do it then.

yaythomas (approved):

PR #351, opened 2026-04-22
[feature]: add nesting_type to concurrency operations

### LINKED ISSUE #350: [Feature]: add nesting option to Map/Parallel
### What would you like?

A way to config Ma/Parallel to create nesting for each iteration/branch or keep it flat and lower the cost and increase max iterations possible

Similar to
- https://github.com/aws/aws-durable-execution-sdk-js/pull/456
- https://github.com/aws/aws-durable-execution-sdk-java/pull/335

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

- fixes: https://github.com/aws/aws-durable-execution-sdk-python/issues/350

*Description of changes:*

- added a new nesting mode parameter to map and parallel operation
- added a new checkpoint mode parameter to child context operation

```
============================= test session starts ==============================
collecting ... collected 1 item

examples/test/map/test_map_operations_flat.py::test_map_operations 

========================= 1 passed, 1 warning in 0.55s =========================
PASSED [100%]

============================= test session starts ==============================
collecting ... collected 1 item

examples/test/parallel/test_parallel_flat.py::test_parallel_flat 

======================== 1 passed, 3 warnings in 1.55s =========================
PASSED  [100%]
```

TESTING_SDK_BRANCH=nest_type

Fixed integration workflow.

```
examples/test/parallel/test_parallel_flat.py::test_parallel_flat PASSED  [ 97%]
```

### DIFF
--- a/src/aws_durable_execution_sdk_python/concurrency/executor.py
+++ b/src/aws_durable_execution_sdk_python/concurrency/executor.py
@@ -20,7 +20,10 @@ from aws_durable_execution_sdk_python.concurrency.models import (
     ExecutionCounters,
     SuspendResult,
 )
-from aws_durable_execution_sdk_python.config import ChildConfig
+from aws_durable_execution_sdk_python.config import (
+    ChildConfig,
+    NestingType,
+)
 from aws_durable_execution_sdk_python.exceptions import (
     OrphanedChildException,
     SuspendExecution,
@@ -134,6 +137,7 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
 
     def __init__(
         self,
+        operation_identifier: OperationIdentifier,
         executables: list[Executable[CallableType]],
         max_concurrency: int | None,
         completion_config: CompletionConfig,
@@ -143,6 +147,7 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
         serdes: SerDes | None,
         item_serdes: SerDes | None = None,
         summary_generator: SummaryGenerator | None = None,
+        nesting_type: NestingType = NestingType.NESTED,
     ):
         """Initialize ConcurrentExecutor.
 
@@ -153,6 +158,7 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
                 handle large BatchResult payloads efficiently. Matches TypeScript behavior in
                 run-in-child-context-handler.ts.
         """
+        self.operation_identifier = operation_identifier
         self.executables = executables
         self.max_concurrency = max_concurrency
         self.completion_config = completion_config
@@ -160,6 +166,7 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
         self.sub_type_iteration = sub_type_iteration
         self.name_prefix = name_prefix
         self.summary_generator = summary_generator
+        self.nesting_type = nesting_type
 
         # Event-driven state tracking for when the executor is done
         self._completion_event = threading.Event()
@@ -406,7 +413,14 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
             executable.index
         )
         name = f"{self.name_prefix}{executable.index}"
-        child_context = executor_context.create_child_context(operation_id)
+        non_virtual_parent_id = (
+            self.operation_identifier.operation_id
+            if self.nesting_type is NestingType.FLAT
+            else None
+        )
+        child_context = executor_context.create_child_context(
+            operation_id, non_virtual_parent_id
+        )
         operation_identifier = OperationIdentifier(
             operation_id,
             executor_context._parent_id,  # noqa: SLF001
@@ -424,6 +438,7 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
                 serdes=self.item_serdes or self.serdes,
                 sub_type=self.sub_type_iteration,
                 summary_generator=self.summary_generator,
+                is_virtual=self.nesting_type is NestingType.FLAT,
             ),
         )
         child_context.state.track_replay(operation_id=operation_id)
--- a/src/aws_durable_execution_sdk_python/config.py
+++ b/src/aws_durable_execution_sdk_python/config.py
@@ -76,6 +76,18 @@ class TerminationMode(Enum):
     ABANDON = "ABANDON"
 
 
+class NestingType(Enum):
+    """
+    How child operations should inherit context from their parent.
+
+    - NESTED: Each child operation runs in its own isolated context
+    - FLAT: All child operations share the same parent context
+    """
+
+    NESTED = "NESTED"
+    FLAT = "FLAT"
+
+
 @dataclass(frozen=True)
 class CompletionConfig:
     """Configuration for determining when parallel/map operations complete.
@@ -187,6 +199,10 @@ class ParallelConfig:
             Used internally by map/parallel operations to handle large BatchResult payloads.
             Signature: (result: T) -> str
 
+        nesting_type: How child operations should inherit context from their parent.
+            - NESTED: Each branch runs in its own isolated context (default)
+            - FLAT: All branches share the same parent context
+
     Example:
         # Run at most 3 branches concurrently, succeed if any one succeeds
         config = ParallelConfig(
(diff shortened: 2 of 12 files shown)


src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -160,6 +165,7 @@ def __init__(
         self.sub_type_iteration = sub_type_iteration
         self.name_prefix = name_prefix
         self.summary_generator = summary_generator
+        self.nesting_type = nesting_type
```
yaythomas, line 169: strictly speaking it's not a _Type_ in the sense of a https://docs.python.org/3/library/stdtypes.html. 

suggestion: align w reference typescript

```
 nesting?: NestingType;
```

i.e

```
self.nesting: NestingType = nesting
```

src/aws_durable_execution_sdk_python/operation/child.py
```diff
@@ -118,7 +118,11 @@ def check_result_status(self) -> CheckResult[T]:
             checkpointed_result.raise_callable_error()
 
         # Create START checkpoint if not exists
-        if not checkpointed_result.is_existent():
+        if (
+            not checkpointed_result.is_existent()
+            and self.config.checkpoint_mode
+            == CheckpointMode.CHECKPOINT_AT_START_AND_FINISH
```
yaythomas: enum checks are `is` rather than `==`

src/aws_durable_execution_sdk_python/operation/child.py
```diff
@@ -203,18 +207,21 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
...
-                payload=serialized_result,
-                sub_type=self.sub_type,
-                context_options=ContextOptions(replay_children=replay_children),
-            )
-            # Checkpoint child context SUCCEED with blocking (is_sync=True, default).
-            # Must ensure the child context result is persisted before returning to the parent.
-            # This guarantees the result is durable and child operations won't be re-executed on replay
-            # (unless replay_children=True for large payloads).
-            self.state.create_checkpoint(operation_update=success_operation)
+            if self.config.checkpoint_mode != CheckpointMode.NO_CHECKPOINT:
```
yaythomas: `is not`

src/aws_durable_execution_sdk_python/operation/map.py
```diff
@@ -57,6 +58,7 @@ def __init__(
             serdes=serdes,
             summary_generator=summary_generator,
             item_serdes=item_serdes,
+            nesting_type=nesting_type,
```
yaythomas, line 63: ditto 

`nesting` rather than `nesting_type`

src/aws_durable_execution_sdk_python/config.py
```diff
@@ -76,6 +76,18 @@ class TerminationMode(Enum):
     ABANDON = "ABANDON"
 
 
+class NestingType(Enum):
+    """
+    Defines how child operations inherit context from their parent.
```
yaythomas: imperative mood and active voice.

```
Define how child operations inherit context from their parent.

       Attributes:
           NESTED: Child operations run in their own isolated nested context.
           FLAT: Child operations run within the same context as their parent.
```

src/aws_durable_execution_sdk_python/config.py
```diff
@@ -259,17 +275,16 @@ class ChildConfig(Generic[T]):
 
             Used internally by map/parallel operations to handle large BatchResult payloads.
             Signature: (result: T) -> str
-    Note:
-        checkpoint_mode field is commented out as it's not currently implemented.
-        When implemented, it will control when checkpoints are created:
+
+        checkpoint_mode: controls when checkpoints are created
```
yaythomas: "Control when to checkpoint during operation execution."

but if this becomes 

```
   context_mode: How this child context appears in the operation hierarchy.
   - STANDARD: Checkpoint at start and completion. Child operations appear under
   this
           context in the hierarchy (default).
   - TRANSPARENT: Skip context checkpointing. Child operations appear under the
   parent
           context in the hierarchy.
```

src/aws_durable_execution_sdk_python/concurrency/executor.py
yaythomas: shouldn't this use the grandparent id?

```
child_parent_id = (
       executor_context._parent_id  # grandparent
       if self.nesting_type is NestingType.FLAT
       else operation_id
   )
```

and then in run_in_child_context something like?

```
   operation_id = self._create_step_id()

   child_parent_id = self._parent_id if (config and config.checkpoint_mode is
    CheckpointMode.NO_CHECKPOINT) else operation_id

   def callable_with_child_context():
       return func(self.create_child_context(parent_id=child_parent_id))
```

src/aws_durable_execution_sdk_python/config.py
```diff
@@ -259,17 +275,16 @@ class ChildConfig(Generic[T]):
...
         - CHECKPOINT_AT_START_AND_FINISH: Checkpoint at both start and completion (default)
-        - CHECKPOINT_AT_FINISH: Only checkpoint when operation completes
+        - CHECKPOINT_AT_FINISH: Only checkpoint when operation completes (not implemented)
```
yaythomas: removed

src/aws_durable_execution_sdk_python/config.py
```diff
@@ -259,17 +275,16 @@ class ChildConfig(Generic[T]):
...
         - NO_CHECKPOINT: No automatic checkpointing
 
     See TypeScript reference: aws-durable-execution-sdk-js/src/types/index.ts
     """
 
-    # checkpoint_mode: CheckpointMode = CheckpointMode.CHECKPOINT_AT_START_AND_FINISH
+    checkpoint_mode: CheckpointMode = CheckpointMode.CHECKPOINT_AT_START_AND_FINISH
```
yaythomas: this doesn't really accurately describe the effect of this enum. 

The enum influences which parent id child operations will get. 

how about

```
ContextMode:
    STANDARD
    VIRTUAL

context_mode: ContextMode
```


and CheckpointMode can go away entirely.

yaythomas (changes requested):

src/aws_durable_execution_sdk_python/concurrency/executor.py
zhongkechen: `nesting`? Java is using `nestingType`. And the name here is internal to the SDK, which doesn't impact how users use the feature.

src/aws_durable_execution_sdk_python/concurrency/executor.py
zhongkechen: No, the operation id is still generated the same way as a real child context, so we still need the parent. But I think we need to update the operation's `parentId` to be its grand parent when sending the updates to backend. I will update that later.

src/aws_durable_execution_sdk_python/operation/child.py
zhongkechen: Both work. Do we prefer `is` over `==`? `==` wouldn't work in some scenarios using `StrEnum`/`IntEnum`. Mixing both styles could be confusing and  may lead to bugs.

src/aws_durable_execution_sdk_python/config.py
zhongkechen: Will remove CheckpointMode entirely

src/aws_durable_execution_sdk_python/config.py
zhongkechen: Will remove CheckpointMode entirely

src/aws_durable_execution_sdk_python/config.py
zhongkechen: Will remove CheckpointMode entirely

src/aws_durable_execution_sdk_python/operation/child.py
yaythomas: Use `is` for all internal state checks where you are comparing an Enum member against another member. 

Use == only when you specifically need to compare an Enum member against a raw data type (like a string from a JSON response)

yaythomas (approved):

PR #358, opened 2026-04-28
feat: Classify non-retryable customer errors for Durable API calls

denzyll:
*Issue #, if available:*

N/A

*Description of changes:*

Non-retryable customer errors from Lambda (e.g., KMSAccessDeniedException, KMSDisabledException) arrive as HTTP 502 during CheckpointDurableExecution and GetDurableExecutionState API calls. Without this change, these 502s are classified as retryable invocation errors, causing the SDK to retry invocations that will never self-resolve.

Extract shared classification logic into BotoClientError with a _NON_RETRYABLE_CUSTOMER_ERROR_CODES set and _classify_error_category method. Both CheckpointError and GetExecutionStateError now inherit from_exception() with unified classification so that non-retryable errors (KMS 502s, 4xx client errors) return Status: FAILED immediately, while retryable errors (5xx, 429, network errors) continue to raise for Lambda retry.

Add is_retryable() to InvocationError hierarchy so execution.py handlers use a single interface instead of isinstance checks.

Add backward-compatible aliases for CheckpointErrorCategory and is_retriable().

Testing — parameterized classification tests across both error types for all four KMS codes, 4xx, 429, 5xx, and 502 scenarios. Integration tests for non-retryable/retryable errors across all three execution.py code paths: initial pagination, background thread, and user thread.

### DIFF
--- a/src/aws_durable_execution_sdk_python/exceptions.py
+++ b/src/aws_durable_execution_sdk_python/exceptions.py
@@ -146,9 +146,14 @@ class BotoClientError(InvocationError):
         response = getattr(exception, "response", {})
         response_metadata = response.get("ResponseMetadata")
         error = response.get("Error")
-        error_category = BotoClientError._classify_error_category(error, response_metadata)
+        error_category = BotoClientError._classify_error_category(
+            error, response_metadata
+        )
         return cls(
-            message=str(exception), error_category=error_category, error=error, response_metadata=response_metadata
+            message=str(exception),
+            error_category=error_category,
+            error=error,
+            response_metadata=response_metadata,
         )
 
     @staticmethod
@@ -171,7 +176,9 @@ class BotoClientError(InvocationError):
         if error_code and error_code in _NON_RETRYABLE_CUSTOMER_ERROR_CODES:
             return DurableApiErrorCategory.EXECUTION
 
-        status_code: int | None = (response_metadata and response_metadata.get("HTTPStatusCode")) or None
+        status_code: int | None = (
+            response_metadata and response_metadata.get("HTTPStatusCode")
+        ) or None
         if (
             status_code
             and BAD_REQUEST_ERROR <= status_code < SERVICE_ERROR
@@ -179,7 +186,9 @@ class BotoClientError(InvocationError):
             and error
             and not (
                 (error.get("Code") or "") == INVALID_PARAMETER_VALUE_EXCEPTION
-                and (error.get("Message") or "").startswith(INVALID_CHECKPOINT_TOKEN_PREFIX)
+                and (error.get("Message") or "").startswith(
+                    INVALID_CHECKPOINT_TOKEN_PREFIX
+                )
             )
         ):
             return DurableApiErrorCategory.EXECUTION
--- a/src/aws_durable_execution_sdk_python/exceptions.py
+++ b/src/aws_durable_execution_sdk_python/exceptions.py
@@ -13,6 +13,22 @@ from typing import TYPE_CHECKING, Self, TypedDict
 BAD_REQUEST_ERROR: int = 400
 TOO_MANY_REQUESTS_ERROR: int = 429
 SERVICE_ERROR: int = 500
+INVALID_PARAMETER_VALUE_EXCEPTION: str = "InvalidParameterValueException"
+INVALID_CHECKPOINT_TOKEN_PREFIX: str = "Invalid Checkpoint Token"
+
+# Non-retriable customer error codes that arrive as non-4xx (e.g. HTTP 502) from Lambda.
+# Unlike typical 5xx errors, these require customer intervention (e.g., fixing
+# a KMS key configuration) and will never succeed on retry.
+# Add new non-retriable error codes here — they are automatically classified
+# as EXECUTION (non-retriable) by _classify_error_category().
+_NON_RETRIABLE_CUSTOMER_ERROR_CODES: frozenset[str] = frozenset(
+    {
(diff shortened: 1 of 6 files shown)


src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -86,27 +110,98 @@ def __init__(self, message: str, callback_id: str | None = None):
...
+CheckpointErrorCategory = DurableApiErrorCategory
+
+
 class BotoClientError(InvocationError):
+    """Error from a Lambda API call (e.g., CheckpointDurableExecution, GetDurableExecutionState).
+
+    Extends InvocationError because the default behavior for API failures is to retry
+    the Lambda invocation. However, some errors are non-retryable (e.g., 4xx client errors,
+    KMS key misconfiguration) and should fail the execution instead. The error_category field
+    and is_retryable() method distinguish these cases at runtime.
```
zhongkechen, line 128: This seems confusing. BotoClientError is inherited from InvocationError which means the exception will fail only the invocation but actually it will fail the execution sometimes. So I think whether an exception such as BotoClientError will fail the execution or retry the invocation should be determined purely by its retryable method at runtime, instead of its class hierarchy at compile time.

src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -86,27 +110,98 @@ def __init__(self, message: str, callback_id: str | None = None):
         self.callback_id = callback_id
 
 
+class DurableApiErrorCategory(Enum):
+    INVOCATION = "INVOCATION"
+    EXECUTION = "EXECUTION"
```
zhongkechen, line 115: Why is this category needed? Doesn't `retryable` work for the same purpose already? retryable=true means `INVOCATION` and retryable=false means `EXECUTION`. And we also have two base exception classes `InvocationError` and `ExecutionError`, which makes this even more confusing.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -303,11 +303,26 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
-            invocation_input.checkpoint_token,
-            invocation_input.initial_execution_state.next_marker,
-        )
+        try:
+            execution_state.fetch_paginated_operations(
+                invocation_input.initial_execution_state.operations,
+                invocation_input.checkpoint_token,
+                invocation_input.initial_execution_state.next_marker,
+            )
+        except BotoClientError as e:
```
zhongkechen, line 269: shouldn't this error handling logic be in the method `fetch_paginated_operations`? `fetch_paginated_operations` is called in two places

src/aws_durable_execution_sdk_python/execution.py
zhongkechen, line 366: Just curious, why do we wrap checkpoint error with this error type `BackgroundThreadError`?

src/aws_durable_execution_sdk_python/exceptions.py
zhongkechen: I would suggest create a new exception type under `UnrecoverableError`

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -303,11 +303,26 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
             else ReplayStatus.NEW,
         )
 
-        execution_state.fetch_paginated_operations(
-            invocation_input.initial_execution_state.operations,
-            invocation_input.checkpoint_token,
-            invocation_input.initial_execution_state.next_marker,
-        )
+        try:
+            execution_state.fetch_paginated_operations(
```
zhongkechen, line 264: I just moved this call to another class `ExecutionState`. This PR has to be updated accordingly.

src/aws_durable_execution_sdk_python/exceptions.py
zhongkechen: Making this change would bring in a breaking change. Let's keep it like this for now.

src/aws_durable_execution_sdk_python/exceptions.py
zhongkechen: This enum already existed before this change. Let's keep this for now.

src/aws_durable_execution_sdk_python/execution.py
denzyll: Updated to account for this.

src/aws_durable_execution_sdk_python/execution.py
denzyll: All decisions to fail the execution or retry the invocation are made in the execution file.

src/aws_durable_execution_sdk_python/execution.py
denzyll: We use it to make sure the error is propagated from the background checkpointing thread to the handler thread in execution.py for error handling.

```
# Background checkpointing errors will propagate through CompletionEvent.wait() as BackgroundThreadError
```

denzyll: Also, the `BackgroundThreadError` docstring has more details: https://github.com/aws/aws-durable-execution-sdk-python/blob/61c403a47a988dc3daf02f424c6c2614f0c37740/src/aws_durable_execution_sdk_python/exceptions.py#L233

It says that `BackgroundThreadError` inherits from `BaseException` to bypass normal exception handlers.

zhongkechen (approved):

PR #364, opened 2026-04-29
Virtual Context bug on parent id and FAIL checkpoint emitting

### LINKED ISSUE #362: Virtual child contexts emit FAIL checkpoints in FLAT mode
**Summary**

When a virtual child context (a FLAT-mode `map` or `parallel` branch) fails, the Python SDK writes a `FAIL` checkpoint for the branch. The SDK does not write corresponding `START` or `SUCCEED` checkpoints for virtual branches.

**Expected behavior**

A virtual branch should emit no lifecycle checkpoints at all, regardless of outcome. This matches the JS reference implementation, which suppresses START, SUCCEED, and FAIL writes uniformly when `isVirtual === true`:

`aws/aws-durable-execution-sdk-js`, `src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts`:
- START gated on `!isVirtual` at line 287
- SUCCEED gated on `!isVirtual` at line 362
- FAIL gated on `!isVirtual` at line 401

Link: https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/handlers/run-in-child-context-handler/run-in-child-context-handler.ts#L401

**Actual behavior (current Python, `main`)**

`src/aws_durable_execution_sdk_python/operation/child.py`, `ChildOperationExecutor.execute`: the exception handler writes a `create_context_fail` update unconditionally. The `self.config.is_virtual` check that guards START (in `check_result_status`) and SUCCEED (in the success path of `execute`) is not applied to the FAIL write. When a FLAT branch fails, the Python SDK sends a CONTEXT/FAIL `OperationUpdate` for an operation id that was never started.

**Reproduction**

Run a `map` or `parallel` with `nesting_type=NestingType.FLAT` where at least one branch raises an exception. Inspect execution history — a FAILED CONTEXT operation appears for each failed branch, with `startedTimestamp == closedTimestamp`. In the DurableExecutionsWorkerService metrics, `numBillableOperations` will have incremented for each of these phantom entries.

**Fix**

Gate the FAIL checkpoint on the same `is_virtual` condition that guards START and SUCCEED, so all three lifecycle entries are uniformly suppressed for virtual contexts. The exception must still propagate to the concurrency executor so it can record the branch's failure in the `BatchResult` and apply its completion-tolerance logic.

### LINKED ISSUE #363: virtual child context orphaned parent_ids in execution history
## Summary

When `run_in_child_context(fn, config=ChildConfig(is_virtual=True))` is called, operations executed inside `fn` stamp a `parent_id` that refers to the virtual branch's own operation id. The branch's own START/SUCCEED checkpoints are suppressed by `is_virtual=True`, so that id is never written to the execution history.

Any subsequent checkpoint call for an inner operation therefore carries a `ParentId` that the backend cannot validate against a known CONTEXT operation. The backend correctly rejects this with `InvalidParameterValueException(CHECKPOINT_INVALID_PARENT_OPERATION_ID)`. The SDK should never have produced the invalid `parent_id` in the first place.

`NestingType.FLAT` on `map` and `parallel` is **not** affected. The concurrency executor takes a different code path that explicitly propagates the map/parallel operation id as `non_virtual_parent_id`, so FLAT branches correctly stamp a real, checkpointed parent.

The bug compounds with nesting: `run_in_child_context(is_virtual=True)` inside another `run_in_child_context(is_virtual=True)` produces inner operations whose `parent_id` chain points through one or more virtual branches, none of which are checkpointed. The failure mode is the same (invalid parent id), just at varying depth.

## Impact

- Any code that sets `ChildConfig(is_virtual=True)` and passes it to `run_in_child_context`.
- The first inner operation attempting to checkpoint will fail with a backend validation error.
- Execution does not complete; the user's function surfaces a checkpoint-rejection error rather than the result they expected.

## Root cause

`run_in_child_context` (on `origin/main`, `src/aws_durable_execution_sdk_python/context.py` around line 498) builds the child context via:

```python
def callable_with_child_context():
    return func(self.create_child_context(parent_id=operation_id))
```

It passes only `parent_id` to `create_child_context`, never `non_virtual_parent_id`. The child context's `__init__` then falls back to:

```python
self._non_virtual_parent_id = non_virtual_parent_id or parent_id
```

which becomes `operation_id` — the branch's own id. Every inner operation stamps `self._non_virtual_parent_id`, so they all point at the branch.

Meanwhile, `run_in_child_context` passes the user's `ChildConfig(is_virtual=True)` straight through to `child_handler`, which correctly suppresses START/SUCCEED for the branch. The two decisions are decoupled: `child_handler` knows the branch is virtual and must not checkpoint, but `create_child_context` doesn't, so inner operations still report to a virtual (never-checkpointed) branch id. The SDK then tries to submit a checkpoint referencing that id as a parent, which the backend rejects — correctly.

For contrast, the `ConcurrentExecutor._execute_item_in_child_context` path (used by `map`/`parallel` FLAT) explicitly computes and passes `non_virtual_parent_id` so that FLAT branches report to the enclosing map/parallel op. That's why `NestingType.FLAT` on `map` and `parallel` works today.

## Suggested fix direction

Keep the two decisions coupled. Either:

- Make `create_child_context` aware of virtualness (e.g. `create_child_context(operation_id, is_virtual=False)`) and derive the reporting parent internally, so callers can't forget, or
- Have `run_in_child_context` pass `non_virtual_parent_id=self._non_virtual_parent_id` when `config.is_virtual`, matching what the concurrency executor does.

The first option additionally handles nested virtuals consistently. The JS SDK already does the equivalent, passing the calling context's `parentId` into a virtual child.

## Related

A secondary issue — virtual branches still writing a FAIL checkpoint despite suppressed START/SUCCEED — is tracked at #362. That FAIL is itself an orphan (no matching START), and would similarly be rejected by backend validation if submitted.

## Environment

- Repo: `aws/aws-durable-execution-sdk-python`
- Branch/commit: `origin/main` at `61c403a`

yaythomas:
*Description of changes:*

Fixes two functional bugs in virtual child contexts and refactors the supporting mechanism.

**#363 — `run_in_child_context(is_virtual=True)` sends invalid `parent_id` to the backend.** On `main`, inner operations inside a virtual `run_in_child_context` stamp the branch's own operation id as their `parent_id`. The branch itself writes no START/SUCCEED under `is_virtual=True`, so that id does not exist in the execution history. The backend correctly rejects the checkpoint with `InvalidParameterValueException(CHECKPOINT_INVALID_PARENT_OPERATION_ID)`, and the user's execution fails at the first inner checkpoint. Nesting virtual scopes (and FLAT `map`/`parallel` inside a virtual scope) compounds the failure. After this change, inner operations stamp the enclosing non-virtual ancestor, which is a real checkpointed parent.

**#362 — virtual branches wrote a FAIL checkpoint on exceptions.** On `main`, an exception inside a virtual branch caused `ChildOperationExecutor.execute` to unconditionally write a FAIL checkpoint for the branch — producing a phantom FAILED CONTEXT entry in the execution history with no matching START. On the FLAT `map`/`parallel` path this was cosmetic (the FAIL's parent id was the valid map/parallel op, the backend accepts a CONTEXT FAIL without a prior START); on the `run_in_child_context(is_virtual=True)` path it compounded #363 (the FAIL's parent id was the same orphan branch id, which the backend rejects). Either way it's incorrect: virtual branches should emit no lifecycle entries at all. After this change, virtual branches emit no lifecycle checkpoints; failures still propagate to the concurrency executor's `BatchResult` (or to the caller of `run_in_child_context`), and completion-tolerance logic still applies.

**Mechanism changes (applied alongside the fixes):**

- Single source of truth for the virtual-vs-real decision. `create_child_context` computes two fields (`_parent_id`, `_step_id_prefix`) at construction; no per-operation-method knowledge is required. New operations just read `self._parent_id` and work correctly under both modes.
- Field names match their roles. `_parent_id` is "the id my inner operations stamp as their parent"; `_step_id_prefix` is "how I prefix step ids". Each field has one job.
- `is_virtual` is encapsulated in the context as a cached property. Callers opt in with `create_child_context(..., is_virtual=True)`; the property makes the state inspectable.
- Nested virtual-in-virtual matches the JS reference. A virtual child of a virtual parent inherits its parent's reporting ancestor, so chained FLAT layers collapse to the outermost non-virtual context without dangling parent-id references.
- `ChildConfig.is_virtual` drives lifecycle-checkpoint suppression in `ChildOperationExecutor` (START, SUCCEED, FAIL) and, via `run_in_child_context`, the child context's own virtual-ness. The field remains public, matching the JS SDK's `ChildConfig.virtualContext`.
- Fewer parameters threaded through. `operation_identifier` is gone from `ConcurrentExecutor`, `MapExecutor`, and `ParallelExecutor` constructors and `from_items`/`from_callables`; the concurrency layer no longer needs an `OperationIdentifier` to figure out the reporting parent.
- Tests exercise the invariants directly with a real `DurableContext` and assert wire-format decisions, including nested-virtual-in-virtual coverage.

**Impact:**

- Users can call `ctx.run_in_child_context(config=ChildConfig(is_virtual=True))` without their execution being rejected at checkpoint time.
- FLAT `map`/`parallel` no longer emits phantom FAILED CONTEXT entries on branch failure.
- Python SDK virtual-context semantics now match the JS reference SDK exactly (same parent-id propagation rule; nested virtuals collapse identically).
- Cost: eliminates billable FAILED CONTEXT entries that were previously emitted for failed virtual branches.

Also includes minor CONTRIBUTING.md improvement (hatch venv symlink tip for editors that mangle paths with spaces).

*Issue #, if available:*

Fixes #362, Fixes #363

### DIFF
--- a/src/aws_durable_execution_sdk_python/config.py
+++ b/src/aws_durable_execution_sdk_python/config.py
@@ -77,15 +77,39 @@ class TerminationMode(Enum):
 
 
 class NestingType(Enum):
-    """
-    How child operations should inherit context from their parent.
+    """Control how child contexts are created for batch operations.
 
-    - NESTED: Each child operation runs in its own isolated context
-    - FLAT: All child operations share the same parent context
+    Applies to `map` and `parallel`. Each branch or iteration runs inside a
+    child context.
+
+        - NESTED: full checkpointed context
+        - FLAT: a virtual context that skips checkpoints for the branch/iteration.
+
     """
 
     NESTED = "NESTED"
+    """Create CONTEXT operations for each branch/iteration with full checkpointing.
+
+    Operations within each branch/iteration are wrapped in their own context.
+
+    - Observability: high — each branch/iteration appears as a separate
+      operation in execution history.
+    - Cost: higher — consumes more operations due to CONTEXT creation
+      overhead.
+    - Scale: lower maximum iterations due to operation limits.
+    """
+
     FLAT = "FLAT"
+    """Skip CONTEXT operations for branches/iterations using virtual contexts.
+
+    Operations execute directly without individual context wrapping.
+
+    - Observability: lower — branches/iterations don't appear as separate
+      operations in execution history.
+    - Cost: ~30% lower — reduces operation consumption by skipping CONTEXT
+      overhead.
+    - Scale: higher maximum iterations possible within operation limits.
+    """
 
 
 @dataclass(frozen=True)
@@ -271,9 +295,13 @@ class ChildConfig(Generic[T]):
             Used internally by map/parallel operations to handle large BatchResult payloads.
             Signature: (result: T) -> str
 
-        is_virtual: Whether the child operation is virtual (doesn't represent a real operation).
-            Virtual contexts are used for concurrency operations and don't appear in
-            the final execution history. Default is False.
+        is_virtual: When True, skip all checkpoints (START, SUCCEED,
+            FAIL) for this child context and propagate the caller's reporting
+            parent id through to operations created inside the child. The
+            branch is a logical scope for step-id prefixing but does not
+            appear in the execution history. Used internally by
+            NestingType.FLAT branches. Use this to group operations without
+            adding a CONTEXT entry to the execution history.
 
     See TypeScript reference: aws-durable-execution-sdk-js/src/types/index.ts
     """
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,9 +139,24 @@ echo "$(hatch env find)/bin/python"
 ```
 
 ### VS Code
+#### Interpreter
 If you're using VS Code, "Python: Select Interpreter" and use the hatch venv Python interpreter
 as found with the `hatch env find` command.
 
+Kiro and VS Code mangles the interpreter path if it contains spaces, which results in
+errors finding the interpreter. You can create a local .venv file symlink _without_ spaces
+in the path:
+
+```bash
+# create a symlink at: ./.venv/bin/python
+rm -rf .venv && ln -s "$(hatch env find)" .venv
+```
+
+When you "Select Interpreter", enter path `./.venv/bin/python`.
+
+You'll have to rerun this command whenever you recreate your hatch envs.
+
+#### Linting
 Hatch uses Ruff for static analysis.
 
 You might want to install the [Ruff extension for VS Code](https://github.com/astral-sh/ruff-vscode)
--- a/src/aws_durable_execution_sdk_python/concurrency/executor.py
+++ b/src/aws_durable_execution_sdk_python/concurrency/executor.py
@@ -137,7 +137,6 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
 
     def __init__(
         self,
-        operation_identifier: OperationIdentifier,
         executables: list[Executable[CallableType]],
         max_concurrency: int | None,
         completion_config: CompletionConfig,
@@ -158,7 +157,6 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
                 handle large BatchResult payloads efficiently. Matches TypeScript behavior in
                 run-in-child-context-handler.ts.
         """
-        self.operation_identifier = operation_identifier
         self.executables = executables
         self.max_concurrency = max_concurrency
         self.completion_config = completion_config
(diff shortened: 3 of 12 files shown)


src/aws_durable_execution_sdk_python/config.py
zhongkechen, line 285: Let's remove `is_virtual` from `ChildConfig`. It's a mistake to expose this field to users. This is supposed to be used only by concurrency operations internally

ParidelPooya: > It's a mistake to expose this field to users

I think it's useful, TypeScript SDK is exposing it.

ParidelPooya (approved):

src/aws_durable_execution_sdk_python/config.py
zhongkechen: That is a different feature, which is not in the scope of FLAT nesting type.

created a feature request for Java.
https://github.com/aws/aws-durable-execution-sdk-java/issues/360

PR #367, opened 2026-05-01
test: make parallel serdes assertions order independent

### LINKED ISSUE #366: fix: parallel test reliability
`tests/operation/parallel_test.py::test_parallel_item_deserialize` is flaky. It's been failing intermittently on Python 3.14 in CI (e.g. [run 25191980640](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/25191980640)).

## Problem

The test asserts on the order of calls to a mocked `deserialize`:

```python
assert mock_deserialize.call_args_list[0][1]["operation_id"] == "child-0"
assert mock_deserialize.call_args_list[1][1]["operation_id"] == "child-1"
```

But `ParallelExecutor.execute` submits branches to a `ThreadPoolExecutor` with no ordering between them. Which branch reaches `deserialize` first is determined by thread scheduling, not by `executable.index`. The `_execute_item_in_child_context` docstring guarantees `operation_id` is execution-order invariant — it makes no such guarantee about observed call order.

So the test's pass/fail depends on which thread wins the race. Only one of the three parametrizations (`item_serdes2-None`) has been failing in any given run, which confirms non-determinism.

## Fix

Assert on the set of calls keyed by `operation_id` instead of on list index:

```python
calls_by_op_id = {
    call.kwargs["operation_id"]: call.kwargs
    for call in mock_deserialize.call_args_list
}

assert set(calls_by_op_id) == {"child-0", "child-1"}
assert calls_by_op_id["child-0"]["serdes"] is expected
assert calls_by_op_id["child-1"]["serdes"] is expected
```

`test_parallel_item_serialize` uses the same pattern and should get the same treatment.

Genmin:
## Summary
- key parallel serialize/deserialize mock assertions by operation_id instead of observed call order
- keep checks for child item serdes fallback and parent batch serdes

Fixes #366

## Validation
- uvx hatch test tests/operation/parallel_test.py -k "parallel_item"
- uvx hatch test tests/operation/parallel_test.py
- uvx hatch fmt --check
- uvx hatch run types:check tests/operation/parallel_test.py
- git diff --check

### DIFF
--- a/tests/operation/parallel_test.py
+++ b/tests/operation/parallel_test.py
@@ -2,6 +2,8 @@
 
 import importlib
 import json
+from collections.abc import Mapping
+from typing import Any
 from unittest.mock import Mock, patch
 
 import pytest
@@ -50,6 +52,11 @@ def create_test_context(
     return DurableContext(
         state=state, execution_context=execution_context, parent_id=parent_id
     )
+
+
+def _mock_call_kwargs_by_operation_id(mock: Mock) -> dict[str, Mapping[str, Any]]:
+    """Return mock call keyword arguments keyed by operation_id."""
+    return {call.kwargs["operation_id"]: call.kwargs for call in mock.call_args_list}
 
 
 def test_parallel_executor_init():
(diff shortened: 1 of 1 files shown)


yaythomas (approved):
thank you so much @Genmin, great contribution, welcome to the repo! 🙇

PR #368, opened 2026-05-01
fix: change format of user agent header

nvasiu:
- Update the UserAgent string format to be consistent with the format used in other language SDKs.

*Issue #, if available:*

N/A

*Description of changes:*

- Updated the format of the `user_agent_extra` to match the new format being used in the other language SDKs.
- The new format is `aws-durable-execution-sdk-python/{version}`.

### DIFF
--- a/src/aws_durable_execution_sdk_python/lambda_service.py
+++ b/src/aws_durable_execution_sdk_python/lambda_service.py
@@ -18,6 +18,7 @@ from aws_durable_execution_sdk_python.exceptions import (
     GetExecutionStateError,
 )
 
+
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -18,6 +18,7 @@
     GetExecutionStateError,
 )
 
+
```
nvasiu, line 21: Empty lines are automatically being added to these files after saving due to the Ruff formatting rules in this repo (see [here](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/pyproject.toml#L100))

yaythomas (approved):

PR #369, opened 2026-05-01
refactor: move ops/ scripts to .github/scripts/

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/345

*Description of changes:* 
- The `ops/` package name was misleading, suggesting SDK runtime tooling rather than CI/workflow scripts. Move all scripts and tests under .github/scripts/ where they naturally belong.
- Also run lintcommit.py via hatch in ci-checks.sh so the script works without a system python on PATH.


CONTRIBUTING.md
```diff
@@ -27,15 +27,15 @@ hatch fmt
 
 There is a convenience script for the above that you can run from the root of the repo as you prepare your PR:
 ```
-ops/ci-checks.sh
+.github/scripts/ci-checks.sh
```
zhongkechen, line 30: curious if we have git hook for this

zhongkechen (approved):

CONTRIBUTING.md
ayushiahjolia: No git hook currently. The linter runs in CI on PRs and locally via ci-checks.sh. We could add a commit-msg hook as a follow-up if that'd be useful.

PR #371, opened 2026-05-01
[Feature]: experimental feature - add plugin interface

### LINKED ISSUE #370: [Feature]: durable execution event hooks
### What would you like?

- onOperationStart, onOperationEnd fire at most once per operation.
- onInvocationStart, onInvocationEnd fire at most once per invocation
- onUserFunctionStart, onUserFunctionEnd fire when a user function is invoked

proposed design doc: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/PYTHON_SDK_OTEL_PLUGIN.md

### Key Differences

There is a key difference in this implementation from JavaScript implementation
- Python calls onUserFunctionStart before a user function is called , while JS calls it when the user function is completed
- Python calls onOperationStart when an operation is started, while JS calls it when the operation is completed
 
### Possible Implementation

Similar work in TypeScript SDK for reference:

- notes about plugin interface: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/PLUGIN_IMPLEMENTATION.md

- plugin interface: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/types/plugin.ts
- To see how onOperationStart, onOperationEnd, onOperationAttemptStart, onOperationAttemptEnd are currently wired in the TS SDK, you'll have to check the [branch](https://github.com/aws/aws-durable-execution-sdk-js/tree/otel-instrumentation-design-v2). For example, [onOperationStart](https://github.com/aws/aws-durable-execution-sdk-js/blame/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/handlers/step-handler/step-handler.ts#L120).
- wiring of onExecutionStart, onInvocationStart, onExecutionEnd, onInvocationEnd: https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/with-durable-execution.ts
- how TS dash0 plugin (v1) deals with duplicate span creations: https://github.com/aws/aws-durable-execution-sdk-js/commit/d75518f5b7a69dc30e65cc8326e9d97dcf4af639#diff-d172a83e0787d5e5a395b4167323ac0d2e24043acca5355467f50a5b7390b7d0

No

### Does this require an RFC?

Yes

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

- fix: #370 

*Description of changes:*

- add plugin interface. The interface exposes the following 6 methods

```
class DurableInstrumentationPlugin:
    """Base class for plugins. Override only the methods you need."""

    def on_invocation_start(self, info: InvocationStartInfo) -> None:
        """Called when an invocation starts. This is called within the thread that runs user function handler.

        Args:
            info: Information about the invocation.
        """
        pass

    def on_invocation_end(self, info: InvocationEndInfo) -> None:
        """Called when an invocation ends. This is called within the thread that runs user function handler.

        Args:
            info: Information about the invocation.
        """
        pass

    def on_operation_start(self, info: OperationStartInfo) -> None:
        """
        Called when an operation checkpoints STARTED status. This is called NOT within the thread that runs operation.

        Args:
            info: Information about the operation.

        """
        pass

    def on_operation_end(self, info: OperationEndInfo) -> None:
        """
        Called when an operation checkpoints a terminal status. This is called NOT within the thread that runs operation.

        Args:
            info: Information about the operation.
        """
        pass

    def on_user_function_start(self, info: UserFunctionStartInfo) -> None:
        """Called when an operation starts to execute user provided function. This is called within the thread that runs user provided function.

        Args:
            info: Information about the operation attempt.
        """
        pass

    def on_user_function_end(self, info: UserFunctionEndInfo) -> None:
        """Called when an operation finishes executing user provided function. This is called within the thread that runs user provided function.

        Args:
            info: Information about the operation attempt.
        """
        pass
```

- wiring the plugin interface into the SDK

    - on_invocation_start/on_invocation_end are wired in `durable_execution` where we start and complete the user durable function handler.
    - on_user_function_start/on_user_function_end are wired in operations (e.g. context, step, etc) where we call user functions.
    - on_operation_start/on_operation_end are wired in `ExecutionState.checkpoint_batches_forever` where we checkpoint operations.

- plugin test case:

https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26666530204/job/78606457703

image

Note: current operation start/end and user function start/end are not correctly ordered because operation start/end methods are called asynchronously and the others are called synchronously. 

### DIFF
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if hatch fmt --check; then
+    echo "Hatch fmt check passed!"
+else
+    hatch fmt
+    echo "Error: hatch fmt modified your files. Please re-stage and commit again."
+    exit 1
+fi
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
@@ -4,10 +4,11 @@ import contextlib
 import functools
 import json
 import logging
+import warnings
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
-from enum import Enum
 from typing import TYPE_CHECKING, Any
+
 
 from aws_durable_execution_sdk_python.context import DurableContext
 from aws_durable_execution_sdk_python.exceptions import (
@@ -26,6 +27,12 @@ from aws_durable_execution_sdk_python.lambda_service import (
     Operation,
     OperationType,
     OperationUpdate,
+    InvocationStatus,
+    DurableExecutionInvocationOutput,
+)
+from aws_durable_execution_sdk_python.plugin import (
+    DurableInstrumentationPlugin,
+    PluginExecutor,
 )
 from aws_durable_execution_sdk_python.state import ExecutionState, ReplayStatus
 
@@ -149,62 +156,6 @@ class DurableExecutionInvocationInputWithClient(DurableExecutionInvocationInput)
         )
 
 
-class InvocationStatus(Enum):
-    SUCCEEDED = "SUCCEEDED"
-    FAILED = "FAILED"
-    PENDING = "PENDING"
-
-
-@dataclass(frozen=True)
-class DurableExecutionInvocationOutput:
-    """Representation the DurableExecutionInvocationOutput. This is what the Durable lambda handler returns.
-
-    If the execution has been already completed via an update to the EXECUTION operation via CheckpointDurableExecution,
-    payload must be empty for SUCCEEDED/FAILED status.
-    """
-
-    status: InvocationStatus
-    result: str | None = None
-    error: ErrorObject | None = None
-
-    @classmethod
-    def from_dict(
-        cls, data: MutableMapping[str, Any]
-    ) -> DurableExecutionInvocationOutput:
-        """Create an instance from a dictionary.
-
-        Args:
-            data: Dictionary with camelCase keys matching the original structure
-
-        Returns:
-            A DurableExecutionInvocationOutput instance
-        """
-        status = InvocationStatus(data.get("Status"))
-        error = ErrorObject.from_dict(data["Error"]) if data.get("Error") else None
-        return cls(status=status, result=data.get("Result"), error=error)
-
-    def to_dict(self) -> MutableMapping[str, Any]:
-        """Convert to a dictionary with the original field names.
-
-        Returns:
-            Dictionary with the original camelCase keys
-        """
-        result: MutableMapping[str, Any] = {"Status": self.status.value}
-
-        if self.result is not None:
-            # large payloads return "", because checkpointed already
-            result["Result"] = self.result
-        if self.error:
-            result["Error"] = self.error.to_dict()
-
-        return result
-
-    @classmethod
-    def create_succeeded(cls, result: str) -> DurableExecutionInvocationOutput:
-        """Create a succeeded invocation output."""
-        return cls(status=InvocationStatus.SUCCEEDED, result=result)
-
-
 # endregion Invocation models
 
 
@@ -212,14 +163,36 @@ def durable_execution(
     func: Callable[[Any, DurableContext], Any] | None = None,
     *,
     boto3_client: Boto3LambdaClient | None = None,
+    plugins: list[DurableInstrumentationPlugin] | None = None,
 ) -> Callable[[Any, LambdaContext], Any]:
+    """
+    Decorator to create a durable execution handler.
+
+    Args:
+        func: The user function to decorate
+        boto3_client: Optional boto3 Lambda client to use
+        plugins: Optional list of plugins to use (EXPERIMENTAL: This
+            feature has known issues and this parameter may change or be removed.)
+    """
     # Decorator called with parameters
     if func is None:
         logger.debug("Decorator called with parameters")
-        return functools.partial(durable_execution, boto3_client=boto3_client)
+        return functools.partial(
+            durable_execution, boto3_client=boto3_client, plugins=plugins
+        )
 
     logger.debug("Starting durable execution handler...")
 
+    if plugins:
+        warnings.warn(
+            "The 'plugins' parameter is provisional and may be altered or removed.",
+            category=FutureWarning,
+            stacklevel=2,  # point the warning to the caller of durable_execution
+        )
+
+    plugin_executor = PluginExecutor(plugins)
+
+    @plugin_executor.handle_durable_output
     def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
         invocation_input: DurableExecutionInvocationInput
         service_client: DurableServiceClient
@@ -255,6 +228,7 @@ def durable_execution(
             operations={},
             service_client=service_client,
             replay_status=ReplayStatus.NEW,
+            plugin_executor=plugin_executor,
         )
 
         try:
@@ -306,6 +280,13 @@ def durable_execution(
             ) as executor,
             contextlib.closing(execution_state) as execution_state,
         ):
+            # execute the plugins
+            plugin_executor.on_invocation_start(
+                execution_arn=invocation_input.durable_execution_arn,
+                lambda_context=context,
+                execution_start_time=execution_state.get_execution_operation().start_timestamp,
+                is_first_invocation=not execution_state.is_replaying(),
+            )
             # Thread 1: Run background checkpoint processing
             executor.submit(execution_state.checkpoint_batches_forever)
 
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
@@ -73,6 +73,29 @@ class OperationType(Enum):
     CALLBACK = "CALLBACK"
     CHAINED_INVOKE = "CHAINED_INVOKE"
 
+    @classmethod
+    def from_sub_type(cls, sub_type: OperationSubType) -> OperationType:
+        match sub_type:
+            case OperationSubType.STEP | OperationSubType.WAIT_FOR_CONDITION:
+                return OperationType.STEP
+            case OperationSubType.WAIT:
+                return OperationType.WAIT
+            case OperationSubType.CHAINED_INVOKE:
+                return OperationType.CHAINED_INVOKE
+            case OperationSubType.CALLBACK:
+                return OperationType.CALLBACK
+            case (
+                OperationSubType.WAIT_FOR_CALLBACK
+                | OperationSubType.RUN_IN_CHILD_CONTEXT
+                | OperationSubType.MAP
+                | OperationSubType.MAP_ITERATION
+                | OperationSubType.PARALLEL
+                | OperationSubType.PARALLEL_BRANCH
+            ):
+                return OperationType.CONTEXT
+            case _:
+                raise ValueError(f"Unknown operation sub-type {sub_type}")
+
 
 class CallbackTimeoutType(Enum):
     TIMEOUT = "Callback.Timeout"
@@ -103,6 +126,70 @@ class OperationSubType(Enum):
     WAIT_FOR_CALLBACK = "WaitForCallback"
     WAIT_FOR_CONDITION = "WaitForCondition"
     CHAINED_INVOKE = "ChainedInvoke"
+
+
+class InvocationStatus(Enum):
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+    PENDING = "PENDING"
+
+    # Used internally only: the invocation failed and the backend will retry
+    RETRY = "RETRY"
+
+
+@dataclass(frozen=True)
+class DurableExecutionInvocationOutput:
+    """Representation the DurableExecutionInvocationOutput. This is what the Durable lambda handler returns.
+
+    If the execution has been already completed via an update to the EXECUTION operation via CheckpointDurableExecution,
+    payload must be empty for SUCCEEDED/FAILED status.
+    """
+
+    status: InvocationStatus
+    result: str | None = None
+    error: ErrorObject | None = None
+
+    @classmethod
+    def from_dict(
+        cls, data: MutableMapping[str, Any]
+    ) -> DurableExecutionInvocationOutput:
+        """Create an instance from a dictionary.
+
+        Args:
+            data: Dictionary with camelCase keys matching the original structure
+
+        Returns:
+            A DurableExecutionInvocationOutput instance
+        """
+        status = InvocationStatus(data.get("Status"))
+        error = ErrorObject.from_dict(data["Error"]) if data.get("Error") else None
+        return cls(status=status, result=data.get("Result"), error=error)
+
+    def to_dict(self) -> MutableMapping[str, Any]:
+        """Convert to a dictionary with the original field names.
+
+        Returns:
+            Dictionary with the original camelCase keys
+        """
+        result: MutableMapping[str, Any] = {"Status": self.status.value}
+
+        if self.result is not None:
+            # large payloads return "", because checkpointed already
+            result["Result"] = self.result
+        if self.error:
+            result["Error"] = self.error.to_dict()
+
+        return result
+
+    @classmethod
+    def create_succeeded(cls, result: str) -> DurableExecutionInvocationOutput:
+        """Create a succeeded invocation output."""
+        return cls(status=InvocationStatus.SUCCEEDED, result=result)
+
+    @classmethod
+    def create_retry(cls, error: ErrorObject) -> DurableExecutionInvocationOutput:
+        """Create a failed invocation output."""
+        return cls(status=InvocationStatus.RETRY, error=error)
 
 
 @dataclass(frozen=True)
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
@@ -154,9 +154,15 @@ class ChildOperationExecutor(OperationExecutor[T]):
             self.operation_identifier.operation_id,
             self.operation_identifier.name,
         )
-
         try:
-            raw_result: T = self.func()
+            # todo: fix attempt (checkpointed_result.is_existent is always True)
+            wrapped_user_func = self.state.wrap_user_function(
+                self.func,
+                self.operation_identifier,
+                checkpointed_result.is_replay_children(),
+                attempt=None if checkpointed_result.is_existent() else 1,
+            )
+            raw_result: T = wrapped_user_func()
 
             if self.is_virtual:
                 logger.debug(
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
@@ -1,0 +1,408 @@
+import contextlib
+import datetime
+import functools
+import logging
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Callable, MutableMapping
+
+from aws_durable_execution_sdk_python.exceptions import SuspendExecution
+from aws_durable_execution_sdk_python.identifier import OperationIdentifier
+from aws_durable_execution_sdk_python.lambda_service import (
+    OperationType,
+    OperationStatus,
+    OperationAction,
+    OperationSubType,
+    ErrorObject,
+    InvocationStatus,
+    Operation,
+    OperationUpdate,
+    DurableExecutionInvocationOutput,
+)
+from aws_durable_execution_sdk_python.types import LambdaContext
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class OperationInfo:
+    operation_id: str
+    operation_type: OperationType
+    sub_type: OperationSubType | None
+    name: str | None
+    parent_id: str | None
+    start_time: datetime.datetime | None
+
+
+@dataclass(frozen=True)
+class OperationStartInfo(OperationInfo):
+    pass
+
+
+@dataclass(frozen=True)
+class OperationEndInfo(OperationInfo):
+    status: OperationStatus
+    end_time: datetime.datetime | None
+    error: ErrorObject | None
+
+
+class UserFunctionOutcome(Enum):
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+    PENDING = "PENDING"
+
+    @classmethod
+    def from_error(cls, error: ErrorObject | None) -> "UserFunctionOutcome":
+        if error is None:
+            return cls(cls.SUCCEEDED)
+        elif error.type == SuspendExecution.__name__:
+            return cls(cls.PENDING)
+        else:
+            return cls(cls.FAILED)
+
+
+@dataclass(frozen=True)
+class UserFunctionStartInfo(OperationInfo):
+    is_replay_children: bool = (
+        False  # True if user function is called to replay children (MAP/PARALLEL)
+    )
+    attempt: int | None = (
+        None  # None for user function called more than once in CONTEXT
+    )
+
+
+@dataclass(frozen=True)
+class UserFunctionEndInfo(OperationInfo):
+    is_replay_children: (
+        bool  # True if user function is called to replay children (MAP/PARALLEL)
+    )
+    attempt: int | None  # None for user function called more than once in CONTEXT
+    outcome: UserFunctionOutcome
+    end_time: datetime.datetime | None
+    error: ErrorObject | None
+
+    @classmethod
+    def from_start_info(
+        cls, start_info: UserFunctionStartInfo, error: ErrorObject | None
+    ) -> "UserFunctionEndInfo":
+        return UserFunctionEndInfo(
+            operation_id=start_info.operation_id,
+            operation_type=start_info.operation_type,
+            sub_type=start_info.sub_type,
+            name=start_info.name,
+            parent_id=start_info.parent_id,
+            start_time=start_info.start_time,
+            is_replay_children=start_info.is_replay_children,
+            attempt=start_info.attempt,
+            outcome=UserFunctionOutcome.from_error(error),
+            end_time=datetime.datetime.now(datetime.UTC),
+            error=error,
+        )
+
+
+@dataclass(frozen=True)
+class InvocationInfo:
+    request_id: str | None
+    execution_arn: str | None
+    start_time: datetime.datetime | None
+    is_first_invocation: bool
+
+
+@dataclass(frozen=True)
+class InvocationStartInfo(InvocationInfo):
+    pass
+
+
+@dataclass(frozen=True)
+class InvocationEndInfo(InvocationInfo):
+    status: InvocationStatus
+    end_time: datetime.datetime | None
+    error: ErrorObject | None
+
+    @classmethod
+    def from_durable_execution_invocation_output(
+        cls,
+        invocation_start_info: InvocationStartInfo,
+        output: "DurableExecutionInvocationOutput",
+    ):
+        return InvocationEndInfo(
+            request_id=invocation_start_info.request_id,
+            execution_arn=invocation_start_info.execution_arn,
+            start_time=invocation_start_info.start_time,
+            is_first_invocation=invocation_start_info.is_first_invocation,
+            status=output.status,
+            end_time=datetime.datetime.now(datetime.UTC),
+            error=output.error,
+        )
+
+
+class DurableInstrumentationPlugin:
+    """Base class for plugins. Override only the methods you need."""
+
+    def on_invocation_start(self, info: InvocationStartInfo) -> None:
+        """Called when an invocation starts. This is called within the thread that runs user function handler.
+
+        Args:
+            info: Information about the invocation.
+        """
+        pass
+
+    def on_invocation_end(self, info: InvocationEndInfo) -> None:
+        """Called when an invocation ends. This is called within the thread that runs user function handler.
+
+        Args:
+            info: Information about the invocation.
+        """
+        pass
+
+    def on_operation_start(self, info: OperationStartInfo) -> None:
+        """
+        Called when an operation checkpoints STARTED status. This is called NOT within the thread that runs operation.
+
+        Args:
+            info: Information about the operation.
+
+        """
+        pass
+
+    def on_operation_end(self, info: OperationEndInfo) -> None:
+        """
+        Called when an operation checkpoints a terminal status. This is called NOT within the thread that runs operation.
+
+        Args:
+            info: Information about the operation.
+        """
+        pass
+
+    def on_user_function_start(self, info: UserFunctionStartInfo) -> None:
+        """Called when an operation starts to execute user provided function. This is called within the thread that runs user provided function.
+
+        Args:
+            info: Information about the operation attempt.
+        """
+        pass
+
+    def on_user_function_end(self, info: UserFunctionEndInfo) -> None:
+        """Called when an operation finishes executing user provided function. This is called within the thread that runs user provided function.
+
+        Args:
+            info: Information about the operation attempt.
+        """
+        pass
+
+    # Todo: further discussions required to finalize the following interface
+    # def enrich_log_context(self, info: OperationStartInfo | None) -> Dict[str, Any] | None: pass
+
+
+class PluginExecutor:
+    def __init__(self, plugins: list[DurableInstrumentationPlugin] | None):
+        self._plugins = plugins or []
+        self._executor: ThreadPoolExecutor | None = None
+        self._invocation_status: InvocationStartInfo | None = None
+
+    @contextlib.contextmanager
+    def run(self):
+        if self._plugins:
+            self._executor = ThreadPoolExecutor(
+                max_workers=1,
+                thread_name_prefix="plugin-executor",
+            )
+        try:
+            yield
+        finally:
+            self._invocation_status = None
+            # Shut down the thread pool, waiting for pending tasks to complete.
+            if self._executor:
+                self._executor.shutdown(wait=True)
+
+    @staticmethod
+    def _dispatch_plugin(plugin: DurableInstrumentationPlugin, info) -> None:
+        """Invoke the appropriate plugin callback. Runs inside the thread pool."""
+        try:
+            match info:
+                case InvocationStartInfo():
+                    plugin.on_invocation_start(info)
+                case InvocationEndInfo():
+                    plugin.on_invocation_end(info)
+                case OperationStartInfo():
+                    plugin.on_operation_start(info)
+                case OperationEndInfo():
+                    plugin.on_operation_end(info)
+                case UserFunctionStartInfo():
+                    plugin.on_user_function_start(info)
+                case UserFunctionEndInfo():
+                    plugin.on_user_function_end(info)
+                case _:
+                    raise RuntimeError(f"Unknown info type: {type(info)}")
+        except Exception:
+            # log and ignore the exception
+            logger.exception("Plugin %s exception ignored", plugin.__class__.__name__)
+
+    def execute_plugins(self, info, sync):
+        if not self._executor:
+            return
+        for plugin in self._plugins:
+            if sync:
+                # this is called synchronously, so plugins will be able to manipulate thread local objects
+                self._dispatch_plugin(plugin, info)
+            else:
+                # this is called asynchronously, so plugins cannot manipulate thread local objects
+                self._executor.submit(self._dispatch_plugin, plugin, info)
+
+    def on_invocation_start(
+        self,
+        execution_arn: str,
+        is_first_invocation: bool,
+        execution_start_time: datetime.datetime | None,
+        lambda_context: LambdaContext | None,
+    ) -> None:
+        aws_request_id = lambda_context.aws_request_id if lambda_context else None
+        invocation_start_time = (
+            datetime.datetime.now(datetime.UTC)
+            if is_first_invocation
+            else execution_start_time
+        )
+        self._invocation_status = InvocationStartInfo(
+            execution_arn=execution_arn,
+            request_id=aws_request_id,
+            is_first_invocation=is_first_invocation,
+            start_time=invocation_start_time,
+        )
+        self.execute_plugins(self._invocation_status, sync=True)
+
+    def on_invocation_end(
+        self,
+        output: "DurableExecutionInvocationOutput",
+    ) -> None:
+        if self._invocation_status is None:
+            # on_invocation_start not called, skip
+            return
+
+        invocation_end_info = (
+            InvocationEndInfo.from_durable_execution_invocation_output(
+                self._invocation_status, output
+            )
+        )
+        self.execute_plugins(invocation_end_info, sync=True)
+
+    def on_user_function_start(
+        self,
+        operation_identifier: OperationIdentifier,
+        is_replay_children: bool = False,
+        attempt: int | None = None,
+    ) -> UserFunctionStartInfo:
+        """Execute any registered plugins for the operation when its user function starts to execute."""
+        start_info = UserFunctionStartInfo(
+            operation_id=operation_identifier.operation_id,
+            operation_type=operation_identifier.type,
+            sub_type=operation_identifier.sub_type,
+            name=operation_identifier.name,
+            parent_id=operation_identifier.parent_id,
+            start_time=datetime.datetime.now(datetime.UTC),
+            is_replay_children=is_replay_children,
+            attempt=attempt,
+        )
+        self.execute_plugins(start_info, sync=True)
+        return start_info
+
+    def on_user_function_end(self, start_info: UserFunctionStartInfo, error) -> None:
+        """Execute any registered plugins for the operation when its user function finishes execution."""
+        self.execute_plugins(
+            UserFunctionEndInfo.from_start_info(start_info, error), sync=True
+        )
+
+    def on_operation_action(self, update: OperationUpdate):
+        """Execute any registered plugins for a given operation when an update is checkpointed
+
+        Args:
+            update: the operation update that is checkpointed
+        """
+        if update.action is OperationAction.START:
+            # we handle only START action here because on_operation_update may not be able to see a STARTED update
+            # when START is checkpointed in batch with terminal status updates.
+            self.execute_plugins(
+                OperationStartInfo(
+                    operation_id=update.operation_id,
+                    operation_type=update.operation_type,
+                    sub_type=update.sub_type,
+                    name=update.name,
+                    parent_id=update.parent_id,
+                    start_time=datetime.datetime.now(datetime.UTC),
+                ),
+                sync=False,
+            )
+
+    def on_operation_update(self, operation: Operation | None):
+        """Execute any registered plugins for a given operation when it receives an update
+
+        Updates such as STARTED might be omitted because START and completion action (e.g. SUCCEED/FAIL) may be
+        checkpointed in batch and the backend returns only the terminal status (e.g. SUCCEEDED/PENDING/FAILED).
+
+        Note: the operation may not be up-to-date if the checkpoint is called asynchronously.
+
+        Args:
+            operation: the operation is just checkpointed
+        """
+        if operation and self._is_terminal_status(operation.status):
+            self.execute_plugins(
+                OperationEndInfo(
+                    operation_id=operation.operation_id,
+                    operation_type=operation.operation_type,
+                    sub_type=operation.sub_type,
+                    name=operation.name,
+                    parent_id=operation.parent_id,
+                    start_time=operation.start_timestamp,
+                    end_time=operation.end_timestamp,
+                    status=operation.status,
+                    error=self._extract_error(operation),
+                ),
+                sync=False,
+            )
+
+    @staticmethod
+    def _extract_error(operation: Operation):
+        if operation.step_details and operation.step_details.error:
+            return operation.step_details.error
+        if operation.callback_details and operation.callback_details.error:
+            return operation.callback_details.error
+        if operation.chained_invoke_details and operation.chained_invoke_details.error:
+            return operation.chained_invoke_details.error
+        if operation.context_details and operation.context_details.error:
+            return operation.context_details.error
+        return None
+
+    @staticmethod
+    def _is_terminal_status(status):
+        return status in [
+            OperationStatus.SUCCEEDED,
+            OperationStatus.FAILED,
+            OperationStatus.TIMED_OUT,
+            OperationStatus.CANCELLED,
+            OperationStatus.STOPPED,
+        ]
+
+    @property
+    def handle_durable_output(self):
+        def decorator(func: Callable[[Any, LambdaContext], MutableMapping[str, Any]]):
+            @functools.wraps(func)
+            def wrapper(event: Any, context: LambdaContext):
+                with self.run():
+                    try:
+                        output = func(event, context)
+
+                        self.on_invocation_end(
+                            output=DurableExecutionInvocationOutput.from_dict(output),
+                        )
+                        return output
+                    except Exception as e:
+                        self.on_invocation_end(
+                            output=DurableExecutionInvocationOutput.create_retry(
+                                ErrorObject.from_exception(e)
+                            ),
+                        )
+                        raise
+
+            return wrapper
+
+        return decorator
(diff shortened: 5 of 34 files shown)


src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,299 @@
...
+                    end_timestamp=operation.end_timestamp,
+                    attempt=operation.step_details.attempt,
+                    succeeded=operation.status is OperationStatus.SUCCEEDED,
+                    error=operation.step_details.error,
+                )
+            )
+
+        if self._is_terminal_status(operation.status):
+            attempt = operation.step_details.attempt if operation.step_details else None
+            self.execute_plugins(OperationStartInfo(**params))
```
SilanHe: do we need to call this here? Since we only want to trigger the `on_operation_attempt_start` once, I think the `on_operation_action` is probably enough, right?

tests/state_test.py
```diff
@@ -3733,3 +3737,411 @@ def test_initial_execution_state_get_input_payload_none():
...
+            operation_type=OperationType.STEP,
+            action=OperationAction.START,
+            name="my-step",
+        )
+        state.create_checkpoint(operation_update, is_sync=True)
+
+        # on_operation_action fires for START
+        assert "operation_start:step-1" in plugin.calls
+        # But on_operation_update should NOT fire operation_end for STARTED status
+        operation_end_calls = [c for c in plugin.calls if c.startswith("operation_end")]
```
SilanHe: hmm what is this test actually testing for? The comments are confusing

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Yeah, forgot to remove this after it was moved to another place. Updated in the latest commit

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -230,43 +233,56 @@ def __init__(
...
-        self.boto3_client = boto3_client
-        self.event = event
-        self.context = context
-        self.invocation_input = self._parse_invocation_input(event)
-        self.service_client = self._parse_service_client(event, boto3_client)
+        self._func = func
+        self._context = context
+        self._plugin_executor = PluginExecutor(plugins)
+
+        self._print_durable_execution_arn(event)
```
SilanHe: nit: what do we need this for?

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -299,143 +307,137 @@ def execute(self):
...
-        if raw_input_payload and raw_input_payload.strip():
-            try:
-                input_event = json.loads(raw_input_payload)
-            except json.JSONDecodeError as e:
-                logger.exception(
-                    "Failed to parse input payload as JSON: payload: %r",
-                    raw_input_payload,
-                )
-                self._handle_execution_output(exception=e, retryable=True)
+        input_event: Any = self._get_input_event()
```
SilanHe: how come we're not using MutableMapping anymore?

tests/e2e/checkpoint_response_int_test.py
```diff
@@ -101,7 +101,7 @@ def my_handler(event, context: DurableContext) -> str:
         mock_client.checkpoint = mock_checkpoint
 
         event = {
-            "DurableExecutionArn": "test-arn",
+            "DurableExecutionArn": "test-arn/execution-1",
```
SilanHe, line 104: nit: why did we need this change?

tests/logger_test.py
```diff
@@ -366,6 +368,11 @@ def test_logger_without_mocked_logger():
 
 
 def test_logger_replay_no_logging():
+    execution = Operation(
```
SilanHe: nit: is this local variable being used?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,317 @@
...
+    name: str | None = None
+    parent_id: str | None = None
+    start_timestamp: datetime.datetime | None = None
+
+
+@dataclass
+class OperationEndInfo(OperationStartInfo):
+    status: OperationStatus = OperationStatus.SUCCEEDED
+    end_timestamp: datetime.datetime | None = None
+    attempt: int | None = None
```
SilanHe: why do we need attempt in OperationEndInfo?

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: `_print_durable_execution_arn` is added to address the comment: https://github.com/aws/aws-durable-execution-sdk-python/pull/375#discussion_r3192207198 which I believe rarely happens though

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Should be any because the input event could be string, list or whatever types that Json allows

tests/e2e/checkpoint_response_int_test.py
zhongkechen: This was from a previous PR https://github.com/aws/aws-durable-execution-sdk-python/pull/378 where execution operation is strictly checked and exception is raised when missing. That PR was abandoned due to circular dependencies. It's no harm to keep this and we may need to add strict execution operation checking.

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: I supposed we can provide the info of how many times that the step has been attempted until it reaches the terminal state. Maybe `attempts` make more sense?

tests/logger_test.py
zhongkechen: thx. removed.

src/aws_durable_execution_sdk_python/plugin.py
SilanHe: yeah, that makes more sense

SilanHe (dismissed):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -230,43 +242,56 @@ def __init__(
...
+        # parsing input, which may throw exceptions
+        self._invocation_input = self._parse_invocation_input(event)
+        self._service_client = self._parse_service_client(event, boto3_client)
+
+        self._execution_state = ExecutionState(
+            durable_execution_arn=self._invocation_input.durable_execution_arn,
+            initial_checkpoint_token=self._invocation_input.checkpoint_token,
+            operations={},
+            service_client=self._service_client,
+            replay_status=ReplayStatus.NEW,
```
yaythomas, line 230: this doesn't pass `PluginExecutor`, is this deliberate? 

because as it stands that means there is an ` self._plugin_executor = PluginExecutor(plugins)` defined here on the class, but then in `ExecutionState` there is _another_ instance of `PluginExecutor`.

so `on_invocation_start` and `on_invocation_end` is on the `DurableExecutionExecutor`'s instance of `PluginExecutor`, and `on_operation_action` and `on_operation_update` is on the other instance that is in `ExecutionState`.

as it is, no test catches this either.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -476,32 +485,48 @@ def _handle_large_error(
...
-            execution_state.create_checkpoint_sync(failed_operation)
+            self._execution_state.create_checkpoint_sync(failed_operation)
 
             # return fail without an ErrorObject
             return None
 
         return error
 
+    def _handle_execution_exception(self, exception: Exception, retryable: bool):
+        execution_arn: str | None = getattr(self, "_durable_execution_arn", None)
```
yaythomas: as far as I can tell `_durable_execution_arn` and `execution_state` doesn't exist on self at all?

I _guess_ these `getattr` are here to tolerate the possibility of partial construction? so if init is not fully defined yet, but an error throws out of parse_invocation_input and `self._invocation_input` and `self._execution_state` don't exist yet? Because the `__init__` is doing a bunch of work that can fail, and this method is called from init -> _parse_invocation_input.

this is actually a good structural demonstration of why this repo in the contribution guide here https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#initialization-and-conversion states
> Class constructors must be light and not do more than initialize the class. In a dataclass you shouldn't even need an __init__. Use a @classmethod factory method instead to encapsulate more advanced logic. For example, if a class depends on logic that might fail, encapsulate this in a create classmethod:

when there are side-effects or things that can fail in `__init__` it can easily lead to surprising unintended side-effects like these. this problem here would have been easier to notice with a type-checker if the code have relied on `self.` rather than getattr.

reasons like these are why the previous PR asked please to use a factory method like `create` in this discussion https://github.com/aws/aws-durable-execution-sdk-python/pull/375#discussion_r3192183068.

this would have bypassed the issue here entirely. the other issue further up with the get() on durable execution arn also came to be for this same cause.

in summary, moving the risky parsing out of `__init__` into classmethods, so __init__ can't fail:


```python
  class DurableExecutionInvocationInput:
       @classmethod
       def from_event(cls, event) -> DurableExecutionInvocationInput:
           if isinstance(event, DurableExecutionInvocationInputWithClient):
               parsed = event
           else:
               try:
                   parsed = cls.from_json_dict(event)
               except (KeyError, TypeError, AttributeError) as e:
                   raise ExecutionError(
                       "Unexpected payload provided to start the durable execution. "
                       "Check your resource configurations to confirm the   durability is set."
                   ) from e
           logger.debug("durableExecutionArn: %s", parsed.durable_execution_arn)
           return parsed
```

and then

```python
   class DurableExecutionExecutor:
       def __init__(
           self,
           func,
           invocation_input: DurableExecutionInvocationInput,
           service_client: DurableServiceClient,
           execution_state: ExecutionState,
           plugin_executor: PluginExecutor,
           context: LambdaContext,
       ):
           # pure assignment — every line is typed, no line can raise
           self._func = func
           self._invocation_input = invocation_input
           self._service_client = service_client
           self._execution_state = execution_state
           self._plugin_executor = plugin_executor
           self._context = context

       @classmethod
       def create(cls, func, boto3_client, plugins, event, context):
           invocation_input = DurableExecutionInvocationInput.from_event(event)
           service_client = cls._build_service_client(event, boto3_client)
           plugin_executor = PluginExecutor(plugins)
           execution_state = ExecutionState(
               durable_execution_arn=invocation_input.durable_execution_arn,
               initial_checkpoint_token=invocation_input.checkpoint_token,
               operations={},
               service_client=service_client,
               plugin_executor=plugin_executor,
               replay_status=ReplayStatus.NEW,
           )
           return cls(func, invocation_input, service_client, execution_state,
   plugin_executor, context)
```

And then the decorator wrapper becomes the single place that decides what to do on parse failure:

```python
   def wrapper(event, context):
       try:
           executor = DurableExecutionExecutor.create(
               func, boto3_client, plugins, event, context
           )
       except ExecutionError:
           raise   # Lambda retries per its retry policy. No plugin to notify —
   we never started.
       return executor.execute()
```

yaythomas: We could "get around" the issue here by 
a) fixing the attribute names on the getattr, but the point is that If this had used self._invocation_input directly the type checker would have flagged the typo. 

b) shuffle the initialization order in the __init__ so those fields are defined before you could potentially call the err handler, but now we're coding in subtle relationships between methods.

Avoiding subtleties like these is why the repo style guide favors factories for complex construction with light inits.

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+        execution_operation: Operation | None,
+        output: "DurableExecutionInvocationOutput",
+    ) -> None:
+        start_timestamp = (
+            execution_operation.start_timestamp if execution_operation else None
+        )
+        # the actual end timestamp may be unknown because it's not checkpointed yet
+        end_timestamp: datetime.datetime = (
+            execution_operation.end_timestamp if execution_operation else None
+        ) or datetime.datetime.now()
```
yaythomas: all the dates here should be timezoned
```
datetime.datetime.now(tz=datetime.UTC)
```

I haven't fully traced this through, but I think this could result in timezone offset errors, if a plugin does

```
   def on_invocation_end(self, info):
       duration = info.end_timestamp - info.start_timestamp
```

where start_timestamp comes from the `Operation.start_timestamp`, which _is_ timezoned/offset aware, but then end_timestamp is naive (no offset) because the invocation doesn't end in a terminal state?

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -299,143 +316,138 @@ def execute(self):
...
         )
 
         # Use ThreadPoolExecutor for concurrent execution of user code and background checkpoint processing
         with (
             ThreadPoolExecutor(
                 max_workers=2, thread_name_prefix="dex-handler"
             ) as executor,
-            contextlib.closing(execution_state) as execution_state,
+            contextlib.closing(self._execution_state),
+            contextlib.closing(self._plugin_executor),
```
yaythomas: this means plugin executor will exit before execution state?

PluginExecutor is blocking until it's shutdown. But `checkpoint_batches_forever()`, which is running on another thread, could still be mid-flow and trying to call `self._plugin_executor.on_operation_action`, which will end up throwing an error because its ThreadPoolExecutor is shut down already.

possible fixes:
a) 
```python
with (
    ThreadPoolExecutor(max_workers=2, thread_name_prefix="dex-handler") as executor,
    contextlib.closing(self._plugin_executor),   # close last
    contextlib.closing(self._execution_state),   # close first (signals stop)
):
```

b) 
in a previous closed PR you had a very nifty idea with a context manager like this, with a minor tweak to make it blocking on exit could look like this....

```python
   @contextmanager
   def checkpointing(
       self,
       executor: Executor,
       initial_checkpoint_token: str,
   ) -> Iterator[None]:
       future = executor.submit(self._checkpoint_batches_forever,
   initial_checkpoint_token)
       try:
           yield
       finally:
           self.stop_checkpointing()
           future.result()   # block until the checkpoint loop actually exits
``` 


and with something like that, this here could become:

```python
   with (
       ThreadPoolExecutor(max_workers=2, thread_name_prefix="dex-handler") as
    executor,
       contextlib.closing(self._plugin_executor),
   ):
       with self._execution_state.checkpointing(executor):
           user_future = executor.submit(self._func, input_event, durable_context)
           return self._handle_user_future_result(user_future)
```

tests/plugin_test.py
```diff
@@ -0,0 +1,1116 @@
...
+        )
+        plugin.on_operation_start(
+            OperationStartInfo(operation_id="o", operation_type=OperationType.WAIT)
+        )
+
+        self.assertEqual(plugin.calls, ["execution_start:r", "operation_start:o"])
+
+    def test_cannot_instantiate_abc_directly(self):
+        """DurableExecutionPlugin is abstract but has no abstract methods, so it can be instantiated via a subclass."""
+        self.assertTrue(issubclass(DurableExecutionPlugin, object))
```
yaythomas: the test name and test behaviour don't match.

as is the `class DurableExecutionPlugin(ABC):` doesn't do much other than maybe signal this is overridable?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+    def on_operation_end(self, info: OperationEndInfo) -> None:
+        pass
+
+    def on_operation_attempt_start(self, info: AttemptStartInfo) -> None:
+        pass
+
+    def on_operation_attempt_end(self, info: AttemptEndInfo) -> None:
+        pass
+
+    # Todo: further discussions required to finalize the following interface
```
yaythomas: yes. Also not sure whether the ABC is actually necessary here, unless there's going to be at least 1 abstractmethod.

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+@dataclass
+class AttemptStartInfo(OperationStartInfo):
+    attempt: int = 1
+
+
+@dataclass
+class AttemptEndInfo(AttemptStartInfo):
+    succeeded: bool | None = None
```
yaythomas: why is this a bool? 

rather than something like

```
status: OperationStatus = OperationStatus.SUCCEEDED   # ← enum
```

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+    end_timestamp: datetime.datetime | None = None
+    error: ErrorObject | None = None
+    next_attempt_delay_seconds: int | None = None
```
yaythomas: this doesn't get populated currently?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+        pass
+
+    def on_operation_attempt_start(self, info: AttemptStartInfo) -> None:
+        pass
+
+    def on_operation_attempt_end(self, info: AttemptEndInfo) -> None:
+        pass
+
+    # Todo: further discussions required to finalize the following interface
+    # def enrich_log_context(self, info: OperationStartInfo | None) -> Dict[str, Any] | None: pass
```
yaythomas: TODO: create tracking issue

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+@dataclass
+class ExecutionStartInfo(InvocationStartInfo):
+    pass
+
+
+@dataclass
+class ExecutionEndInfo(ExecutionStartInfo):
+    status: InvocationStatus = InvocationStatus.SUCCEEDED
+    end_timestamp: datetime.datetime | None = None
+    error: ErrorObject | None = None
```
yaythomas: js has more fields....

```typescript
// aws-durable-execution-sdk-js/src/types/plugin.ts
export interface ExecutionEndInfo extends InvocationInfo {
    status: "SUCCEEDED" | "FAILED";
    executionResult?: unknown;
    executionError?: Error;
    executionInput: unknown;
    operations: Record<string, Operation>;
}
```

If the idea of including these is so the execution span's attributes can include aggregate counts ("3 steps", "1 retry") without the plugin maintaining its own parallel state across events?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+            if self.plugins
+            else None
+        )
+
+    def close(self) -> None:
+        """Shut down the thread pool, waiting for pending tasks to complete."""
+        if self._executor is None:
+            return
+        self.flush()
+        self._executor.shutdown(wait=True)
```
yaythomas: so this will wait to flush all before closing. 

maybe make clear on the contract somewhere for `DurableExecutionPlugin` (or on the decorator) so consumers know when or if they have to flush themselves or not.

so `on_invocation_end` and `on_execution_end` are guaranteed to complete before the handler returns? and operation-level hooks (`on_operation_*`, `on_operation_attempt_*`) are dispatched asynchronously?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+    def _extract_error(operation: Operation):
+        if operation.step_details and operation.step_details.error:
+            return operation.step_details.error
+        if operation.callback_details and operation.callback_details.error:
+            return operation.callback_details.error
+        if operation.chained_invoke_details and operation.chained_invoke_details.error:
+            return operation.chained_invoke_details.error
+        if operation.context_details and operation.context_details.error:
+            return operation.context_details.error
+        return None
```
yaythomas: In `state`, ` CheckpointedResult.create_from_operation(cls, operation)` already does something of the sort, so this here is replicating logic.  The existing logic extracts both result and error).


but `state` imports `plugin` and we'd like to avoid circular reference.

so we should probably refactor and move this existing logic so everything can use it.

it might be an idea to do this as a no-func change refactor seperately in a preceding PR first, but (I'm typing quickly here just to give the ide) something like:


```
   @dataclass(frozen=True)
   class Operation:
       ...

       def get_error(self) -> ErrorObject | None:
           match self.operation_type:
               case OperationType.STEP:
                   return self.step_details.error if self.step_details else None
               case OperationType.CALLBACK:
                   return self.callback_details.error if self.callback_details else None
               case OperationType.CHAINED_INVOKE:
                   return self.chained_invoke_details.error if self.chained_invoke_details else None
               case OperationType.CONTEXT:
                   return self.context_details.error if self.context_details
               else None
                   return None

       def get_result(self) -> str | None:
           ... 
```

and then in CheckpointedResult:

```python
   return cls(
       operation=operation,
       status=operation.status,
       result=operation.get_result(),
       error=operation.get_error(),
   )

```

and here in plugin:

```
   OperationEndInfo(
       **params,
       ...,
       error=operation.get_error(),
   )

```

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+
+    @staticmethod
```
yaythomas: rather than have this ambient bit of unattached logic, we can keep it with the other `is_succeeded`, `is_failed` logic on the `Operation` object, so we have

```
   @dataclass(frozen=True)
   class Operation:
       _TERMINAL_STATUSES: ClassVar[frozenset[OperationStatus]] = frozenset({
           OperationStatus.SUCCEEDED,
           OperationStatus.FAILED,
           OperationStatus.CANCELLED,
           OperationStatus.TIMED_OUT,
           OperationStatus.STOPPED,
       })

       # ... fields ...

       def is_terminal(self) -> bool:
           """Return True if this operation is in a terminal state."""
           return self.status in Operation._TERMINAL_STATUSES

```

or maybe on the enum itself, but I'm not super fond of the forward reference lol:

```
  class OperationStatus(Enum):
      STARTED = "STARTED"
      PENDING = "PENDING"
      READY = "READY"
      SUCCEEDED = "SUCCEEDED"
      FAILED = "FAILED"
      CANCELLED = "CANCELLED"
      TIMED_OUT = "TIMED_OUT"
      STOPPED = "STOPPED"

      def is_terminal(self) -> bool:
          return self in _TERMINAL_STATUSES # this would be a fwd ref

```

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+        )
+
+        if update.operation_type is OperationType.STEP:
+            attempt = (
+                operation.step_details.attempt
+                if operation and operation.step_details
+                else 1
+            )
+            self.execute_plugins(
+                AttemptStartInfo(
```
yaythomas: is there an  OperationStartInfo and a AttemptStartInfo for each attempt?

from https://github.com/aws/aws-durable-execution-sdk-python/issues/370
>  - onOperationStart, onOperationEnd, onOperationAttemptStart,
   onOperationAttemptEnd fire at most once per operation/attempt.
>  - onOperationAttemptStart and onOperationAttemptEnd will manage a span that is a child of the span managed by onOperationStart and onOperationEnd. That  is retry attempts for steps and wait-for-condition operations.


I could easily be mistracing this, is this what happens? Maybe it's an artifact of the previous PR #381 introducing the READY logic so every step retry checkpoints START?

1. operation_start       (attempt 1 START)
2. attempt_start=1       (attempt 1 START)
3. attempt_end=1         (RETRY → PENDING)
4. operation_start        (attempt 2 START, across invocations)
5. attempt_start         (attempt 2 START)
6. attempt_end=2         (SUCCEED)
7. operation_end

might be an idea to introduce a functional test to verify/confirm the behaviour.

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+                    operation_id=update.operation_id,
+                    operation_type=update.operation_type,
+                    sub_type=update.sub_type,
+                    name=update.name,
+                    parent_id=update.parent_id,
+                    start_timestamp=datetime.datetime.now(),
```
yaythomas: utc

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+                    raise ValueError(f"Unknown info type: {type(info)}")
+        except Exception:
+            # log and ignore the exception
+            logger.exception("Plugin %s exception ignored", plugin.__class__.__name__)
+
+    def execute_plugins(self, info):
+        if not self.plugins:
+            return
+        for plugin in self.plugins:
+            future = self._executor.submit(self._dispatch_plugin, plugin, info)
```
yaythomas: this means strictly speaking plugins could execute out of order.

at present I don't think the design covers ordering or preserving it explicitly? But I'm thinking specifically for parent -> child relationships, order might well be important here?

It looks like the js runs synchronously on the handler thread (i.e sequentially) https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/utils/plugin/plugin-runner.ts

i.e js preserves ordering

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+            self._pending_futures.append(future)
```
yaythomas: doesn't this need a lock?

`_pending_futures` mutates from the main thread (for on_invocation_start/end), and on the background thread (on_operation_action etc.), and flush() iterates + clear non atomically.

but I'm also not sure to what degree this goes away depending on where we end up for synchronous/ordering, is the tracking list necessary?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+            if self.plugins
+            else None
+        )
+
+    def close(self) -> None:
+        """Shut down the thread pool, waiting for pending tasks to complete."""
+        if self._executor is None:
+            return
+        self.flush()
```
yaythomas: doesn't the next line shutdown(wait=True) effectively to the flush?

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

> if wait is True then this method will not return until all the pending futures are done executing and the resources associated with the executor have been freed. .. Regardless of the value of wait, the entire Python program will not exit until all pending futures are done executing.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -230,43 +242,56 @@ def __init__(
...
+            initial_checkpoint_token=self._invocation_input.checkpoint_token,
+            operations={},
+            service_client=self._service_client,
+            replay_status=ReplayStatus.NEW,
+        )
+
+    @staticmethod
+    def _print_durable_execution_arn(event):
+        try:
+            arn = event.get("DurableExecutionArn")
```
yaythomas: typings please.

And typings would've warned that `DurableExecutionInvocationInputWithClient` will throw on .get

prior comment in PR #378 asked why this moves to init. https://github.com/aws/aws-durable-execution-sdk-python/pull/378#discussion_r3192006599

to avoid the issue, if there were a factory method here that could use the regularized/sanitized input _after_ parsing happened then this also would've been an issue avoided, rather than re-doing the work in init.

Please see further down for suggested factory refactor that avoids this problem.

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+
+if TYPE_CHECKING:
+    from aws_durable_execution_sdk_python.execution import (
+        DurableExecutionInvocationOutput,
+    )
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
```
yaythomas: these can be frozen?

src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,351 @@
...
+class ExecutionEndInfo(ExecutionStartInfo):
+    status: InvocationStatus = InvocationStatus.SUCCEEDED
+    end_timestamp: datetime.datetime | None = None
+    error: ErrorObject | None = None
+
+
+class DurableExecutionPlugin(ABC):
```
yaythomas: how does context extractor work? 

there is no `context_extractor` parameter on `@durable_execution` or here.

so how will the plugins get the values for traceid/parent? is the idea to use the $ENVs, but the context_extractor interface makes me think that's not the idea?

https://github.com/aws/aws-durable-execution-sdk-js/blob/otel-instrumentation-design-v2/packages/aws-durable-execution-sdk-js/src/documents/PYTHON_SDK_OTEL_PLUGIN.md#context-propagation--configurable-context_extractor


this might be unimplemented in the current JS implementation also, we should align and disambiguate the plan here.

maybe?
1. Add lambda_context: LambdaContext | None to InvocationStartInfo    2. Add context_extractor parameter to @durable_execution
3. Call context_extractor in the executor before on_invocation_start fires and attach/detach the returned OTel Context around the invocation 

the extractor must receive `lambda_context` and fire per-invocation.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -476,32 +485,48 @@ def _handle_large_error(
...
             # before the execution completes.
-            execution_state.create_checkpoint_sync(failed_operation)
+            self._execution_state.create_checkpoint_sync(failed_operation)
 
             # return fail without an ErrorObject
             return None
 
         return error
 
+    def _handle_execution_exception(self, exception: Exception, retryable: bool):
```
yaythomas: general comment `on _handle_execution_output` and `_handle_execution_exception`, and [continuing on comment from previous PR](https://github.com/aws/aws-durable-execution-sdk-python/pull/375#discussion_r3192103864), these are in danger of being/becoming God Methods (https://wiki.c2.com/?GodMethod).

there's at least 
1. clear bug with the getattr, that we could have avoided, see further down
2. misuse potential on _handle_execution_output with constructions like `{Status: FAILED, Result: "x", Error: {...}}` if misused.
3. the "make type checker happy" raise Exceptions ibid, like
```
  # add a redundant raise to make type checker happy
   raise ExecutionError(msg)"
```

Please consider whether we might benefit by narrowing this down so that we follow the Single Responsiblity Principle and make each method responsible for one thing.
 
Please also note the 'NoReturn` avoiding the redundant raise.

Please note this also avoids the recursion between `_handle_execution_output` and `_handle_execution_exception`.

```python
from typing import NoReturn

  def _exit_success(self, result: str) -> MutableMapping[str, Any]:
      """End the invocation with a SUCCEEDED response."""
      output = DurableExecutionInvocationOutput(
          status=InvocationStatus.SUCCEEDED, result=result
      )
      self._plugin_executor.on_invocation_end(
          self._invocation_input.durable_execution_arn,
          self._context,
          self._execution_state.get_execution_operation(),
          output,
      )
      return output.to_dict()


  def _exit_pending(self) -> MutableMapping[str, Any]:
      """End the invocation with a PENDING response (execution suspended)."""
      output = DurableExecutionInvocationOutput(status=InvocationStatus.PENDING)
      self._plugin_executor.on_invocation_end(
          self._invocation_input.durable_execution_arn,
          self._context,
          self._execution_state.get_execution_operation(),
          output,
      )
      return output.to_dict()


  def _exit_failed(self, error: ErrorObject | None) -> MutableMapping[str, Any]:
      """End the invocation with a FAILED response.

      `error` may be None when a large error payload has been checkpointed
      separately (see `_handle_large_error`).
      """
      output = DurableExecutionInvocationOutput(
          status=InvocationStatus.FAILED, error=error
      )
      self._plugin_executor.on_invocation_end(
          self._invocation_input.durable_execution_arn,
          self._context,
          self._execution_state.get_execution_operation(),
          output,
      )
      return output.to_dict()


  def _notify_retry(self, exception: Exception) -> None:
      """Fire on_invocation_end(RETRY) for plugin observability, without
  raising.

      Callers are responsible for raising afterwards (typically via the caller's
      own except-block context so that exception chaining is explicit).
      """
      self._plugin_executor.on_invocation_end(
          self._invocation_input.durable_execution_arn,
          self._context,
          self._execution_state.get_execution_operation(),
          DurableExecutionInvocationOutput(
              status=InvocationStatus.RETRY,
              error=ErrorObject.from_exception(exception),
          ),
      )


  def _exit_retry_or_fail(
      self, exception: Exception, retryable: bool
  ) -> MutableMapping[str, Any]:
      """For caught exceptions with optional retry semantics.

      If `retryable`, notifies the plugin with RETRY and raises `exception`.
      Callers that want explicit `raise X from Y` chaining should use
      `_notify_retry(...)` + an explicit `raise ... from ...` instead.

      If not `retryable`, returns a FAILED response containing the exception.
      """
      if not retryable:
          return self._exit_failed(ErrorObject.from_exception(exception))
      self._notify_retry(exception)
      raise exception
```


Post-rewrite call sites — just the exits

```python
   # _parse_invocation_input  (pre-init bad payload)
   raise ExecutionError(msg) from e

   # execute()  (BotoClientError during initial state fetch)
   return self._exit_retry_or_fail(e, retryable=e.is_retryable())

   # _handle_user_future_result
   #   success
   return self._exit_success(serialized_result)

   #   BackgroundThreadError, non-retryable
   return self._exit_failed(ErrorObject.from_exception(cause))

   #   BackgroundThreadError, retryable  (explicit chaining preserved)
   self._notify_retry(cause)
   raise cause from bg_error

   #   SuspendExecution
   return self._exit_pending()

   #   CheckpointError
   return self._exit_retry_or_fail(e, retryable=e.is_retryable())

   #   InvocationError
   return self._exit_retry_or_fail(e, retryable=e.is_retryable())

   #   ExecutionError  (always non-retryable)
   return self._exit_failed(ErrorObject.from_exception(e))

   #   Exception → inner CheckpointError (large-error path failed)
   return self._exit_retry_or_fail(
       checkpoint_error, retryable=checkpoint_error.is_retryable()
   )

   #   Exception → user error
   return self._exit_failed(error)

   # _get_input_event  (JSONDecodeError)
   self._exit_retry(e)   # NoReturn — no `return` needed
```

yaythomas (changes requested):

src/aws_durable_execution_sdk_python/execution.py
yaythomas: b) also has benefit that this also lets us drop self._current_checkpoint_token. It's set once at init and only read at the top of checkpoint_batches_forever, and it's confusingly named (pre-existing issue).

https://github.com/aws/aws-durable-execution-sdk-python/pull/378#discussion_r3192447405

yaythomas (dismissed):

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: plugin_executor parameter gets lost when rebase code

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
zhongkechen: plugin_executor parameter gets lost when rebase code

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Will leave option b for future improvement. There are more higher priority tasks

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: the other status isn't needed

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: I'll revert the previous PR #375  and focus on plugin and leave the refactoring for future.

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: we can add more when needed

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: #370

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: you're right, it's not necessary

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: yes

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Removed

tests/plugin_test.py
zhongkechen: Removed ABC

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: no manual flush required except in test cases

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Will leave optimization and refactoring for future.

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: We have a test case `test_step_creates_start_checkpoint_when_status_is_ready` to verify this behavior. Feel free to create an issue if this isn't enough.

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: limited `max_workers` to 1 for now. May require some extra work if we want a different threading model

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Yes, it does

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Removed the refactoring code in execution.py

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: removed `_pending_futures`

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Will move if it's reused somewhere else.

src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Probably need to add a context parameter to  on_invocation_start (or all methods). It doesn't worth adding a new parameter specifically for OTel

zhongkechen:
The current implementation does not support the operations that are completed when the execution is sleeping.

This is blocked by another issue #389

zhongkechen:
This is no longer blocked by #389 as we are going to seek a different backend based approach.

.github/hooks/pre-commit
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
SilanHe, line 1: nit: file name could be pre-commit.sh

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -903,3 +922,38 @@ def _calculate_operation_size(queued_op: QueuedOperation) -> int:
...
+            except Exception as e:
+                self._plugin_executor.on_user_function_end(
+                    start_info, ErrorObject.from_exception(e)
+                )
+                raise
+
+        return wrapper
+
+    def on_operation_update(self, operation: Operation | None):
+        self._plugin_executor.on_operation_update(operation)
```
SilanHe: nit: is this EOF? should probably include whitespace

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -903,3 +922,38 @@ def _calculate_operation_size(queued_op: QueuedOperation) -> int:
...
+                raise
+            except Exception as e:
+                self._plugin_executor.on_user_function_end(
+                    start_info, ErrorObject.from_exception(e)
+                )
+                raise
+
+        return wrapper
+
+    def on_operation_update(self, operation: Operation | None):
```
zhongkechen: This can be removed.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -0,0 +1,402 @@
+import contextlib
```
SilanHe, line 1: does Python have a way to indicate that the feature is experimental?

.github/hooks/pre-commit
zhongkechen: `pre-commit` is the file name of git hook. We can create a link to it in the `.git/hooks/` folder. `pre-commit.sh` wouldn't work if we don't rename it.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
zhongkechen: Yes, added in the next commit: https://github.com/aws/aws-durable-execution-sdk-python/compare/6cc6584a7cb6a5c28e0f0e9b7b99834388508c59..e0c1638d704f26fce147e607daaf801529589ae5

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
zhongkechen: removed

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
zhongkechen: Removed this method

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -105,6 +128,70 @@ class OperationSubType(Enum):
     CHAINED_INVOKE = "ChainedInvoke"
 
 
+class InvocationStatus(Enum):
```
SilanHe, line 131: why did we move this and DurableExecutionInvocationOutput?

zhongkechen: Avoid cyclic dependencies

.github/hooks/pre-commit
SilanHe: so it triggers automatically on `git commit`?

SilanHe:
can you update the description of the PR to detail what exactly we're introducing in this PR? I'm finding it hard to review. Perhaps I'm misundertstanding but I'm not seeing how wait, invoke and callbacks plugin hooks are wired while it seems like the user function for child context, step and wait-for-condition  plugin hooks are wired.

.github/hooks/pre-commit
SilanHe: is this what you're introducing? https://pre-commit.com/

I'm slightly confused, this isn't located in `.git/hooks/` folder. Will it even work as is?

zhongkechen: yeah, after you run `ln -s .github/hooks/pre-commit .git/hooks/`

zhongkechen: No, https://pre-commit.com/ will do more than this but we'll need to add an additional dependency

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
```diff
@@ -154,9 +154,15 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
             self.operation_identifier.operation_id,
             self.operation_identifier.name,
         )
-
         try:
-            raw_result: T = self.func()
+            # todo: fix attempt (checkpointed_result.is_existent is always True)
```
zhongkechen, line 158: refactoring required to implement this: https://github.com/aws/aws-durable-execution-sdk-python/issues/414

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -306,6 +280,13 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
             ) as executor,
             contextlib.closing(execution_state) as execution_state,
         ):
+            # execute the plugins
+            plugin_executor.on_invocation_start(
```
SilanHe, line 284: where is on_invocation_end?

zhongkechen: in `handle_durable_output`

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -689,12 +702,18 @@ def checkpoint_batches_forever(self) -> None:
...
                         output.checkpoint_token,
                         output.new_execution_state.next_marker,
                     )
 
+                    for update in updates:
+                        self._plugin_executor.on_operation_action(update)
+
+                    for operation in updated_operations:
+                        self._plugin_executor.on_operation_update(operation)
+
```
SilanHe, line 716: what about updates that happen between invokes?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -149,77 +156,43 @@ def from_durable_execution_invocation_input(
...
+    if plugins:
+        warnings.warn(
+            "The 'plugins' parameter is provisional and may be altered or removed.",
+            category=FutureWarning,
+            stacklevel=2,  # point the warning to the caller of durable_execution
+        )
+
+    plugin_executor = PluginExecutor(plugins)
+
+    @plugin_executor.handle_durable_output
```
SilanHe, line 195: where is this (`handle_durable_output`) defined?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
SilanHe: where is it defined? it's not in the diff

zhongkechen: It's in the diff. You probably forgot to click `Load Diff` link

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
zhongkechen: That's not implemented yet. I'm waiting for the backend changes

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/execution.py
zhongkechen: it's of course in `PluginExecutor` class

SilanHe: we spoke offline, Frank pointed out it's in the plugin file but it was hidden because the change is large

SilanHe (approved):

zhongkechen:
Merged the PR to unblock the otel plugin feature while we still accept new comments/issues with the implementation as this is till an experimental feature.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -903,3 +922,35 @@ def _calculate_operation_size(queued_op: QueuedOperation) -> int:
...
         self.stop_checkpointing()
+
+    def wrap_user_function(
+        self,
+        user_function: Callable,
+        operation_identifier: OperationIdentifier,
+        is_replay_children: bool = False,
+        attempt: int | None = None,
+    ):
+        @functools.wraps(user_function)
```
SilanHe, line 933: so this ensures that the wrapper works even with concurrent operations?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
SilanHe: is this done now? I'm not seeing a commit in our history that seems to be associated

PR #375, opened 2026-05-04
[refactor]: use an executor class instead of closure

zhongkechen:
*Issue #, if available:*

Useful for #370 

*Description of changes:*

Added `DurableExecutionExecutor` class to wrap the required fields that were previously shared via closure. 
- This would make it easier to add more fields later like plugins.
- This would make it easier to split the big method into smaller pieces

plugins and more fields will be added to this new class to be shared across methods.

### DIFF
--- a/src/aws_durable_execution_sdk_python/execution.py
+++ b/src/aws_durable_execution_sdk_python/execution.py
@@ -7,14 +7,13 @@ import logging
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from enum import Enum
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast, Callable
 
 from aws_durable_execution_sdk_python.context import DurableContext
 from aws_durable_execution_sdk_python.exceptions import (
     BackgroundThreadError,
     BotoClientError,
     CheckpointError,
-    DurableExecutionsError,
     ExecutionError,
     InvocationError,
     SuspendExecution,
@@ -24,13 +23,12 @@ from aws_durable_execution_sdk_python.lambda_service import (
     ErrorObject,
     LambdaClient,
     Operation,
-    OperationType,
     OperationUpdate,
 )
 from aws_durable_execution_sdk_python.state import ExecutionState, ReplayStatus
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, MutableMapping
+    from collections.abc import MutableMapping
 
     from mypy_boto3_lambda import LambdaClient as Boto3LambdaClient
 
@@ -51,9 +49,9 @@ class InitialExecutionState:
 
     @staticmethod
     def from_dict(input_dict: MutableMapping[str, Any]) -> InitialExecutionState:
-        operations = []
-        if input_operations := input_dict.get("Operations"):
-            operations = [Operation.from_dict(op) for op in input_operations]
+        operations = [
+            Operation.from_dict(op) for op in input_dict.get("Operations", [])
+        ]
         return InitialExecutionState(
             operations=operations,
             next_marker=input_dict.get("NextMarker", ""),
@@ -61,9 +59,9 @@ class InitialExecutionState:
 
     @staticmethod
     def from_json_dict(input_dict: MutableMapping[str, Any]) -> InitialExecutionState:
-        operations = []
-        if input_operations := input_dict.get("Operations"):
-            operations = [Operation.from_json_dict(op) for op in input_operations]
+        operations = [
+            Operation.from_json_dict(op) for op in input_dict.get("Operations", [])
+        ]
         return InitialExecutionState(
             operations=operations,
             next_marker=input_dict.get("NextMarker", ""),
@@ -199,11 +197,6 @@ class DurableExecutionInvocationOutput:
 
         return result
 
-    @classmethod
-    def create_succeeded(cls, result: str) -> DurableExecutionInvocationOutput:
-        """Create a succeeded invocation output."""
-        return cls(status=InvocationStatus.SUCCEEDED, result=result)
-
 
 # endregion Invocation models
 
@@ -217,54 +210,89 @@ def durable_execution(
     if func is None:
         logger.debug("Decorator called with parameters")
         return functools.partial(durable_execution, boto3_client=boto3_client)
+    else:
+        logger.debug("Starting durable execution handler...")
 
-    logger.debug("Starting durable execution handler...")
+        def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
+            executor = DurableExecutionExecutor(
+                cast(Callable[[Any, DurableContext], Any], func),
+                boto3_client,
+                event,
+                context,
+            )
+            return executor.execute()
 
-    def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
-        invocation_input: DurableExecutionInvocationInput
-        service_client: DurableServiceClient
+        return wrapper
 
+
+class DurableExecutionExecutor:
+    def __init__(
+        self,
+        func: Callable[[Any, DurableContext], Any],
+        boto3_client: Boto3LambdaClient | None,
+        event: Any,
+        context: LambdaContext,
+    ):
+        self.func = func
+        self.boto3_client = boto3_client
+        self.event = event
+        self.context = context
+        self.invocation_input = self._parse_invocation_input(event)
+        self.service_client = self._parse_service_client(event, boto3_client)
+
+    def _parse_invocation_input(self, event: Any) -> DurableExecutionInvocationInput:
         # event likely only to be DurableExecutionInvocationInputWithClient when directly injected by test framework
+        invocation_input: (
+            DurableExecutionInvocationInputWithClient | DurableExecutionInvocationInput
+        )
         if isinstance(event, DurableExecutionInvocationInputWithClient):
-            logger.debug("durableExecutionArn: %s", event.durable_execution_arn)
             invocation_input = event
-            service_client = invocation_input.service_client
         else:
             try:
-                logger.debug(
-                    "durableExecutionArn: %s", event.get("DurableExecutionArn")
-                )
                 invocation_input = DurableExecutionInvocationInput.from_json_dict(event)
-            except (KeyError, TypeError, AttributeError) as e:
+            except (KeyError, TypeError, AttributeError):
                 msg = (
                     "Unexpected payload provided to start the durable execution. "
                     "Check your resource configurations to confirm the durability is set."
                 )
-                raise ExecutionError(msg) from e
+                # throws ExecutionError to terminate the invocation
+                self._handle_execution_output(
+                    exception=ExecutionError(msg), retryable=True
+                )
+                # add a redundant raise to make type checker happy
+                raise ExecutionError(msg)
 
+        logger.debug("durableExecutionArn: %s", invocation_input.durable_execution_arn)
+        return invocation_input
+
+    @staticmethod
+    def _parse_service_client(event, boto3_client):
+        if isinstance(event, DurableExecutionInvocationInputWithClient):
+            return event.service_client
+        elif boto3_client:
+            return LambdaClient(boto3_client)
+        else:
             # Use custom client if provided, otherwise initialize from environment
-            service_client = (
-                LambdaClient(client=boto3_client)
-                if boto3_client is not None
-                else LambdaClient.initialize_client()
-            )
+            return LambdaClient.initialize_client()
 
+    def execute(self):
         execution_state: ExecutionState = ExecutionState(
-            durable_execution_arn=invocation_input.durable_execution_arn,
-            initial_checkpoint_token=invocation_input.checkpoint_token,
+            durable_execution_arn=self.invocation_input.durable_execution_arn,
+            initial_checkpoint_token=self.invocation_input.checkpoint_token,
             operations={},
-            service_client=service_client,
+            service_client=self.service_client,
             # If there are operations other than the initial EXECUTION one, current state is in replay mode
+            # todo: replay status will be wrong if initial_execution_state contains only one operation and more in next pages
             replay_status=ReplayStatus.REPLAY
-            if len(invocation_input.initial_execution_state.operations) > 1
+            if len(self.invocation_input.initial_execution_state.operations) > 1
             else ReplayStatus.NEW,
         )
 
         try:
             execution_state.fetch_paginated_operations(
-                invocation_input.initial_execution_state.operations,
-                invocation_input.checkpoint_token,
-                invocation_input.initial_execution_state.next_marker,
+                self.invocation_input.initial_execution_state.operations,
+                self.invocation_input.checkpoint_token,
+                self.invocation_input.initial_execution_state.next_marker,
             )
         except BotoClientError as e:
             # Non-retryable Durable API errors (e.g., customer configuration issues,
@@ -275,11 +303,9 @@ def durable_execution(
                     "without retry.",
                     extra=e.build_logger_extras(),
                 )
-                return DurableExecutionInvocationOutput(
-                    status=InvocationStatus.FAILED,
-                    error=ErrorObject.from_exception(e),
-                ).to_dict()
-            raise
+            return self._handle_execution_output(
+                exception=e, retryable=e.is_retryable()
+            )
 
         raw_input_payload: str | None = execution_state.get_input_payload()
 
@@ -289,15 +315,15 @@ def durable_execution(
         if raw_input_payload and raw_input_payload.strip():
             try:
                 input_event = json.loads(raw_input_payload)
-            except json.JSONDecodeError:
+            except json.JSONDecodeError as e:
                 logger.exception(
                     "Failed to parse input payload as JSON: payload: %r",
                     raw_input_payload,
                 )
-                raise
+                self._handle_execution_output(exception=e, retryable=True)
 
         durable_context: DurableContext = DurableContext.from_lambda_context(
-            state=execution_state, lambda_context=context
+            state=execution_state, lambda_context=self.context
         )
 
         # Use ThreadPoolExecutor for concurrent execution of user code and background checkpoint processing
@@ -312,13 +338,13 @@ def durable_execution(
 
             # Thread 2: Execute user function
             logger.debug(
-                "%s entering user-space...", invocation_input.durable_execution_arn
+                "%s entering user-space...", self.invocation_input.durable_execution_arn
             )
-            user_future = executor.submit(func, input_event, durable_context)
+            user_future = executor.submit(self.func, input_event, durable_context)
 
             logger.debug(
                 "%s waiting for user code completion...",
-                invocation_input.durable_execution_arn,
+                self.invocation_input.durable_execution_arn,
             )
 
             try:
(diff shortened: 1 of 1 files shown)


src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,165 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
+        exception: Exception | None = None,
+        retryable: bool = False,
+        status: InvocationStatus | None = None,
+    ) -> MutableMapping[str, Any]:
+        if exception:
+            if retryable:
+                # Throw the error to trigger Lambda retry
+                raise exception
+            else:
+                return self._handle_execution_output(
```
SilanHe, line 501: just curious, do python people generally consider this recursion cleaner than directly setting 

```
output = DurableExecutionInvocationOutput(
                status=InvocationStatus.FAILED, result=result, error=error
            )
```
?

zhongkechen: This block handles only `retryable` and `exception` parameters and the recursive call let the other parts handle the rest of the parameters.

SilanHe (approved):

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+        self.boto3_client = boto3_client
+        self.event = event
+        self.context = context
+        self.invocation_input = self._parse_invocation_input(event)
+        self.service_client = self._parse_service_client(event, boto3_client)
+
+    def _parse_invocation_input(self, event: Any) -> DurableExecutionInvocationInput:
         # event likely only to be DurableExecutionInvocationInputWithClient when directly injected by test framework
+        invocation_input: (
+            DurableExecutionInvocationInputWithClient | DurableExecutionInvocationInput
```
yaythomas, line 246: this is unnecessary? 

DurableExecutionInvocationInputWithClient is derived from DurableExecutionInvocationInput.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+                    exception=ExecutionError(msg), retryable=True
+                )
+                # add a redundant raise to make type checker happy
+                raise ExecutionError(msg)
+
+        logger.debug("durableExecutionArn: %s", invocation_input.durable_execution_arn)
+        return invocation_input
 
+    @staticmethod
+    def _parse_service_client(event, boto3_client):
```
yaythomas, line 269: please annotate with types.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+        else:
             # Use custom client if provided, otherwise initialize from environment
-            service_client = (
-                LambdaClient(client=boto3_client)
-                if boto3_client is not None
-                else LambdaClient.initialize_client()
-            )
+            return LambdaClient.initialize_client()
 
+    def execute(self):
```
yaythomas, line 278: typing for return type pls

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,166 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
-    )
+    @staticmethod
+    def _handle_large_error(
+        execution_state: ExecutionState, exception: Exception
+    ) -> ErrorObject | None:
+        # large response handling here. Remember if checkpointing to complete, NOT to include
+        # payload in response
+        error = ErrorObject.from_exception(exception)
+        serialized_error = json.dumps(error.to_dict())
+        if serialized_error and len(serialized_error) > LAMBDA_RESPONSE_SIZE_LIMIT:
```
yaythomas, line 470: this is a behaviour change. previously it would count the envelope also. 

the envelope is probably around ~30 bytes, but still, behaviour change.

- "Status":"FAILED" → ~18 bytes
- "Error": → 8 bytes
- - {} and commas → ~4 bytes

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
             invocation_input = event
-            service_client = invocation_input.service_client
         else:
             try:
-                logger.debug(
-                    "durableExecutionArn: %s", event.get("DurableExecutionArn")
-                )
                 invocation_input = DurableExecutionInvocationInput.from_json_dict(event)
-            except (KeyError, TypeError, AttributeError) as e:
+            except (KeyError, TypeError, AttributeError):
```
yaythomas, line 253: why remove the exception chaining? this loses valuable details on the cause of the error, e.g which field is malformed.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,166 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
+
+        return error
+
+    def _handle_execution_output(
+        self,
+        result: str | None = None,
+        error: ErrorObject | None = None,
+        exception: Exception | None = None,
+        retryable: bool = False,
+        status: InvocationStatus | None = None,
```
yaythomas, line 494: this method takes a bunch of overlapping params, with some unstated logic as to which combinations of these are valid. 

maybe create?

```python
   @staticmethod
   def _exit_success(result: str) -> MutableMapping[str, Any]:
       """End the invocation with a SUCCEEDED response."""
       return DurableExecutionInvocationOutput(
           status=InvocationStatus.SUCCEEDED, result=result
       ).to_dict()

   @staticmethod
   def _exit_pending() -> MutableMapping[str, Any]:
       """End the invocation with a PENDING response (execution suspended)."""
       return DurableExecutionInvocationOutput(
           status=InvocationStatus.PENDING
       ).to_dict()

   @staticmethod
   def _exit_failed(error: ErrorObject | None = None) -> MutableMapping[str,
   Any]:
       """End the invocation with a FAILED response.

       `error=None` when the error was too large and has been checkpointed
       durably instead of returned in the Lambda response envelope.
       """
       return DurableExecutionInvocationOutput(
           status=InvocationStatus.FAILED, error=error
       ).to_dict()

    @staticmethod
    def _exit_retry(self, exception: Exception) -> NoReturn:
        self._notify_retry(exception)
        raise exception

   @staticmethod
   def _exit_retry_or_fail(
       exception: Exception, retryable: bool
   ) -> MutableMapping[str, Any]:
       """End the invocation: raise to trigger a Lambda retry, or return a
   FAILED response.

       Callers must `return` the result — this method may return.
       """
       if retryable:
           raise exception
       return DurableExecutionInvocationOutput(
           status=InvocationStatus.FAILED,
           error=ErrorObject.from_exception(exception),
       ).to_dict()
```

yaythomas: and then each call site:

```python
   # success path
   return self._handle_execution_output(result=serialized_result)
   # →
   return self._exit_success(serialized_result)

   # SuspendExecution
   return self._handle_execution_output(status=InvocationStatus.PENDING)
   # →
   return self._exit_pending()

   # generic Exception with optional ErrorObject from _handle_large_error
   return self._handle_execution_output(status=InvocationStatus.FAILED,
   error=error)
   # →
   return self._exit_failed(error)

   # ExecutionError — always non-retryable
   return self._handle_execution_output(exception=e)
   # →
   return self._exit_failed(ErrorObject.from_exception(e))

   # BotoClientError / CheckpointError / InvocationError / BackgroundThreadError
   return self._handle_execution_output(exception=e, retryable=e.is_retryable())
   # →
   return self._exit_retry_or_fail(e, retryable=e.is_retryable())

   # JSONDecodeError — always-raise case, drop the helper entirely
   self._handle_execution_output(exception=e, retryable=True)
   # →
   raise   # bare re-raise keeps the traceback

   # _parse_invocation_input bad payload — always-raise case, drop the helper
   self._handle_execution_output(exception=ExecutionError(msg), retryable=True)
   raise ExecutionError(msg)   # redundant
   # →
   raise ExecutionError(msg) from e   # direct, preserves the cause

   # inner CheckpointError inside `except Exception` — rename to avoid shadowing
   except CheckpointError as e:
       return self._handle_execution_output(exception=e,
   retryable=e.is_retryable())
   # →
   except CheckpointError as ce:
       return self._exit_retry_or_fail(ce, retryable=ce.is_retryable())
```

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
                     "Unexpected payload provided to start the durable execution. "
                     "Check your resource configurations to confirm the durability is set."
                 )
-                raise ExecutionError(msg) from e
+                # throws ExecutionError to terminate the invocation
+                self._handle_execution_output(
+                    exception=ExecutionError(msg), retryable=True
+                )
+                # add a redundant raise to make type checker happy
+                raise ExecutionError(msg)
```
yaythomas, line 263: nervous about this. if _handle_execution_output behaves surprisingly, this ends up becoming a problem.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -289,15 +315,15 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
             try:
                 input_event = json.loads(raw_input_payload)
-            except json.JSONDecodeError:
+            except json.JSONDecodeError as e:
                 logger.exception(
                     "Failed to parse input payload as JSON: payload: %r",
                     raw_input_payload,
                 )
-                raise
+                self._handle_execution_output(exception=e, retryable=True)
```
yaythomas, line 323: this could cause surprises with JsonDecodeError getting lost depending on what happens in _handle_execution_output, which will help with this.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,166 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
             except Exception as e:
                 # all user-space errors go here
                 logger.exception("Execution failed")
 
-                result = DurableExecutionInvocationOutput(
-                    status=InvocationStatus.FAILED, error=ErrorObject.from_exception(e)
-                ).to_dict()
+                try:
+                    error = self._handle_large_error(execution_state, exception=e)
+                except CheckpointError as e:
```
yaythomas, line 428: variable shadowing. maybe alias differently for clarity.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+                raise ExecutionError(msg)
+
+        logger.debug("durableExecutionArn: %s", invocation_input.durable_execution_arn)
+        return invocation_input
 
+    @staticmethod
+    def _parse_service_client(event, boto3_client):
+        if isinstance(event, DurableExecutionInvocationInputWithClient):
+            return event.service_client
+        elif boto3_client:
```
yaythomas, line 272: probably doesn't matter too much, but in combination with no typing here it might...

original code had stronger check for `if boto3_client is not None`

src/aws_durable_execution_sdk_python/execution.py
yaythomas: the refactor I'm suggesting gets rid of the recursion.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,166 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
                 else:
                     logger.exception("Checkpoint processing failed")
-                raise bg_error.source_exception from bg_error
+
+                retryable = (
+                    not isinstance(cause, BotoClientError) or cause.is_retryable()
+                )
+                return self._handle_execution_output(
+                    exception=cause, retryable=retryable
+                )
```
yaythomas, line 389: this is losing the `from bg_error` error chaining

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+        boto3_client: Boto3LambdaClient | None,
+        event: Any,
+        context: LambdaContext,
+    ):
+        self.func = func
+        self.boto3_client = boto3_client
+        self.event = event
+        self.context = context
+        self.invocation_input = self._parse_invocation_input(event)
+        self.service_client = self._parse_service_client(event, boto3_client)
```
yaythomas, line 241: as a general pattern, prefer doing work like this in a `from_event()` factory, and then the init just takes plain args rather than call factories.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -328,143 +354,166 @@ def wrapper(event: Any, context: LambdaContext) -> MutableMapping[str, Any]:
...
-                    ).to_dict()
-
-                return result
-
-    return wrapper
-
-
-def handle_checkpoint_error(error: CheckpointError) -> DurableExecutionInvocationOutput:
-    if error.is_retryable():
-        raise error from None  # Terminate Lambda immediately and have it be retried
```
yaythomas, line 467: I think this from None is also a behaviour change. 

This one was intentional?

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -217,54 +210,89 @@ def durable_execution(
...
+            DurableExecutionInvocationInputWithClient | DurableExecutionInvocationInput
+        )
         if isinstance(event, DurableExecutionInvocationInputWithClient):
-            logger.debug("durableExecutionArn: %s", event.durable_execution_arn)
             invocation_input = event
-            service_client = invocation_input.service_client
         else:
             try:
-                logger.debug(
-                    "durableExecutionArn: %s", event.get("DurableExecutionArn")
```
yaythomas, line 235: this would log _before_ attempting to parse, so if something goes wrong at least maybe the Arn is in the output. Now it only logs on success?

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Removed in the latest ongoing branch: https://github.com/aws/aws-durable-execution-sdk-python/blob/eb9a080f5d2acddd1b22686767184eeb56055926/src/aws_durable_execution_sdk_python/execution.py#L502

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: will remove the declaration

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: This aligns with successful output where we check only result size

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: No difference between them

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: `from bg_error` is the default behavior if omitted

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: This might be a real concern. From None or bare raise should be used, otherwise the exception will be chained with redundant info.

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: exception chain is the default behavior

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: the latest ongoing version is better:
https://github.com/aws/aws-durable-execution-sdk-python/blob/255afb676b7ebe36b61633a7ef3bf21745094f67/src/aws_durable_execution_sdk_python/execution.py#L305C16-L308

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: This isn't a factory of a single type. The service_client might be different types.

yaythomas: yes, in which case a `create()` style factory can be useful, per style guide in contributing https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#initialization-and-conversion

src/aws_durable_execution_sdk_python/execution.py
yaythomas: there is a difference, depending on what type `boto3_client` is.

src/aws_durable_execution_sdk_python/execution.py
yaythomas: true, but it becomes implicit rather than explicit.

zhongkechen: This is much nicer than previously handling exceptions in many different places.

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Whatever the type is, `if variable` is the Python recommended way of checking variables.

yaythomas: not quite settled as to recommended.
https://github.com/astral-sh/ruff/discussions/8757

There is currently no typing on boto3_client. It can be an int,  it can be a bool, it can be surprising.

 Regardless, probably doesn't matter too much here :-)

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Factory methods pattern is rarely used in python. A single method `__init__` is flexible enough as an entrance  to handle all object initialization scenarios. What was the reason this library use so many  factory methods?

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Fixed in the other PR: https://github.com/aws/aws-durable-execution-sdk-python/blob/5095e67d2d89e6ec3ad386801642213a67f5b408/src/aws_durable_execution_sdk_python/execution.py#L259-L265

src/aws_durable_execution_sdk_python/execution.py
zhongkechen: Fixed in the other PR: https://github.com/aws/aws-durable-execution-sdk-python/blob/5095e67d2d89e6ec3ad386801642213a67f5b408/src/aws_durable_execution_sdk_python/execution.py#L503-L505

src/aws_durable_execution_sdk_python/execution.py
yaythomas: I'm not so sure about that "rarely"... :-) 

Brett Slatkin's book
https://www.informit.com/articles/article.aspx?p=2992597&seqNum=3

More recently:
https://blog.glyph.im/2025/04/stop-writing-init-methods.html

The entire approach of `@dataclass`, which I think is fair to say is where Python has chosen to go for data modelling, defaulting to init for assignment. Boto Client constructor, Pandas dataframe...

Regardless, the prevailing standard in this repo, as documented in the style guide in CONTRIBUTING, is to keep init light and use factories for more advanced initialization.
https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#initialization-and-conversion

I think it is a very well established standard, also in Python, to be consistent with the surrounding code.

https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

PR #377, opened 2026-05-05
fix: replay status after paginated state

### LINKED ISSUE #376: [Bug]: wrong replay_status in edge cases
### Expected Behavior

replay_status is always set to NEW OR REPLAY properly

### Actual Behavior

In case `initial_execution_state` comes with one execution operation and other operations come in next pages, re-invocation of the execution will start mistakenly in NEW mode.

https://github.com/aws/aws-durable-execution-sdk-python/blob/main/src/aws_durable_execution_sdk_python/execution.py#L259

### Steps to Reproduce

N/A

### SDK Version

1.4

### Python Version

3.14

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

_No response_

tejakusireddy:
*Issue #, if available:*

Fixes #376

*Description of changes:*

This updates durable execution initialization so replay status is determined after paginated execution state has been fetched.

Previously, replay status was based only on the first page of `initial_execution_state.operations`. If that first page contained only the execution operation and prior operations were returned on later pages, the invocation could incorrectly start in `NEW` mode.

Changes:
- Initialize execution state as `NEW`
- Fetch paginated operations before deciding replay mode
- Mark execution state as replaying when fetched state contains prior operations
- Add a regression test for paginated initial state with prior operations

Tests:
- `hatch test tests/execution_test.py::test_durable_execution_replays_when_paginated_state_has_prior_operations`
- `hatch test tests/execution_test.py`
- `hatch run types:check`
- `hatch fmt --check`
- `.github/scripts/ci-checks.sh`

### DIFF
--- a/src/aws_durable_execution_sdk_python/execution.py
+++ b/src/aws_durable_execution_sdk_python/execution.py
@@ -254,10 +254,7 @@ def durable_execution(
             initial_checkpoint_token=invocation_input.checkpoint_token,
             operations={},
             service_client=service_client,
-            # If there are operations other than the initial EXECUTION one, current state is in replay mode
-            replay_status=ReplayStatus.REPLAY
-            if len(invocation_input.initial_execution_state.operations) > 1
-            else ReplayStatus.NEW,
+            replay_status=ReplayStatus.NEW,
         )
 
         try:
@@ -280,6 +277,9 @@ def durable_execution(
                     error=ErrorObject.from_exception(e),
                 ).to_dict()
             raise
+
+        if len(execution_state.operations) > 1:
+            execution_state.mark_replaying()
 
         raw_input_payload: str | None = execution_state.get_input_payload()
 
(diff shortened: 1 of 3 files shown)


src/aws_durable_execution_sdk_python/execution.py
zhongkechen, line 288: I'd recommend move this call to ExecutionState

tejakusireddy: Thanks, moved the replay-status decision into ExecutionState via mark_replaying_if_prior_operations_exist() and updated the execution wrapper to call it after paginated state fetch. Re-ran the focused regression test, execution_test.py, fmt, and types.

zhongkechen: I meant `fetch_paginated_operations` could be moved. We could do that in a follow-up

zhongkechen (approved):
Resolved the conflicts and pushed a commit. 

Approving now to avoid future conflicts. I'll follow up with some refactoring.

PR #378, opened 2026-05-05
refactor: move replay status init into ExecutionState

zhongkechen:
*Issue #, if available:*

- refactoring for #376 
- follow-up of #377 

*Description of changes:*

- Move replay status initialization into ExecutionState
- Move input parsing into ExecutionState
- Split giant `execute()` into smaller pieces

### DIFF
--- a/src/aws_durable_execution_sdk_python/context.py
+++ b/src/aws_durable_execution_sdk_python/context.py
@@ -42,7 +42,6 @@ from aws_durable_execution_sdk_python.serdes import (
     SerDes,
     deserialize,
 )
-from aws_durable_execution_sdk_python.state import ExecutionState  # noqa: TCH001
 from aws_durable_execution_sdk_python.threading import OrderedCounter
 from aws_durable_execution_sdk_python.types import Callback as CallbackProtocol
 from aws_durable_execution_sdk_python.types import (
@@ -58,6 +57,7 @@ from aws_durable_execution_sdk_python.types import (
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
 
+    from aws_durable_execution_sdk_python.state import ExecutionState  # noqa: TCH001
     from aws_durable_execution_sdk_python.concurrency.models import BatchResult
     from aws_durable_execution_sdk_python.state import CheckpointedResult
     from aws_durable_execution_sdk_python.types import LambdaContext
--- a/src/aws_durable_execution_sdk_python/execution.py
+++ b/src/aws_durable_execution_sdk_python/execution.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import contextlib
 import functools
 import json
 import logging
@@ -25,7 +24,7 @@ from aws_durable_execution_sdk_python.lambda_service import (
     Operation,
     OperationUpdate,
 )
-from aws_durable_execution_sdk_python.state import ExecutionState, ReplayStatus
+from aws_durable_execution_sdk_python.state import ExecutionState
 
 if TYPE_CHECKING:
     from collections.abc import MutableMapping
@@ -234,11 +233,11 @@ class DurableExecutionExecutor:
         context: LambdaContext,
     ):
         self.func = func
-        self.boto3_client = boto3_client
-        self.event = event
         self.context = context
         self.invocation_input = self._parse_invocation_input(event)
         self.service_client = self._parse_service_client(event, boto3_client)
+        self.durable_execution_arn = self.invocation_input.durable_execution_arn
+        logger.debug("durableExecutionArn: %s", self.durable_execution_arn)
 
     def _parse_invocation_input(self, event: Any) -> DurableExecutionInvocationInput:
         # event likely only to be DurableExecutionInvocationInputWithClient when directly injected by test framework
@@ -262,7 +261,6 @@ class DurableExecutionExecutor:
                 # add a redundant raise to make type checker happy
                 raise ExecutionError(msg)
 
-        logger.debug("durableExecutionArn: %s", invocation_input.durable_execution_arn)
         return invocation_input
 
     @staticmethod
@@ -275,20 +273,11 @@ class DurableExecutionExecutor:
             # Use custom client if provided, otherwise initialize from environment
             return LambdaClient.initialize_client()
 
-    def execute(self):
-        execution_state: ExecutionState = ExecutionState(
-            durable_execution_arn=self.invocation_input.durable_execution_arn,
-            initial_checkpoint_token=self.invocation_input.checkpoint_token,
-            operations={},
-            service_client=self.service_client,
-            replay_status=ReplayStatus.NEW,
-        )
-
+    def execute(self) -> MutableMapping[str, Any]:
         try:
-            execution_state.fetch_paginated_operations(
-                self.invocation_input.initial_execution_state.operations,
-                self.invocation_input.checkpoint_token,
-                self.invocation_input.initial_execution_state.next_marker,
+            execution_state: ExecutionState = ExecutionState(
+                invocation_input=self.invocation_input,
+                service_client=self.service_client,
             )
         except BotoClientError as e:
             # Non-retryable Durable API errors (e.g., customer configuration issues,
@@ -303,22 +292,12 @@ class DurableExecutionExecutor:
                 exception=e, retryable=e.is_retryable()
             )
 
-        execution_state.mark_replaying_if_prior_operations_exist()
-
-        raw_input_payload: str | None = execution_state.get_input_payload()
-
-        # Python RIC LambdaMarshaller just uses standard json deserialization for event
-        # https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/lambda_runtime_marshaller.py#L46
-        input_event: MutableMapping[str, Any] = {}
-        if raw_input_payload and raw_input_payload.strip():
-            try:
-                input_event = json.loads(raw_input_payload)
-            except json.JSONDecodeError as e:
-                logger.exception(
-                    "Failed to parse input payload as JSON: payload: %r",
-                    raw_input_payload,
-                )
-                self._handle_execution_output(exception=e, retryable=True)
+        try:
+            input_event: Any = execution_state.get_input_event()
+        except json.JSONDecodeError as e:
+            self._handle_execution_output(exception=e, retryable=True)
+            # add a redundant raise to make type checker happy
+            raise e
 
         durable_context: DurableContext = DurableContext.from_lambda_context(
             state=execution_state, lambda_context=self.context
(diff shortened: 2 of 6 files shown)


src/aws_durable_execution_sdk_python/context.py
```diff
@@ -58,6 +57,7 @@
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
 
+    from aws_durable_execution_sdk_python.state import ExecutionState  # noqa: TCH001
```
yaythomas, line 60: why's the noqa here?

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -25,7 +24,7 @@
     Operation,
     OperationUpdate,
 )
-from aws_durable_execution_sdk_python.state import ExecutionState, ReplayStatus
+from aws_durable_execution_sdk_python.state import ExecutionState
```
yaythomas, line 27: this is creating a circular reference. 

please see style guide in contributing: runtime environment variables

https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#organization

> Do not allow circular references, even if you can get away with it by using if TYPE_CHECKING.

Maybe move the `# region Invocation models` either to `lambda_service` , or maybe a new module `invocation`.

src/aws_durable_execution_sdk_python/execution.py
```diff
@@ -262,7 +261,6 @@ def _parse_invocation_input(self, event: Any) -> DurableExecutionInvocationInput
                 # add a redundant raise to make type checker happy
                 raise ExecutionError(msg)
 
-        logger.debug("durableExecutionArn: %s", invocation_input.durable_execution_arn)
```
yaythomas, line 265: why move this into init? The log reads as a diagnostic of the parsed invocation input, which is here.

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -309,6 +316,23 @@ def fetch_paginated_operations(
                         {op.operation_id: op for op in all_operations}
                     )
 
+    def get_input_event(self) -> Any:
```
yaythomas, line 319: why is this on ExecutionState? this is json-deserialzation and a concern of the `DurableExecutionInvocationInput`, this is the object that owns the payload.

```
   invocation_input.get_input_event()
```

suggestion: put on `DurableExecutionInvocationInput` pls.

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -682,11 +694,7 @@ def checkpoint_batches_forever(self) -> None:
                     current_checkpoint_token = output.checkpoint_token
 
                     # Fetch new operations from the API before unblocking sync waiters
-                    self.fetch_paginated_operations(
-                        output.new_execution_state.operations,
-                        output.checkpoint_token,
-                        output.new_execution_state.next_marker,
-                    )
+                    self.fetch_paginated_operations(output.new_execution_state)
```
yaythomas, line 697: this isn't using the new current_checkpoint_token? won't it end up using old checkpoint token?

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -284,13 +290,14 @@ def fetch_paginated_operations(
...
-            initial_operations.copy() if initial_operations else []
+            execution_state.operations.copy() if execution_state.operations else []
         )
+        next_marker = execution_state.next_marker
         try:
             while next_marker:
                 output: StateOutput = self._service_client.get_execution_state(
                     durable_execution_arn=self.durable_execution_arn,
-                    checkpoint_token=checkpoint_token,
+                    checkpoint_token=self._current_checkpoint_token,
```
yaythomas, line 300: isn't this going to be stale? 

if fetch_paginated, as it used to, takes
```
  initial_operations: list[Operation],
  checkpoint_token: str,
```

then the circular reference can also go away, I think.

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -258,24 +263,25 @@ def __init__(
 
         # Protects parent_to_children and parent_done
         self._parent_done_lock: Lock = Lock()
-        self._replay_status: ReplayStatus = replay_status
         self._replay_status_lock: Lock = Lock()
         self._visited_operations: set[str] = set()
 
+        # fetch initial operations
+        self.fetch_paginated_operations(invocation_input.initial_execution_state)
```
yaythomas, line 270: per style guide https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#initialization-and-conversion

> Class constructors must be light and not do more than initialize the
   class. … Use a @classmethod factory method instead to encapsulate more
   advanced logic. For example, if a class depends on logic that might fail,
   encapsulate this in a create classmethod.


suggestion:

```python
   @classmethod
   def from_invocation_input(
       cls,
       invocation_input: DurableExecutionInvocationInput,
       service_client: DurableServiceClient,
       batcher_config: CheckpointBatcherConfig | None = None,
   ) -> ExecutionState:
       state = cls(
           durable_execution_arn=invocation_input.durable_execution_arn,
           initial_checkpoint_token=invocation_input.checkpoint_token,
           service_client=service_client,
           batcher_config=batcher_config,
       )
       state.fetch_paginated_operations(
           initial_operations=invocation_input.initial_execution_state.operations,
           next_marker=invocation_input.initial_execution_state.next_marker,
           checkpoint_token=invocation_input.checkpoint_token,
       )
       # Init-time decision; no concurrent writers to operations,
       # no concurrent readers of _replay_status yet.
       if len(state.operations) > 1:
           state._replay_status = ReplayStatus.REPLAY
       return state
```

This assumes __init__ is slimmed back to primitives (takes durable_execution_arn: str, initial_checkpoint_token: str, service_client, batcher_config)

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -258,24 +263,25 @@ def __init__(
...
+        self._replay_status: ReplayStatus = (
+            ReplayStatus.REPLAY if len(self.operations) > 1 else ReplayStatus.NEW
```
yaythomas, line 272: no action note, just to capture thoughts during review for reference/future explorers.

this PR deletes the previous mark_replaying check entirely. 

```
   def mark_replaying_if_prior_operations_exist(self) -> None:
       """Mark execution state as replaying when non-execution operations
   exist."""
       with self._operations_lock:
           has_prior_operations: bool = any(
               op.operation_type is not OperationType.EXECUTION
               for op in self.operations.values()
           )
       if has_prior_operations:
           with self._replay_status_lock:
               self._replay_status = ReplayStatus.REPLAY

```

The new behaviour (`len(self.operations) > 1`) is similar check to what JS is doing. i.e it's _not_ explicitly checking for non execution nodes, just that there is >1. 

the locking is not necessary because it's in init.

tests/state_test.py
```diff
@@ -753,30 +775,29 @@ def mock_get_execution_state(durable_execution_arn, checkpoint_token, next_marke
...
+                ),
+            ),
+            service_client=mock_lambda_client,
         )
 
-    # Initial operation + page 1 should be stored despite page 2 failing
-    assert "0" in state.operations
-    assert "1" in state.operations
-    assert len(state.operations) == 2
+        # Initial operation + page 1 should be stored despite page 2 failing
```
yaythomas, line 797: these asserts are dead code now, inside the with?

yaythomas (changes requested):

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -258,24 +263,25 @@ def __init__(
...
     def fetch_paginated_operations(
         self,
-        initial_operations: list[Operation],
-        checkpoint_token: str,
-        next_marker: str | None,
+        execution_state: InitialExecutionState | CheckpointUpdatedExecutionState,
     ) -> None:
         """Add initial operations and fetch all paginated operations from the Durable Functions API. This method is thread_safe.
 
         The checkpoint_token is passed explicitly as a parameter rather than using the instance variable to ensure thread safety.
```
yaythomas, line 281: not true anymore. but if change back as I suggest further down, then still relevant :-)

yaythomas (changes requested):

src/aws_durable_execution_sdk_python/context.py
zhongkechen: No idea. Noqa above

yaythomas: https://docs.astral.sh/ruff/rules/typing-only-first-party-import/

yes, above it was relevant, because outside of a type checking block.

> Checks for first-party imports that are only used for type annotations, but aren't defined in a type-checking block.

yaythomas (changes requested):

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -225,16 +233,13 @@ class ExecutionState:
...
+        invocation_input: DurableExecutionInvocationInput,
         service_client: DurableServiceClient,
         batcher_config: CheckpointBatcherConfig | None = None,
-        replay_status: ReplayStatus = ReplayStatus.NEW,
     ):
-        self.durable_execution_arn: str = durable_execution_arn
-        self._current_checkpoint_token: str = initial_checkpoint_token
-        self.operations: MutableMapping[str, Operation] = operations
+        self.durable_execution_arn: str = invocation_input.durable_execution_arn
+        self._current_checkpoint_token: str = invocation_input.checkpoint_token
```
yaythomas, line 241: pre-existing issue, but current_checkpoint_token is not really accurate, it's actually _initial_checkpoint_token.

I'm actually thinking to drop it entirely, and make the _start_checkpointing take the current_token explicitly. Reduce the surface area for confusion arising.

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -894,5 +902,12 @@ def _calculate_operation_size(queued_op: QueuedOperation) -> int:
...
         return len(serialized)
 
-    def close(self):
-        self.stop_checkpointing()
+    @contextmanager
+    def start_checkpointing(self, executor: Executor):
+        # start checkpointing
+        executor.submit(self.checkpoint_batches_forever)
+        try:
+            yield self
```
yaythomas, line 910: nice idea :-) 

minor nit. the yield self is redundant, it only needs yield.

suggestion, make it take the checkpoint token:
```
   @contextmanager
   def start_checkpointing(
       self,
       executor: Executor,
       initial_checkpoint_token: str,
   ) -> Iterator[None]:
       executor.submit(self._checkpoint_batches_forever,
   initial_checkpoint_token)
       try:
           yield
       finally:
           self.stop_checkpointing()
```

src/aws_durable_execution_sdk_python/state.py
zhongkechen: Creation of the class and fetching the initial operations have to happen in two different places due to the difference in error handling

zhongkechen:
Abandoning this PR as most changes won't be required for the plugin feature. Required changes will go into the plugin PR if any.

PR #379, opened 2026-05-05
Add Lambda bundled runtime useragent header

nvasiu:
- Add boolean that checks if the SDK is running in a path that matches the path used within the Lambda bundled runtime.
- Update the useragent header string to include "-bundledruntime" if the above boolean is true.

*Issue #, if available:*

N/A

*Description of changes:*

**Note:** we are waiting to confirm the Lambda bundled runtime SDK path. Do NOT merge this PR until we get confirmation.

Updating the UserAgent header string to indicate whether the SDK is being run within the Lambda bundled runtime.

- A new boolean checks if the SDK's current path matches the path where it would be installed in the Lambda bundled runtime.
- If the above boolean is true, the string inserted into the UserAgent header will include "-bundled". Example:
`aws-durable-execution-sdk-python/{version}-bundled`

### DIFF
--- a/src/aws_durable_execution_sdk_python/lambda_service.py
+++ b/src/aws_durable_execution_sdk_python/lambda_service.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import copy
 import datetime
 import logging
+import os
 from collections.abc import MutableMapping
 from dataclasses import dataclass, field
 from enum import Enum
@@ -32,6 +33,8 @@ if TYPE_CHECKING:
 ReplayChildren: TypeAlias = bool
 OperationPayload: TypeAlias = str
 TimeoutSeconds: TypeAlias = int
+
+_IS_RUNTIME_BUNDLED: bool = os.path.dirname(__file__).startswith("/var/lang")
 
 logger = logging.getLogger(__name__)
 
@@ -1061,7 +1064,7 @@ class LambdaClient(DurableServiceClient):
                 config=Config(
                     connect_timeout=5,
                     read_timeout=50,
-                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}",
+                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}{'-bundled' if _IS_RUNTIME_BUNDLED else ''}",
                 ),
             )
         return cls(client=cls._cached_boto_client)
--- a/src/aws_durable_execution_sdk_python/lambda_service.py
+++ b/src/aws_durable_execution_sdk_python/lambda_service.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import copy
 import datetime
 import logging
-import os
 from collections.abc import MutableMapping
 from dataclasses import dataclass, field
 from enum import Enum
@@ -34,9 +33,16 @@ ReplayChildren: TypeAlias = bool
 OperationPayload: TypeAlias = str
 TimeoutSeconds: TypeAlias = int
 
-_IS_RUNTIME_BUNDLED: bool = os.path.dirname(__file__).startswith("/var/lang")
-
 logger = logging.getLogger(__name__)
+
+
+def _is_in_var_dir(module_file: str = __file__) -> bool:
+    """Return True if this SDK is installed under /var/lang/.
+
+    Lambda bundled Python runtimes install packages at
+    /var/lang/lib/pythonX.Y/site-packages/.
+    """
+    return module_file.startswith("/var/lang/")
 
 
 # region model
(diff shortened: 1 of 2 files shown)


src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -1061,7 +1064,7 @@ def initialize_client(cls) -> LambdaClient:
                 config=Config(
                     connect_timeout=5,
                     read_timeout=50,
-                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}",
+                    user_agent_extra=f"aws-durable-execution-sdk-python{'-bundledruntime' if _IS_RUNTIME_BUNDLED else ''}/{__version__}",
```
zhongkechen: should this `-bundledruntime` be part of the sdk name or part of the version? 

And `-bundledruntime` sounds weird like the runtime is bundled in SDK. Why don't we simply used `-bundled`?

nvasiu: I think it makes sense as part of the name, because the bundled runtime doesn't use a special SDK version.

I can make an update to just say "-bundled".

leandrodamascena: Sorry to jump into the conversation, but I would add the `AWS_EXECUTION_ENV` environment variable here - if not set assume "NA" for example. While this SDK is 100% focused on Lambda and only works with the specific payload sent to Lambda, you never know how creatively customers are using it. In Powertools for AWS Lambda, we identified clients using it on a wide variety of platforms other than Lambda.

But of course, it's just an idea, and I don't know if this affects how you consume this string in the backend.

yaythomas: agree re the bundling effectively being a version variant. something like `aws-durable-execution-sdk-python/v.1.2.3-bundled`

yaythomas: re AWS_EXECUTION_ENV

if exists could be:
```
# bring your own
aws-durable-execution-sdk-js/1.2.3#nodejs24.x

# bundled
aws-durable-execution-sdk-js/1.2.3-bundled#runtime-nodejs24.x

# no execution env variable, use current interpreter version
aws-durable-execution-sdk-js/1.2.3-bundled#node-v20.19.0
```

src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -33,6 +34,8 @@
 OperationPayload: TypeAlias = str
 TimeoutSeconds: TypeAlias = int
 
+_IS_RUNTIME_BUNDLED: bool = os.path.dirname(__file__).startswith("/var/lang")
```
yaythomas: should this be `/var/lang/`?

`/var/langsurprise/` will return true here.

also, os.path.dirname isn't adding much here, might as well just use `__file__.startswith("/var/lang/")`

tests/lambda_service_test.py
```diff
@@ -2047,6 +2047,45 @@ def test_lambda_client_initialize_client_no_endpoint(
     assert isinstance(client, LambdaClient)
 
 
+@patch("aws_durable_execution_sdk_python.lambda_service._IS_RUNTIME_BUNDLED", True)
+@patch.dict("os.environ", {}, clear=True)
```
yaythomas: does this actually do anything? there aren't any env vars under test?

src/aws_durable_execution_sdk_python/lambda_service.py
yaythomas: also, var name should prob be _IS_IN_VAR_DIR.

but the `__file__` lookup is also just a lookup in the global namespace, so at the module level this isn't winning much, might as well have the check in the initialize check and then unit test on the actual __file__ name too. 

```
   def _is_in_var_dir(module_file: str = __file__) -> bool:
       """Return True if this SDK is installed under /var/lang/.

       Lambda BYOL Python runtimes install bundled packages at
       /var/lang/lib/pythonX.Y/site-packages/ 
       """
       return module_file.startswith("/var/lang/")
```

yaythomas (dismissed):

yaythomas:
Reference: the similar change in ts/js https://github.com/aws/aws-durable-execution-sdk-js/pull/527/

ParidelPooya (approved):

src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -1061,7 +1070,7 @@ def initialize_client(cls) -> LambdaClient:
                 config=Config(
                     connect_timeout=5,
                     read_timeout=50,
-                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}",
+                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}{'-bundled' if _is_in_var_dir() else ''}",
```
bchampp, line 1073: I feel like `-runtime` might be more accurate (since it's _bundled_ in the **runtime**)

PR #381, opened 2026-05-06
fix: checkpoint START when retry a failed step

### LINKED ISSUE #373: [Bug]: Console doesn't show Step attempts properly when using Python SDK
### Expected Behavior

Console shows attempts in the timeline, e.g. with TypeScript SDK:

Image

### Actual Behavior

Console doesn't show the attempts properly in the timeline with Python SDK:

Image

### Steps to Reproduce

1. create a step that is always failing
2. trigger an execution
3. check the console for its attempts

### SDK Version

1.4

### Python Version

3.14

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

Fixes: #373 

*Description of changes:*

Step operations doesn't checkpoint START action when retrying a failed step operation, leading to missing attempts in console and also in plugin interface #371 

Without the fix:

image

With the fix:

image

The additional STARTED event:

image

### DIFF
--- a/src/aws_durable_execution_sdk_python/operation/step.py
+++ b/src/aws_durable_execution_sdk_python/operation/step.py
@@ -152,8 +152,8 @@ class StepOperationExecutor(OperationExecutor[T]):
         ):
             return CheckResult.create_is_ready_to_execute(checkpointed_result)
 
-        # Create START checkpoint if not exists
-        if not checkpointed_result.is_existent():
+        # Create START checkpoint if nonexistent or READY
+        if not checkpointed_result.is_existent() or checkpointed_result.is_ready():
             start_operation: OperationUpdate = OperationUpdate.create_step_start(
                 identifier=self.operation_identifier,
             )
--- a/src/aws_durable_execution_sdk_python/state.py
+++ b/src/aws_durable_execution_sdk_python/state.py
@@ -175,6 +175,13 @@ class CheckpointedResult:
             return False
         return op.status is OperationStatus.PENDING
 
+    def is_ready(self) -> bool:
+        """Return True if the checkpointed operation is READY."""
+        op = self.operation
+        if not op:
+            return False
+        return op.status is OperationStatus.READY
+
     def is_timed_out(self) -> bool:
         """Return True if the checkpointed operation is TIMED_OUT."""
         op = self.operation
--- a/tests/operation/step_test.py
+++ b/tests/operation/step_test.py
@@ -894,3 +894,58 @@ def test_step_executes_function_when_second_check_returns_started():
         mock_state.get_checkpoint_result.call_count == 1
     )  # Only one check for AT_LEAST_ONCE
     assert mock_state.create_checkpoint.call_count == 2  # START + SUCCEED checkpoints
+
+
+def test_step_creates_start_checkpoint_when_status_is_ready():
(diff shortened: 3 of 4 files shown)


src/aws_durable_execution_sdk_python/state.py
```diff
@@ -175,6 +175,13 @@ def is_pending(self) -> bool:
             return False
         return op.status is OperationStatus.PENDING
 
+    def is_ready(self) -> bool:
+        """Return True if the checkpointed operation is PENDING."""
```
SilanHe: nit: the comment is wrong. I'm getting it should be READY instead of PENDING

src/aws_durable_execution_sdk_python/operation/step.py
```diff
@@ -152,8 +152,8 @@ def check_result_status(self) -> CheckResult[T]:
         ):
             return CheckResult.create_is_ready_to_execute(checkpointed_result)
 
-        # Create START checkpoint if not exists
-        if not checkpointed_result.is_existent():
+        # Create START checkpoint if nonexistent or READY
+        if not checkpointed_result.is_existent() or checkpointed_result.is_ready():
```
SilanHe, line 156: hmm, with the added `or checkpointed_result.is_ready()`, does this means it's possible for SDK to checkpoint the start_operation more than once? why isn't the existing condition enough

src/aws_durable_execution_sdk_python/state.py
zhongkechen: Fixed.

src/aws_durable_execution_sdk_python/operation/step.py
zhongkechen: Both JS and Java SDK checkpoint a START action when a step is retried. Even the waitForCondition operation in Python also does that. Lacking of this is insufficient when backend/console as well as the plugins want to know when a step operation is actually started to retry.

yaythomas: The earlier branches in `check_result_status` already handle SUCCEEDED/FAILED/PENDING/STARTED, so by the time we hit this block the status is either nonexistent (first attempt) or READY (post-retry-timer).

yaythomas, line 156: for even less touch, the logic here could also be:

```python
     # Create START checkpoint unless already STARTED
     if not checkpointed_result.is_started():
         ...
```

zhongkechen: changes to previous blocks may impact this block if we don't specify the condition here explicitly.

yaythomas: I assume above comment was meant in reply to my comment further down on `  if not checkpointed_result.is_started():`. The isStarted check is similar to how JS/TS encode the logic.

but yes. sure, you could argue this guards against unintended fallthrough on hypothetical future refactors.

yaythomas (approved):

src/aws_durable_execution_sdk_python/operation/step.py
zhongkechen: I would like to limit the change scope to only this specific status `READY`

PR #387, opened 2026-05-08
feat: parallel and map branch name

### LINKED ISSUE #303: [Feature]: branch/iteration namer for concurrent operations
### What would you like?

Name an iteration of map:
```
const results = await context.map(
  "process-items",
  items,
  async (ctx, item, index) => {
    return await ctx.step(`process-${index}`, async () => process(item));
  },
  {
    maxConcurrency: 5,
    itemNamer: (item, index) => `item-${item.id}`, // Custom iteration names
  },
);
```

Name a branch of parallel:
```
const results = await context.parallel(
  "parallel-ops",
  [
    { 
      name: "fetch-user-data",  // Custom name for this branch
      func: async (ctx) => ctx.step(async () => fetchData1()) 
    },
    { 
      name: "fetch-order-history",  // Custom name for this branch
      func: async (ctx) => ctx.step(async () => fetchData2()) 
    },
  ],
  { maxConcurrency: 2 },
);
```

```
export interface NamedParallelBranch<TResult, Logger extends DurableLogger> {
  name?: string;
  func: ParallelFunc<TResult, Logger>;
}
```

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

wangyb-A:
*Issue #, if available:* 
close #303

*Description of changes:*

Add support for custom naming of map iterations and parallel branches.

Map: 
- item_namer on MapConfig[T]
- Make MapConfig generic over T (the item type)
- Add optional item_namer: Callable[[T, int], str] field
- When provided, replaces the default map-item-{index} naming scheme
- Receives the item and its index, returns a custom name string

Parallel: 
- Add ParallelBranch frozen dataclass with func and optional name
- ParallelBranch implements __call__, making it directly usable anywhere a plain callable is accepted
- When name is provided, replaces the default parallel-branch-{index} naming in execution history

Shared mechanism:
- Add get_iteration_name(index) method to ConcurrentExecutor base class
- MapExecutor and ParallelExecutor override it for custom naming

*Testing*:

iShot_2026-05-07_17 18 37
iShot_2026-05-07_17 19 32

iShot_2026-05-11_16 35 37

### DIFF
--- a/examples/src/parallel/parallel_with_named_branches.py
+++ b/examples/src/parallel/parallel_with_named_branches.py
@@ -1,34 +1,50 @@
-"""Example demonstrating parallel operations with named branches."""
+"""Example demonstrating all parallel branch patterns."""
 
 from typing import Any
 
 from aws_durable_execution_sdk_python.config import ParallelBranch, ParallelConfig
-from aws_durable_execution_sdk_python.context import DurableContext
+from aws_durable_execution_sdk_python.context import (
+    DurableContext,
+    durable_parallel_branch,
+)
 from aws_durable_execution_sdk_python.execution import durable_execution
+
+
+@durable_parallel_branch(name="fetch-orders")
+def fetch_orders(ctx: DurableContext) -> str:
+    return ctx.step(lambda _: "orders-loaded", name="load_orders")
+
+
+@durable_parallel_branch()
+def fetch_preferences(ctx: DurableContext) -> str:
+    return ctx.step(lambda _: "prefs-loaded", name="load_prefs")
 
 
 @durable_execution
 def handler(_event: Any, context: DurableContext) -> list[str]:
-    """Execute named parallel branches using ParallelBranch."""
+    """Execute parallel branches using all supported patterns."""
 
     return context.parallel(
         functions=[
+            # 1. Named parallel branch with ParallelBranch
             ParallelBranch(
                 func=lambda ctx: ctx.step(
                     lambda _: "user-data-loaded", name="load_user"
                 ),
                 name="fetch-user-data",
             ),
+            # 2. Named parallel branch with decorator
+            fetch_orders(),
+            # 3. Unnamed parallel branch with decorator
+            fetch_preferences(),
+            # 4. Unnamed parallel branch with ParallelBranch
             ParallelBranch(
                 func=lambda ctx: ctx.step(
-                    lambda _: "orders-loaded", name="load_orders"
+                    lambda _: "metrics-loaded", name="load_metrics"
                 ),
-                name="fetch-order-history",
             ),
-            ParallelBranch(
-                func=lambda ctx: ctx.step(lambda _: "prefs-loaded", name="load_prefs"),
-                name="fetch-preferences",
-            ),
+            # 5. No wrapper, just a raw callable
+            lambda ctx: ctx.step(lambda _: "config-loaded", name="load_config"),
         ],
         name="load_all_data",
         config=ParallelConfig(max_concurrency=3),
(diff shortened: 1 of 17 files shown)


examples/src/parallel/parallel_with_named_branches.py
```diff
@@ -0,0 +1,35 @@
...
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+
+@durable_execution
+def handler(_event: Any, context: DurableContext) -> list[str]:
+    """Execute named parallel branches using ParallelBranch."""
+
+    return context.parallel(
+        functions=[
+            ParallelBranch(
```
zhongkechen, line 30: I would rather provide this functionality as a decorator, e.g. `@durable_parallel_branch(name=...)`. This would be in align with the other decorators we have now.

wangyb-A: Do we want both - decorator and `ParallelBranch()` or just the decorator one?

examples/template.yaml
```diff
@@ -941,6 +941,42 @@
...
+        "Handler": "parallel_with_named_branches.handler",
+        "Description": "Parallel operation with named branches using ParallelBranch",
+        "Role": {
+          "Fn::GetAtt": [
+            "DurableFunctionRole",
+            "Arn"
+          ]
+        },
+        "DurableConfig": {
+          "RetentionPeriodInDays": 7,
```
zhongkechen, line 976: nit: this could be in Globals

src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -410,7 +418,7 @@ def _execute_item_in_child_context(
         operation_id: str = executor_context._create_step_id_for_logical_step(  # noqa: SLF001
             executable.index
         )
-        name: str = f"{self.name_prefix}{executable.index}"
```
zhongkechen, line 413: Not sure why we did this for both parallel and map. For parallel branches that use Python functions, they already have their unique names in `__name__`. We should leverage that to construct the branch names.

src/aws_durable_execution_sdk_python/operation/map.py
```diff
@@ -61,13 +63,14 @@ def __init__(
             nesting_type=nesting_type,
         )
         self.items = items
+        self._item_namer = item_namer
```
zhongkechen, line 66: Why is `_item_namer` internal while the other fields are public? Do we have a guideline? I feel it's quite arbitrary when deciding whether or not I field should be internal.

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -119,6 +121,52 @@ def function_with_arguments(child_context: DurableContext):
     return wrapper
 
 
+def durable_parallel_branch(
+    name: str | None = None,
+) -> Callable[
+    [Callable[Concatenate[DurableContext, Params], T]],
+    Callable[Params, ParallelBranch[T]],
```
zhongkechen, line 128: Nice! But is the return type readable?

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -496,7 +544,7 @@ def map_in_child_context() -> BatchResult[R]:
 
     def parallel(
         self,
-        functions: Sequence[Callable[[DurableContext], T]],
+        functions: Sequence[Callable[[DurableContext], T] | ParallelBranch[T]],
```
zhongkechen, line 547: Isn't ParallelBranch also a Callable

zhongkechen (approved):

PR #393, opened 2026-05-13
feat: add per-context replay status tracking

### LINKED ISSUE #389: [Feature]: per context replay status
### What would you like?

I would like to have accurate per-context replay status

Currently the replay status is global, which means it's inaccurate when there are multiple branches running concurrently. When one branch exits REPLAY mode, the whole execution will be in EXECUTION mode although some other branches are still replaying existing operations. This leads to inaccurate logging behavior as well as a blocker to correct plugin behavior (blocking #370  #386  )

### Possible Implementation

Implementation in JS for reference:

https://github.com/aws/aws-durable-execution-sdk-js/blob/5f951b1bae4a83c24a8d40a024d0a1b5346677fb/packages/aws-durable-execution-sdk-js/src/context/durable-context/durable-context.ts#L231-L239

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

Fixes #389 

*Description of changes:*


ayushiahjolia (dismissed):

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -417,6 +426,19 @@ def _create_step_id(self) -> str:
...
+        if self._replay_status is ReplayStatus.NEW:
+            return
+        # check if next operation exists
+        next_counter = self._step_counter.get_current() + 1
+        next_step_id = self._create_step_id_for_logical_step(next_counter)
+        if not self.state.get_checkpoint_result(next_step_id).is_existent():
+            # update the context replay status to NEW
+            self._replay_status = ReplayStatus.NEW
+            # update the execution replay status to NEW
+            self.state.transition_replay_status()
```
SilanHe, line 454: why do we still need the replay status in state if we're planning to set one in context?

zhongkechen: mainly for backward compatibility. Users may be using `state.is_replaying` now

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -417,6 +426,19 @@ def _create_step_id(self) -> str:
...
         return self._create_step_id_for_logical_step(new_counter)
 
+    def _track_replay(self) -> None:
+        """Transition replay status to NEW if the next operation has not been checkpointed"""
+        if self._replay_status is ReplayStatus.NEW:
+            return
+        # check if next operation exists
+        next_counter = self._step_counter.get_current() + 1
+        next_step_id = self._create_step_id_for_logical_step(next_counter)
+        if not self.state.get_checkpoint_result(next_step_id).is_existent():
```
zhongkechen, line 450: I'm afraid this wouldn't work if the next operation is a virtual context. TypeScript SDK also has this issue

zhongkechen: Pushed a commit to fix this

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -277,6 +277,42 @@ def result(self) -> T | None:
...
+        self._operation_counter: OrderedCounter = OrderedCounter()
+        self._virtual_operation_counter: OrderedCounter = OrderedCounter()
+        # child operations use this to generate deterministic step ids.
+        # differs from `parent_id` only for virtual contexts.
+        self._step_id_prefix: str | None = (
+            step_id_prefix if step_id_prefix is not None else parent_id
+        )
+
+    def peek_next_step_id(self):
+        next_step = self._operation_counter.get_current() + 1
```
ayushiahjolia, line 291: Do we need to look at `_virtual_operation_counter`?

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -277,6 +277,42 @@ def result(self) -> T | None:
...
+        ).increment()
+        return self.create_step_id_for_logical_step(new_counter, is_virtual=is_virtual)
+
+    def create_step_id_for_logical_step(self, step: int, is_virtual: bool) -> str:
+        """
+        Generate a step_id based on the given logical step.
+        This allows us to recover operation ids or even look
+        forward without changing the internal state of this context.
+        """
+        parts = [self._step_id_prefix, "v" if is_virtual else None, step]
```
ayushiahjolia, line 311: Isn't this a breaking change?

zhongkechen: How SDK generates the operation id is internal to the SDK. Users will only see a hash value.

src/aws_durable_execution_sdk_python/context.py
zhongkechen: This method is used to determine the replay status. `_virtual_operation_counter` is not used for that.

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -277,6 +277,42 @@ def result(self) -> T | None:
...
+
+        This method is an internal implementation detail. Do not rely the exact format of
+        the id generated by this method. It is subject to change without notice.
+        """
+        new_counter: int = (
+            self._virtual_operation_counter if is_virtual else self._operation_counter
+        ).increment()
+        return self.create_step_id_for_logical_step(new_counter, is_virtual=is_virtual)
+
+    def create_step_id_for_logical_step(self, step: int, is_virtual: bool) -> str:
```
SilanHe, line 305: nit: personally I think it would have been better to keep the virtual child context id tracking in a separate PR

src/aws_durable_execution_sdk_python/context.py
SilanHe: ~~I think we still want deterministic Ids AND unique Id for all virtual operations. With the current proposal, if there are multiple virtual child context between "real" steps, wouldn't they all have the same ID?~~

[UPDATE] sorry I misunderstood the counter. Currently, a virtual child context in the user context (pre hashing) would look like the following:

```
v-1
v-2
1
2
v-3
3-v-1
```

The only thing that bugs me here is that the virtual id doesn't preserve order within the context as a whole, but I guess it doesn't really matter in python since it's all hashed anyway.

tests/operation/map_test.py
```diff
@@ -306,7 +311,9 @@ def callable_func(ctx, item, idx, items):
     )
 
     executor_context = Mock()
-    executor_context._create_step_id_for_logical_step = lambda *args: "1"  # noqa SLF001
+    executor_context._operation_id_generator.create_step_id_for_logical_step = (
+        lambda *args: "1"
+    )  # noqa SLF001
```
SilanHe, line 316: nit: interesting pattern, since it looks like we have to override the linter, I do wonder if there are other (preferred) testing patterns.

tests/context_test.py
```diff
@@ -2151,14 +2194,61 @@ def test_should_propagate_outer_parent_id_when_virtual_is_nested_in_virtual():
...
+    mock_state.get_checkpoint_result.return_value = (
+        CheckpointedResult.create_not_found()
+    )
+    context._track_replay()
+    assert not context.is_replaying
+
+    # op id for 1
+    mock_state.get_checkpoint_result.assert_called_with(
+        "1ced8f5be2db23a6513eba4d819c73806424748a7bc6fa0d792cc1c7d1775a97"
+    )
```
SilanHe, line 2252: nit: are we missing new test asserting the behaviour of _track_replay() with a virtual child context as the next operation?

src/aws_durable_execution_sdk_python/context.py
ayushiahjolia: If next operation is a virtual operation then what would `self._operation_counter.get_current() + 1` return?

SilanHe: it returns the id of the next "real" operation

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -651,8 +688,8 @@ def step(
             ),
             context_logger=self.logger,
         )
+        self._track_replay()
```
yaythomas, line 691: `_track_replay()` runs after `create_step_id()` increments the counter, so `peek_next_step_id` resolves to N+1 instead of N. The check asks "is the op *after* the current one checkpointed?" rather than "is the current one?" JS does the inverse order on purpose (`checkAndUpdateReplayMode` runs before `createStepId`)

Same pattern for the other operations.

This will result in `is_replaying` being incorrect during the last replayed op's processing window.

No customer-observable effect today because the only internal reader of the flag is `Logger._should_log`.

src/aws_durable_execution_sdk_python/context.py
```diff
@@ -277,6 +277,42 @@ def result(self) -> T | None:
...
+        self._virtual_operation_counter: OrderedCounter = OrderedCounter()
+        # child operations use this to generate deterministic step ids.
+        # differs from `parent_id` only for virtual contexts.
+        self._step_id_prefix: str | None = (
+            step_id_prefix if step_id_prefix is not None else parent_id
+        )
+
+    def peek_next_step_id(self):
+        next_step = self._operation_counter.get_current() + 1
+        return self.create_step_id_for_logical_step(next_step, is_virtual=False)
```
yaythomas, line 292: Workflows that use `ChildConfig(is_virtual=True)` produce checkpoints at `prefix-v-N`, so won't the peek miss them?

This will result in any context whose next op is a virtual `run_in_child_context` flipping to `is_replaying=False` immediately at the next call into a virtual run_in_child_context. Logs at the top of the handler emit on every replay and any `if not ctx.is_replaying: ...` branches fire on every replay.

this is reachable via the public `ChildConfig.is_virtual` setting and will regress existing code, if any.

src/aws_durable_execution_sdk_python/logger.py
```diff
@@ -128,4 +125,4 @@ def _log(
         log_func(msg, *args, extra=merged_extra)
 
     def _should_log(self) -> bool:
-        return not self._execution_state.is_replaying()
+        return not self._context.is_replaying
```
yaythomas, line 128: `Logger` holds a `DurableContext` reference, and `with_log_info` (line 83) preserves it across `Logger.from_log_info(..., context=self._context)`.

When `create_child_context` builds a child logger via `parent_logger.with_log_info(...)`, the child's `_context` is permanently the parent. `_should_log` reads `not parent.is_replaying`, never the child's own flag.

This means this PR doesn't actually fix what it intends to, which is to say it keeps the broken behaviour where FLAT `parallel`/`map` branches log identically to each other, gated on the parent context's flag. Previously this happened via the global `state.is_replaying()`.

src/aws_durable_execution_sdk_python/state.py
```diff
@@ -343,39 +343,12 @@ def get_execution_operation(self) -> Operation | None:
...
-                }
-                if completed_ops.issubset(self._visited_operations):
-                    logger.debug(
-                        "Transitioning from REPLAY to NEW status at operation %s",
-                        operation_id,
-                    )
-                    self._replay_status = ReplayStatus.NEW
+    def transition_replay_status(self) -> None:
+        """Transition to NEW status"""
+        if self._replay_status is ReplayStatus.REPLAY:
```
yaythomas, line 348: Shouldn't this be under a lock? Two threads can both pass the `if`, both serially acquire the lock, both log "Transitioning…", both reassign.

since `_track_replay` is now called per-context-per-op this could potentially happen more frequently too.

zhongkechen:
Abandoned in favor of a backend enhancement to help SDK track the updated operation during the execution is suspended. That would be nicer than whatever SDK can use to guess that.

src/aws_durable_execution_sdk_python/context.py
zhongkechen: Skip the virtual context purposely to avoid flipping the flag when replay. This works in some cases but not in all.

PR #397, opened 2026-05-22
feat(sdk): monorepo repository to add new otel plugin package

SilanHe:
*Issue #, if available:*

*Description of changes:*

## Monorepo Restructure

The repo was converted from a single-package layout to a **monorepo** with a `packages/` directory:

- **Core SDK** moved from the repo root (`src/`, `tests/`, `examples/`, `pyproject.toml`) into `packages/aws-durable-execution-sdk-python/`
- All source, tests, and examples were relocated (no content changes, just path moves)
- I moved all hatch env "testing" scripts to the root pyproject.toml and only left the build/publishing artifacts in the per package pyproject.toml. Working environments are defined in root pyproject.toml.

## New Package: OpenTelemetry Instrumentation

A new package was added at `packages/aws-durable-execution-sdk-python-otel/` (v0.1.0):

## CI & Workflow Updates

- **`ci.yml`** — Loops over all `packages/*/` directories for format checks, type checking, tests, and builds
- **`ci-checks.sh`** — Iterates over a `PACKAGES` array instead of running checks at the repo root
- **`pypi-publish.yml`** — Uses a matrix strategy to build and publish each package independently; per-package artifact names
- **`sync-package.yml`** — Deleted (no longer needed)
- **`deploy-examples.yml`** / **`integration-tests.yml`** — Minor path adjustments

## Documentation

| File | Change |
|------|--------|
| `CONTRIBUTING.md` | Updated with monorepo structure overview; instructions to `cd` into package dirs before running hatch |
| `RELEASING.md` (new) | Release process: per-package versioning, tagging convention (`sdk-vX.Y.Z`, `otel-vX.Y.Z`), PyPI publishing flow, release notes format, pre-publish checklist |
| Core SDK `README.md` (new) | Package-level README inside the moved package directory |

## Misc

- `.gitignore` — Relaxed example build/zip patterns from root-anchored to relative (works with nested layout)


.github/workflows/pypi-publish.yml
SilanHe, line 1: I haven't actually tested this one but it looks quite reasonable. I had to work a little more closely with AI on this one.

packages/aws-durable-execution-sdk-python/.coverage
zhongkechen, line 1: What's in this file?

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/__about__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: 2025-present Amazon.com, Inc. or its affiliates.
+#
+# SPDX-License-Identifier: Apache-2.0
+__version__ = "0.1.0"
```
zhongkechen, line 4: Should we start with the same version as the SDK?

packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py
zhongkechen, line 1: `examples` worth its own package. We can do this later.

.github/workflows/ci.yml
```diff
@@ -43,10 +43,42 @@ jobs:
...
+            echo "=== Type checking: $pkg ==="
+            cd "$pkg"
+            hatch run types:check
+            cd "$GITHUB_WORKSPACE"
+          fi
+        done
     - name: Run tests + coverage
-      run: hatch run test:cov
+      run: |
+        for pkg in packages/*/; do
```
zhongkechen: Does the order matter when running tests/build?

.github/workflows/ci.yml
```diff
@@ -43,10 +43,42 @@ jobs:
...
+          fi
+        done
     - name: Build distribution
-      run: hatch build
+      run: |
+        for pkg in packages/*/; do
+          if [ -f "$pkg/pyproject.toml" ]; then
+            echo "=== Building: $pkg ==="
+            cd "$pkg"
+            hatch build
```
zhongkechen, line 65: how does this build rely on a local version of the other package instead of a version from pypi?

packages/aws-durable-execution-sdk-python/.coverage
SilanHe: woops, these are build artifacts that shouldn't be included

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/__about__.py
SilanHe: yeah sure, I didn't consider the version since this is currently an empty package

.github/workflows/ci.yml
SilanHe: ah you're right, it does not. I forgot to consider that for this PR. The currently generated empty otel plugin package doesn't depend on the aws-durable-execution-sdk-python which is why it didn't come up.

.github/workflows/ci.yml
SilanHe: it should, I forgot that the packages should use the local dependencies first rather than pypi

SilanHe: so I did some work and actually it does not matter, each hatch run test+cov is run in isolation.

.github/workflows/ci.yml
SilanHe: Fixed in newer commit

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/__about__.py
SilanHe: left this at 0.1.0 for now since the publish workflow would be merged as part of this pr, we don't want to publish accidentally.

RELEASING.md
```diff
@@ -0,0 +1,106 @@
...
+5. Write release notes following the format described in [Release Notes Format](#release-notes-format).
+6. Click **Publish release**.
+
+### Tagging Convention
+
+The tag should be the version of the package being bumped, prepended with a descriptive prefix:
+
+- **SDK only:** `sdk-v<version>` (e.g., `sdk-v1.6.0`)
+- **OTel only:** `otel-v<version>` (e.g., `otel-v0.3.0`)
+- **Both packages in a single release:** comma-separate the sub-tags (e.g., `sdk-1.6.0,otel-0.3.0`)
```
zhongkechen, line 44: This wouldn't be practically doable when we have more packages (e.g. testing, etc) in this repo and we want to release all of them. And also the version dependency becomes less obvious, e.g. otel-0.3.0 requires sdk-1.6.0+ and a future version otel-0.4.0 requires sdk-1.7.0+.

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
```diff
@@ -0,0 +1,67 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "aws-durable-execution-sdk-python-examples"
```
zhongkechen, line 6: The project name doesn't match the directory name

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+addopts = "-v --strict-markers"
+pythonpath = ["."]
+markers = [
+  "example: marks tests as example tests (deselect with '-m \"not example\"')",
+  "durable_execution: marks tests that use the durable_runner fixture (not used for test selection)",
+]
+
+[tool.hatch.envs.examples]
+dependencies = [
+  "aws-durable-execution-sdk-python-testing>=1.0.0",
```
zhongkechen: This dependency is already in `dependencies` above. Is this still needed?

zhongkechen:
linting and tests failed.

RELEASING.md
SilanHe: For the tags, this is the best approach we've got. At the moment, with the number of packages we have this is currently doable. We can re-evaluate if we exceed tag size limits. We can always only bump and release one package at a time.

For the dependency tree, the source of truth should be the pyproject.toml for the package.

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
SilanHe: ~~yeah I didn't define [project] dependencies, they're only defined per environment right now~~

oops looked at wrong file, I can run a quick test

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
SilanHe: ah yeah I can change this

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
SilanHe: works without the extra dependencies section for the env, so I removed

SilanHe:
> linting and tests failed.

Ah sorry, I was still working through the different worklows, there was an issue with a workflow which would check out the testing sdk into the root. Then the root testing scripts would pick up the testing sdk pyproject.toml

packages/aws-durable-execution-sdk-python/.coverage
zhongkechen: Can we remove this? It's pretty big

zhongkechen (dismissed):

yaythomas (changes requested):
still have to do a full review, but worried about a few things

1. ${{ github.event.release.target_commitish }} is the branch the release was created from (typically main). With this ref:, the build will check out the branch's HEAD at workflow run time rather than the tagged commit. That's a problem if main has moved on between when the operator pressed "Publish release" and when the workflow actually runs (concurrent merges, multi-package release windows): we'd ship whatever is on main, not the code at the tag.

2.  test:cov lost --cov-fail-under=98; CI no longer enforces 98
3. .gitignore out ot date with new changes
4. CONTRIBUTING guide out of data with new changes

zhongkechen:
out of date changes because this PR is open for too long. It's better to merge now than later and we can add the identified missing changes later. Otherwise we'll see more missing changes.

SilanHe:
> ${{ github.event.release.target_commitish }} is the branch the release was created from (typically main). With this ref:, the build will check out the branch's HEAD at workflow run time rather than the tagged commit. That's a problem if main has moved on between when the operator pressed "Publish release" and when the workflow actually runs (concurrent merges, multi-package release windows): we'd ship whatever is on main, not the code at the tag.

woops sorry, I meant to pin it to the tag but picked the wrong variable. I had just noticed that the workflow never did that in the past so I figured I would fix it. Thanks for pointing this out, going to use `github.event.release.tag_name
` instead.

> test:cov lost --cov-fail-under=98; CI no longer enforces 98

spoke offline, we'll follow up

> .gitignore out ot date with new changes

updated gitignore to use * instead of referencing the path directly

> CONTRIBUTING guide out of data with new changes

use package name instead of examples

.gitignore
```diff
@@ -30,7 +31,7 @@ dist/
 
 .kiro/
 
-/examples/build/*
-/examples/*.zip
+*/build/*
```
yaythomas, line 34: not blocking, but fix in follow-up pls:

`*/build/*` doesn't match `packages/<pkg>/build/<file>` because `*` doesn't span `/` ([gitignore(5)](https://git-scm.com/docs/gitignore#_pattern_format)). Either `**/build/*` + `**/*.zip`, or (preferred — scoped to `packages/`) `packages/*/build/` + `packages/*/*.zip`.

packages/aws-durable-execution-sdk-python/tests/operation/map_test.py
```diff
@@ -980,168 +980,190 @@ def get_checkpoint_result(self, operation_id):
...
+                )
+
+            assert isinstance(result, BatchResult)
+            assert len(mock_serialize.call_args_list) == 3
+            parent_call = mock_serialize.call_args_list[2]
+            assert parent_call[1]["serdes"] is custom_serdes
+            assert isinstance(parent_call[1]["value"], BatchResult)
+    finally:
+        importlib.reload(child)
         assert parent_call[1]["value"] is result
```
yaythomas, line 1167: This `assert ... is result` is in the `finally:` clause (after `importlib.reload(child)`), not inside the `try:`/`with patch(...)` body. Asymmetric to `parallel_test.py` L1140 where the equivalent line is correctly inside the `with` block. As-is, if the try-body raises before `parent_call` binds, the `finally` raises `UnboundLocalError` and masks the original exception. Could you move this line back inside the `with` block (3-line indent fix)?

packages/aws-durable-execution-sdk-python-otel/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```bash
+pip install aws-durable-execution-sdk-python-otel
+```
+
+## Quick Start
+
+```python
+from aws_durable_execution_sdk_python import DurableContext, durable_execution
+from aws_durable_execution_sdk_python_otel import instrument_durable_execution
```
yaythomas, line 19: not blockling: these don't match the code ,`__init__.py` only re-exports `__version__, `instrument_durable_execution` doesn't exist yet, and the Features section below documents capabilities (auto span creation, replay-aware tracing, error recording) that aren't implemented. 

maybe add something like DRAFT 'v0.1.0 reserves the package name; instrumentation lands in v0.2.0'.

.github/scripts/ci-checks.sh
```diff
@@ -1,18 +1,62 @@
-
 #!/bin/sh
```
yaythomas, line 1: Not blocking: Shebang is sh `#!/bin/sh` but lines 39 and 45 use bash array syntax (`PACKAGES=( ... )`, `"${PACKAGES[@]}"`).

Will fail under dash (the default `/bin/sh` on Debian/Ubuntu). 

either amend sheabing to `#!/usr/bin/env bash`, or rewrite to sh compliant.

yaythomas (changes requested):

yaythomas (approved):

.github/scripts/ci-checks.sh
SilanHe: https://github.com/aws/aws-durable-execution-sdk-python/pull/402

packages/aws-durable-execution-sdk-python/tests/operation/map_test.py
SilanHe: https://github.com/aws/aws-durable-execution-sdk-python/pull/402

.gitignore
SilanHe: https://github.com/aws/aws-durable-execution-sdk-python/pull/402

packages/aws-durable-execution-sdk-python-otel/README.md
SilanHe: https://github.com/aws/aws-durable-execution-sdk-python/pull/402

pyproject.toml
```diff
@@ -119,16 +146,3 @@ lines-after-imports = 2
   "SIM117",
   "TRY301",
 ]
-
```
SilanHe, line 122: > There are some warnings in the newest integration test [CI](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26660305247/job/78580909048?pr=402) workflow
packages/aws-durable-execution-sdk-python-examples/test/step/test_step.py:10
  /home/runner/work/aws-durable-execution-sdk-python/aws-durable-execution-sdk-python/language-sdk/packages/aws-durable-execution-sdk-python-examples/test/step/test_step.py:10: PytestUnknownMarkWarning: Unknown pytest.mark.example - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.exampleIs this something caused by the mono repo change? There was no warning before (e.g. [test run history](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26416296229/job/77761378697)) - @wangyb-A 

hmm so I deleted these but moved them to the aws-durable-execution-sdk-python-examples pyproject.toml.

I think since I'm running the tests from the root, it's not reading the pyproject.toml from the package itself

SilanHe: https://github.com/aws/aws-durable-execution-sdk-python/pull/403

PR #403, opened 2026-05-29
ci: centralize pytest configuration in root

SilanHe:
*Issue #, if available:*

*Description of changes:* We are getting new warnings for unknown pytest mark when running integration tests.

> There are some warnings in the newest integration test CI workflow
packages/aws-durable-execution-sdk-python-examples/test/step/test_step.py:10
/home/runner/work/aws-durable-execution-sdk-python/aws-durable-execution-sdk-python/language-sdk/packages/aws-durable-execution-sdk-python-examples/test/step/test_step.py:10: PytestUnknownMarkWarning: Unknown pytest.mark.example - is this a typo? You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
@pytest.mark.example Is this something caused by the mono repo change? There was no warning before (e.g. test run history) - @wangyb-A

For this PR, I'm moving duplicated `[tool.pytest.ini_options]` sections from sub-package `pyproject.toml` files to the root `pyproject.toml`. All pytest runs happen from the repo root, so markers, testpaths, and addopts only need to live in the root config.

Issue accidentally introduced in https://github.com/aws/aws-durable-execution-sdk-python/pull/397/changes#r3327105014

*Testing*: I ran the `hatch run test:cov` from the project root and there are no more warnings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,13 +22,19 @@ cov = "pytest --cov-report=term-missing --cov {args}"
 examples-integration = "pytest --runner-mode=cloud -m example packages/aws-durable-execution-sdk-python-examples/test/ -v {args}"
 
 [tool.pytest.ini_options]
-addopts = "--import-mode=importlib"
+# Declare custom markers to avoid warnings with --strict-markers
+addopts = "-v --strict-markers --import-mode=importlib"
 testpaths = [
     "packages/aws-durable-execution-sdk-python/tests",
     "packages/aws-durable-execution-sdk-python-otel/tests",
     "packages/aws-durable-execution-sdk-python-examples/test",
 ]
-
+markers = [
+  # Used for test selection with -m example
+  "example: marks tests as example tests (deselect with '-m \"not example\"')",
+  # Used for configuration - passes handler and lambda_function_name to durable_runner fixture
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -22,13 +22,19 @@ cov = "pytest --cov-report=term-missing --cov {args}"
 examples-integration = "pytest --runner-mode=cloud -m example packages/aws-durable-execution-sdk-python-examples/test/ -v {args}"
 
 [tool.pytest.ini_options]
-addopts = "--import-mode=importlib"
+# Declare custom markers to avoid warnings with --strict-markers
+addopts = "-v --strict-markers --import-mode=importlib"
```
zhongkechen, line 26: curious why we need `--import-mode=importlib`

zhongkechen (approved):

pyproject.toml
SilanHe: it's because of the new monorepo, if we have tests in 2 packages with the same file name, they will normally collide since the `--import-mode` is by default `prepend`. importlib mode uses Python's importlib to import each test file as a unique module based on its full file path, so there's no collision. 

https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

Issue #400, opened 2026-05-28
[Bug]: parallel_first_successful example crashes. BatchResult has no successful_results attribute


Nahuel990:
### Expected Behavior

The published example examples/src/parallel/parallel_first_successful.py accesses results.successful_results on
   the return value of context.parallel(...). That attribute does not exist on BatchResult anywhere in src/.
  Running the example raises AttributeError: 'BatchResult' object has no attribute 'successful_results'.         
                                                                      
  The three parallel branches all complete successfully; the failure is purely in the example's post-processing
  of the result object.

The example returns "First successful result: Task 1" (or whichever branch completes first) without an         
  exception. 

### Actual Behavior

The durable execution fails with:                                                                              
   
  ErrorType: AttributeError                                                                                      
  ErrorMessage: 'BatchResult' object has no attribute 'successful_results'

  The execution history shows all three parallel branches reaching StepSucceeded and ContextSucceeded for the    
  parallel context before the handler crashes.

### Steps to Reproduce

                                                                  
  1. git clone https://github.com/aws/aws-durable-execution-sdk-python.git
  2. cd aws-durable-execution-sdk-python
  3. grep -n "successful_results" examples/src/parallel/parallel_first_successful.py → matches line 23           
  4. grep -rn "successful_results" src/ → no matches
  5. Deploy examples/src/parallel/parallel_first_successful.py as a durable Lambda (or run via the in-repo local 
  test runner) and invoke it.    

### SDK Version

1.5.0

### Python Version

3.12

### Is this a regression?

No

### Additional Context

examples/src/parallel/parallel_first_successful.py:23:

  first_result = (
      results.successful_results[0] if results.successful_results else "None"
  )

  BatchResult class: src/aws_durable_execution_sdk_python/concurrency/models.py:88. Members defined on the      
  class:
                                                                                                                 
  succeeded, failed, started, status, has_failure, throw_if_error, get_results, get_errors, success_count,       
  failure_count, started_count, total_count, from_dict, from_items, to_dict.
                                                                                                                 
  No successful_results.                                              

  BatchResult Protocol:  src/aws_durable_execution_sdk_python/types.py:83. Only declares get_results(). No       
  successful_results.

SilanHe:
[...]

PR #413, opened 2026-06-02
fix(examples):parallel_first_successful update to match sdk

SilanHe:
*Issue #, if available:* #400

*Description of changes:*

`successful_results` was never formally introduced in the SDK but the example was not updated to reflect this. Using existing get_results() to fetch successful results from branches.

Added a new test to our integration tests. See successful run [here](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26851761724/job/79185322768?pr=413#step:6:1261).

### DIFF
--- a/packages/aws-durable-execution-sdk-python-examples/src/parallel/parallel_first_successful.py
+++ b/packages/aws-durable-execution-sdk-python-examples/src/parallel/parallel_first_successful.py
@@ -21,7 +21,6 @@ def handler(_event: Any, context: DurableContext) -> str:
     )
 
     # Extract the first successful result
-    first_result = (
-        results.successful_results[0] if results.successful_results else "None"
-    )
+    successful = results.get_results()
+    first_result = successful[0] if successful else "None"
     return f"First successful result: {first_result}"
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-python-examples/src/parallel/parallel_first_successful.py
```diff
@@ -21,7 +21,6 @@ def handler(_event: Any, context: DurableContext) -> str:
     )
 
     # Extract the first successful result
-    first_result = (
-        results.successful_results[0] if results.successful_results else "None"
```
SilanHe, line 25: example was failing to run on this line since successful_results is not a field of BatchResult

packages/aws-durable-execution-sdk-python-examples/src/parallel/parallel_first_successful.py
SilanHe, line 14: I could make this more deterministic so that one of these branches is always completing first. I opted not to, in order to preserve most of the example. Let me know if that's preferable.

yaythomas (approved):

PR #417, opened 2026-06-04
feat: add otel plugin (#371)

### LINKED ISSUE #386: [Feature]: Create an OTel plugin
### What would you like?

I would like to have a plugin for OTel so that I can see traces on OTel platform such as X-Ray

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

fix: #386 

*Description of changes:*

initial implementation of Otel plugin

- start an invocation span when an invocation starts and end it when the invocation completes
- start an operation span when an operation is started and end it when the operation completes or the invocation completes.

Know issues
- not implemented: start and end an operation span when an operation is changed to terminal status when replay, and link the span to the previous span for the same operation
- `onOperationStart` is called multiple times for Step operation if it's retried
- the attempt is always None for CONTEXT, leading to missing initial span for context operations

### DIFF
--- a/packages/aws-durable-execution-sdk-python-examples/cli.py
+++ b/packages/aws-durable-execution-sdk-python-examples/cli.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import shutil
+import subprocess
 import sys
 import time
 import zipfile
@@ -35,6 +36,7 @@ def build_examples():
 
     build_dir = Path(__file__).parent / "build"
     src_dir = Path(__file__).parent / "src"
+    packages_dir = Path(__file__).parent.parent
 
     logger.info("Building examples...")
 
@@ -57,15 +59,29 @@ def build_examples():
         logger.exception("Failed to copy testing library")
         return False
 
-    # Copy SDK source from the main SDK package
-    testing_src = (
-        Path(__file__).parent.parent
-        / "aws-durable-execution-sdk-python"
-        / "src"
-        / "aws_durable_execution_sdk_python"
-    )
-    logger.info("Copying SDK from %s", testing_src)
-    shutil.copytree(testing_src, build_dir / "aws_durable_execution_sdk_python")
+    # Install local packages so their runtime dependencies are included in
+    # the Lambda deployment package.
+    runtime_packages = [
+        packages_dir / "aws-durable-execution-sdk-python",
+        packages_dir / "aws-durable-execution-sdk-python-otel",
+    ]
+    try:
+        subprocess.run(
+            [
+                sys.executable,
+                "-m",
+                "pip",
+                "install",
+                "--upgrade",
+                "--target",
+                str(build_dir),
+                *[str(package) for package in runtime_packages],
+            ],
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        logger.exception("Failed to install runtime dependencies")
+        return False
 
     # Copy example functions
     logger.info("Copying examples from %s", src_dir)
--- a/packages/aws-durable-execution-sdk-python-examples/src/plugin/execution_with_otel.py
+++ b/packages/aws-durable-execution-sdk-python-examples/src/plugin/execution_with_otel.py
@@ -1,0 +1,53 @@
+"""Demonstrates handler execution without any durable operations."""
+
+from typing import Any
+
+from opentelemetry import trace
+
+from aws_durable_execution_sdk_python import StepContext
+from aws_durable_execution_sdk_python.config import Duration, StepConfig, StepSemantics
+from aws_durable_execution_sdk_python.context import (
+    DurableContext,
+    durable_step,
+    durable_with_child_context,
+)
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+from aws_durable_execution_sdk_python_otel import DurableExecutionOtelPlugin
+
+
+# use default provider
+tracer_provider = trace.get_tracer_provider()
+otel = DurableExecutionOtelPlugin(tracer_provider)
+
+
+@durable_step
+def add_numbers(_step_context: StepContext, a: int, b: int) -> int:
+    return a + b
+
+
+@durable_with_child_context
+def add_numbers_in_child(child_context: DurableContext, a: int, b: int):
+    result: int = child_context.step(
+        add_numbers(a, b),
+        name=f"step-{b}",
+    )
+    child_context.wait(
+        Duration.from_seconds(1),
+        name=f"wait-{b}",
+    )
+    return result
+
+
+@durable_execution(plugins=[otel])
+def handler(_event: Any, context: DurableContext) -> int:
+    result = 0
+    for i in range(3):
+        result += context.run_in_child_context(
+            add_numbers_in_child(6, i),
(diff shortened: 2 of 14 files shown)


packages/aws-durable-execution-sdk-python-otel/pyproject.toml
```diff
@@ -25,6 +25,8 @@ dependencies = [
   "aws-durable-execution-sdk-python>=1.5.0",
   "opentelemetry-api>=1.20.0",
   "opentelemetry-sdk>=1.20.0",
+  "opentelemetry-exporter-otlp",
+  "opentelemetry-propagator-aws-xray",
```
zhongkechen, line 29: Not sure if these dependencies are required. If not, we can add them to the dependencies of example package

packages/aws-durable-execution-sdk-python-examples/cli.py
```diff
@@ -57,15 +59,29 @@ def build_examples():
...
+                "--upgrade",
+                "--target",
+                str(build_dir),
+                *[str(package) for package in runtime_packages],
+            ],
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        logger.exception("Failed to install runtime dependencies")
+        return False
```
zhongkechen, line 84: This would break local building of the package on MacOs if any of the dependencies rely on native extensions

zhongkechen: For now it seems OK. OTel doesn't have native extensions.

packages/aws-durable-execution-sdk-python-examples/cli.py
```diff
@@ -57,15 +59,29 @@ def build_examples():
...
-        / "src"
-        / "aws_durable_execution_sdk_python"
-    )
-    logger.info("Copying SDK from %s", testing_src)
-    shutil.copytree(testing_src, build_dir / "aws_durable_execution_sdk_python")
+    # Install local packages so their runtime dependencies are included in
+    # the Lambda deployment package.
+    runtime_packages = [
+        packages_dir / "aws-durable-execution-sdk-python",
+        packages_dir / "aws-durable-execution-sdk-python-otel",
```
SilanHe, line 66: thanks!

packages/aws-durable-execution-sdk-python-examples/template.yaml
```diff
@@ -995,6 +995,24 @@
           "ExecutionTimeout": 300
         }
       }
+    },
```
SilanHe, line 998: nit: this could be misleading since it doesn't include the xray otel layer

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -0,0 +1,436 @@
...
+                    Link(
+                        context=SpanContext(
+                            trace_id=self._id_generator.generate_trace_id(),
+                            span_id=span_id,
+                            is_remote=False,
+                            trace_flags=TraceFlags(TraceFlags.SAMPLED),
+                        )
+                    )
+                ]
+                self._id_generator.set_next_span_id(None)
```
SilanHe, line 195: ok, I see the next span has nothing set and I assume it randomly generates an id

packages/aws-durable-execution-sdk-python-examples/src/plugin/execution_with_otel.py
```diff
@@ -0,0 +1,53 @@
...
+        name=f"step-{b}",
+    )
+    child_context.wait(
+        Duration.from_seconds(1),
+        name=f"wait-{b}",
+    )
+    return result
+
+
+@durable_execution(plugins=[otel])
```
SilanHe, line 42: for follow up in another CR, what are the other setup steps we need for this plugin?

SilanHe (approved):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -0,0 +1,436 @@
...
+            )
+        else:
+            span.set_status(StatusCode.OK)
+
+        end_timestamp = info.end_time
+        if end_timestamp is not None and end_timestamp == info.start_time:
+            end_timestamp += datetime.timedelta(microseconds=1)
+        self._end_span(info.operation_id, end_timestamp)
+
+    def on_user_function_start(self, info: UserFunctionStartInfo) -> None:
```
SilanHe, line 339: so this doesn't work with concurrent operations out of the box right? it's relying on 

https://github.com/aws/aws-durable-execution-sdk-python/pull/371/changes#r3359319079?

packages/aws-durable-execution-sdk-python-examples/template.yaml
zhongkechen: Yeah, this example case doesn't run properly now. Without the layer, all spans are `NonRecordingSpan`

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
zhongkechen: Yeah. This doesn't work well if there are concurrent operations. We need a fix of this.

packages/aws-durable-execution-sdk-python-examples/template.yaml
zhongkechen: Created an issue: https://github.com/aws/aws-durable-execution-sdk-python/issues/421

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
zhongkechen: any of the plugin methods can be called when another part of user code is running concurrently in parallel branch/map item

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
```diff
@@ -0,0 +1,109 @@
...
+            execution_arn: The durable execution ARN (used for the hash portion).
+            start_timestamp: start time of invocation
+        """
+        self._execution_trace_id = _to_otel_trace_id(execution_arn, start_timestamp)
+
+    def generate_trace_id(self) -> int:
+        """Generate a 128-bit trace ID."""
+        return self._execution_trace_id or self._random_id_generator.generate_trace_id()
+
+    def generate_span_id(self) -> int:
```
zhongkechen, line 106: This method could be called concurrently so set then generate pattern may not work well in a multithreading env

zhongkechen:
There are known issues to this plugin. I'm going to merge now and follow-up on the issues later

Issue #385, opened 2026-05-07
[Feature]: logging interface for plugin


zhongkechen:
### What would you like?

I would like to allow plugins to add contexts to logger

Spin off from https://github.com/aws/aws-durable-execution-sdk-python/issues/370

Consider the alternatives, e.g. https://docs.aws.amazon.com/powertools/python/latest/core/logger/

### Is this a breaking change?

No

### Does this require an RFC?

No

zhongkechen:
A python native way of structured logging should be used:

```
import logging

logger = logging.getLogger()

span = ContextVar("span")

# 1. Define a filter that dynamically attaches attributes to the record
class CustomContextFilter(logging.Filter):
    def filter(self, record):
        # Assign attributes directly to the record object
        record.span_id = span.get()
        return True

# 2. Instantiate and register the filter to Lambda's active logger handler
context_filter = CustomContextFilter()

if logger.handlers:
    for handler in logger.handlers:
        handler.addFilter(context_filter)

def lambda_handler(event, context):
    # 3. Update the filter state in plugin
    span.set("span id")
        
    # This log implicitly gets the "span_id" JSON property
    logger.info("Application execution checkpoint reached")

```

wangyb-A:
Do we want the customer to add the logger filter on their side? Or keep it like the previous implementation? (with a boolean parameter during plugin initialization)

zhongkechen:
This will be transparent to users. Similar in AWS Lambda Python runtime: 

https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/f11e7c5c5cc7caf7a44263d3978e0dfda4f15250/awslambdaric/bootstrap.py#L339-L343

PR #427, opened 2026-06-09
feat: add otel logger injection

wangyb-A:
*Issue #, if available:* #385

## Description

- Add a `wrap_logger` hook to DurableInstrumentationPlugin and chain it in `PluginExecutor.wrap_logger`.
- Wire logger wrapping into DurableContext.set_logger, and trigger it once on the root context in durable_execution. Child contexts inherit the already-wrapped logger; wrapping is idempotent.
- New `OtelEnrichedLogger` that injects the active span context into each log extra.
- Plugin resolves the correct span via `get_current_span_context`: the active operation span inside a step/child context, the enclosing operation span between steps, and the invocation span (from the registry) at the top level.
- `on_user_function_end` now restores the enclosing operation span (parent_id) rather than always the invocation span, so between-step logs inside a child context correlate to that child context.

## Testing

- Unit test
- Example logging:
```
// 1. handler() line 62:  context.logger.info("Workflow started")
//    Top-level handler code → invocation span
{
  "message": "Workflow started", "logger": "root",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "bbe5fdb86682dac4",   // invocation
  "otel.trace_sampled": true
}

// 2. greet() line 45:  step_context.logger.info("Greeting inside step", ...)
//    Invoked via context.step(greet("world"), name="top-greet") at line 64
//    → top-greet step span
{
  "message": "Greeting inside step", "logger": "root",
  "operationName": "top-greet", "attempt": 1, "greeting_name": "world",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "796a9b22593b6959",   // top-greet
  "otel.trace_sampled": true
}

// 3. greet_in_child() line 52:  child_context.logger.info("Entering child context")
//    Invoked via context.run_in_child_context(..., name="child-context") at line 65
//    → child-context span
{
  "message": "Entering child context", "logger": "root",
  "parentId": "c5faca15...3947e6",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "be02ea8c17bd3715",   // child-context
  "otel.trace_sampled": true
}

// 4. greet() line 45 again, this time invoked via
//    child_context.step(greet("nested"), name="child-greet") at line 53
//    → child-greet step span (nested under child-context)
{
  "message": "Greeting inside step", "logger": "root",
  "operationName": "child-greet", "attempt": 1, "greeting_name": "nested",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "67fdcf54688f3644",   // child-greet
  "otel.trace_sampled": true
}

// 5. greet_in_child() line 54:  child_context.logger.info("Leaving child context", ...)
//    Runs AFTER the inner child-greet step finished → restored to child-context span
//    (this is the parent-restore fix: NOT the invocation span)
{
  "message": "Leaving child context", "logger": "root",
  "parentId": "c5faca15...3947e6", "result": "hello nested",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "be02ea8c17bd3715",   // child-context (matches #3)
  "otel.trace_sampled": true
}

// 6. handler() line 70:  context.logger.info("Workflow completed", ...)
//    Back at top level after the child context unwound → invocation span
{
  "message": "Workflow completed", "logger": "root",
  "top": "hello world", "nested": "hello nested",
  "otel.trace_id": "6a288a880695a2d879ef07517e60531b",
  "otel.span_id":  "bbe5fdb86682dac4",   // invocation (matches #1)
  "otel.trace_sampled": true
}


```

### DIFF
--- a/packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
+++ b/packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
@@ -1,0 +1,67 @@
+"""Demonstrates OTel-enriched logging in a durable execution.
+
+The DurableExecutionOtelPlugin wraps the execution logger (enrich_logger=True
+by default) so every log line emitted through context.logger / step_context.logger
+is automatically enriched with the active OpenTelemetry trace context
+(otel.trace_id, otel.span_id, otel.trace_sampled). This lets logs correlate to
+the spans the plugin emits without any user code changes.
+
+Logs emitted:
+- at the top level correlate to the invocation span
+- inside a step correlate to that step's span
+- inside a child context correlate to the child-context span
+"""
+
+from typing import Any
+
+from aws_durable_execution_sdk_python_otel import DurableExecutionOtelPlugin
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+
+from aws_durable_execution_sdk_python import StepContext
+from aws_durable_execution_sdk_python.context import (
+    DurableContext,
+    durable_step,
+    durable_with_child_context,
+)
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+
+tracer_provider = TracerProvider()
+trace.set_tracer_provider(tracer_provider)
+
+# enrich_logger defaults to True, so the execution logger is wrapped with OTel
+# trace context injection (otel.trace_id, otel.span_id, otel.trace_sampled).
+otel = DurableExecutionOtelPlugin(tracer_provider)
+
+
+@durable_step
+def greet(step_context: StepContext, name: str) -> str:
+    # Logged inside a step: enriched with this step's span_id.
+    # Note: avoid reserved LogRecord keys (e.g. "name") in extra.
+    step_context.logger.info("Greeting inside step", extra={"greeting_name": name})
+    return f"hello {name}"
+
+
+@durable_with_child_context
+def greet_in_child(child_context: DurableContext, name: str) -> str:
+    # Logged inside a child context: enriched with the child-context span_id.
+    child_context.logger.info("Entering child context")
+    result: str = child_context.step(greet(name), name="child-greet")
+    child_context.logger.info("Leaving child context", extra={"result": result})
+    return result
+
+
+@durable_execution(plugins=[otel])
+def handler(_event: Any, context: DurableContext) -> str:
+    # Logged at the top level: enriched with the invocation span_id.
+    context.logger.info("Workflow started")
+
+    top: str = context.step(greet("world"), name="top-greet")
+    nested: str = context.run_in_child_context(
+        greet_in_child("nested"), name="child-context"
+    )
+
+    context.logger.info("Workflow completed", extra={"top": top, "nested": nested})
+    return f"{top} | {nested}"
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
@@ -20,7 +20,6 @@ from aws_durable_execution_sdk_python.plugin import (
 )
 from opentelemetry import context, trace
 from opentelemetry.context import Context
-from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace import TracerProvider as SdkTracerProvider
 from opentelemetry.sdk.trace.sampling import TraceIdRatioBased
 from opentelemetry.trace import (
@@ -40,10 +39,11 @@ from aws_durable_execution_sdk_python_otel.deterministic_id_generator import (
     DeterministicIdGenerator,
     operation_id_to_span_id,
 )
+from aws_durable_execution_sdk_python_otel.logger import OtelEnrichedLogger
 
 
 if TYPE_CHECKING:
-    pass
+    from aws_durable_execution_sdk_python.types import LoggerInterface
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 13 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -124,6 +146,34 @@ def _get_span(self, operation_id: str | None) -> Span | None:
...
+
+        Resolution order:
+        1. The span attached to the OTel thread-local context. Inside a step or
+           child context this is the active operation span (attached in
+           on_user_function_start), and between operations it is the enclosing
+           operation span (restored in on_user_function_end).
+        2. The invocation span from the plugin registry. This is the path used
+           for top-level handler code: the invocation span is never attached to
+           the worker thread's context, so the registry is the only way to
+           resolve it.
```
wangyb-A, line 164: Currently, on_invocation_start is running on main thread (instead of worker thread). So, even if we attach the span to the context, it's invisible for worker thread.

packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
```diff
@@ -0,0 +1,67 @@
...
+)
+from aws_durable_execution_sdk_python.execution import durable_execution
+
+
+tracer_provider = TracerProvider()
+trace.set_tracer_provider(tracer_provider)
+
+# enrich_logger defaults to True, so the execution logger is wrapped with OTel
+# trace context injection (otel.trace_id, otel.span_id, otel.trace_sampled).
+otel = DurableExecutionOtelPlugin(tracer_provider)
```
SilanHe, line 36: nit: For follow up in future pr, I think we also want to make the DurableExecutionOtelPlugin as "no-touch" as possible. The typical workflow for a customer adding OpenTelemetry is just adding a lambda layer and changing a few lambda environment variables and it should "just work".

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -124,6 +146,34 @@ def _get_span(self, operation_id: str | None) -> Span | None:
...
+
+        Returns:
+            A valid SpanContext, or None if no span is active.
+        """
+        span_context = trace.get_current_span().get_span_context()
+        if span_context and span_context.is_valid:
+            return span_context
+
+        invocation_span = self._get_span(None)
```
SilanHe, line 173: does the `self._get_span(None)` always return the invocation span? I don't see it being set anywhere. Do we even need this line and the block after it?

packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
```diff
@@ -0,0 +1,67 @@
...
+    step_context.logger.info("Greeting inside step", extra={"greeting_name": name})
+    return f"hello {name}"
+
+
+@durable_with_child_context
+def greet_in_child(child_context: DurableContext, name: str) -> str:
+    # Logged inside a child context: enriched with the child-context span_id.
+    child_context.logger.info("Entering child context")
+    result: str = child_context.step(greet(name), name="child-greet")
+    child_context.logger.info("Leaving child context", extra={"result": result})
```
SilanHe, line 52: I don't see this log in your example logs, can you check if it actually is present and that the spanId is the same as the log for "Entering child context"?

packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
SilanHe: In our case, we'd like it to be as simple as the following (if possible):

1. Adding ADOT layer (or whatever layer we recommend)
2. adding the aws-durable-execution-sdk-python-otel to their dependencies
3. changing lambda env var
4. enabling tracing on the lambda function
5. profit

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
SilanHe: ah never mind, I see that `on_invocation_start` calls `_start_span(None,...)` which will call

```
self._operation_spans[operation_id] = span
```

setting the invocation_span with operation_id = None

https://github.com/aws/aws-durable-execution-sdk-python/blame/f337788752001b5579c3a5b6ad7ce54d1ce51b79/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py#L266

But do we need this fallback code? I would think that `trace.get_current_span().get_span_context()` should return the current span, including the invocation span.

SilanHe (dismissed):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
wangyb-A: > But do we need this fallback code?

Yes, we need it for current implementation, reason:

1. Currently we only register span in `on_user_function_start`. We still need to get the invocation span if user call logger outside of user func (step, child)

2. Currently `on_invocation_start` is called on main thread, `on_user_function_start` is called on worker thread. Handler code is called on worker thread. So, even if we register the span in `on_invocation_start`, the logger in worker thread cannot find the span.

packages/aws-durable-execution-sdk-python-examples/src/otel/otel_logger_example.py
wangyb-A: Update in the pr description

SilanHe (approved):

Issue #428, opened 2026-06-09
[Feature][OTel]: Move on invocation start and end to worker thread


wangyb-A:
### What would you like?

- Move on invocation start and end to worker thread, instead of main thread.
- Attach and de-attach to otel context

### Is this a breaking change?

No

### Does this require an RFC?

No

AdityaAudi:
@wangyb-A  I took a look and wanted to sanity-check intent before picking this up, since the design has a couple of forks:                                         
                                                                            
1. Should all plugin callbacks move to the worker thread, or just invocation start/end? Moving only the two creates an ordering      
hazard, operation callbacks (still sync) could fire on the user-code thread before the invocation span exists.                                       
2. My lean is to capture the Context snapshot in `on_invocation_start` and pass it explicitly into downstream `start_span(context=...)` calls, rather than relying on `context.attach()` (per-thread via contextvars, so it doesn't cross the worker→user-code boundary cleanly). Open to a different model if you had one in mind.      
3. Assuming this means the missing `context.detach()` at the end of `on_user_function_end` (there's an existing comment skipping it at [plugin.py](https://github.com/aws/aws-durable-execution-sdk-python/blob/a244a18976fb33659f98441ffe6a495552a3cbab/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py#L410)) or did you mean something broader?
                                                                            
Happy to start on a PR once these are settled.

PR #467, opened 2026-06-10
fix(otel): Use id generator fallback

### LINKED ISSUE #426: [Feature]: id generator in should fall back to the original one
### What would you like?

id generator in Python should fall back to the original one, instead of a random one.

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

wangyb-A:
Import TracerProvider from opentelemetry.sdk.trace instead of the API
base class so id_generator and sampler attributes resolve correctly
under mypy. Make DeterministicIdGenerator subclass RandomIdGenerator
for type compatibility with the provider's id_generator attribute.

Change DeterministicIdGenerator to fall back to the tracer provider's
original id generator rather than always creating a new
RandomIdGenerator. The plugin now captures the provider's existing
generator and passes it as the fallback before overwriting it.

Add unit tests covering the fallback behavior and update existing
tests to reference the renamed _fallback_id_generator attribute.

*Issue #, if available:* close #426

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
@@ -5,11 +5,11 @@ from __future__ import annotations
 import hashlib
 import os
 import re
-from datetime import datetime, UTC
+from datetime import UTC, datetime
 
-from opentelemetry.sdk.trace import RandomIdGenerator
+from opentelemetry.sdk.trace import IdGenerator, RandomIdGenerator
 
-HASH_LENGTH = 16
+
 HASHED_ID_PATTERN = re.compile(r"^[0-9a-f]{16}$")
 
 
@@ -67,19 +67,25 @@ def operation_id_to_span_id(operation_id: str) -> int:
 
 class DeterministicIdGenerator(RandomIdGenerator):
     """An ID generator that produces deterministic span IDs when a pending
-    operation ID is set, and random IDs otherwise.
+    operation ID is set, and falls back to the provided generator otherwise.
 
     Trace IDs are deterministic when an execution ARN is set, ensuring all
-    invocations of the same durable execution share a single trace.
+    invocations of the same durable execution share a single trace. When no
+    deterministic ID is available, generation is delegated to the fallback
+    generator (the tracer provider's original ID generator by default).
 
     Trace IDs embed a real timestamp so they satisfy the X-Ray format
     requirement (first 8 hex chars = Unix epoch seconds).
+
+    Args:
+        fallback_id_generator: Generator used when no deterministic ID is
+            available. Defaults to a new ``RandomIdGenerator``.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, fallback_id_generator: IdGenerator | None = None) -> None:
         self._next_span_id: int | None = None
         self._execution_trace_id: int | None = None
-        self._random_id_generator = RandomIdGenerator()
+        self._fallback_id_generator = fallback_id_generator or RandomIdGenerator()
 
     def set_next_span_id(self, span_id: int | None) -> None:
         """Set the operation ID to use for the next span's ID.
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
```diff
@@ -65,21 +65,29 @@ def operation_id_to_span_id(operation_id: str) -> int:
...
+            available. Defaults to a new ``RandomIdGenerator``.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, fallback_id_generator: IdGenerator | None = None) -> None:
         self._next_span_id: int | None = None
         self._execution_trace_id: int | None = None
-        self._random_id_generator = RandomIdGenerator()
+        self._fallback_id_generator = (
+            fallback_id_generator if fallback_id_generator else RandomIdGenerator()
```
zhongkechen: Pythonic style: fallback_id_generator or RandomIdGenerator()

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
wangyb-A: i'm following the style we have in sdk, e.g. https://github.com/aws/aws-durable-execution-sdk-python/blob/main/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py#L301

Update it with `fallback_id_generator is not None`

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
```diff
@@ -67,19 +67,25 @@ def operation_id_to_span_id(operation_id: str) -> int:
 
 class DeterministicIdGenerator(RandomIdGenerator):
```
zhongkechen, line 68: The base class was to make type checker happy. But I feel it's not correct

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
```diff
@@ -67,19 +67,25 @@ def operation_id_to_span_id(operation_id: str) -> int:
...
     """
 
-    def __init__(self) -> None:
+    def __init__(self, fallback_id_generator: IdGenerator | None = None) -> None:
         self._next_span_id: int | None = None
         self._execution_trace_id: int | None = None
-        self._random_id_generator = RandomIdGenerator()
+        self._fallback_id_generator = fallback_id_generator or RandomIdGenerator()
 
     def set_next_span_id(self, span_id: int | None) -> None:
```
zhongkechen, line 90: Will the threading issue be addressed in a separate PR?

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
wangyb-A: yes, threading issue is tracked here: #428

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
wangyb-A: I think the reason is the checker is complaining this like

`self._provider.id_generator = self._id_generator`

The provider's id_generator should use the base class, but they are using the `RandomIdGenerator` class

zhongkechen: Can you find a way to fix this?

zhongkechen (approved):

PR #471, opened 2026-06-11
fix: re-enable type checking

### LINKED ISSUE #469: [Bug]: type checking failed silently
### Expected Behavior

If type check fails, its parent workflow should fail

### Actual Behavior

The type checking command failed silently:

error: Can't determine which types to install with no files to check (and no cache from previous mypy run)

https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/27314829576/job/80693005560#step:6:1

### Steps to Reproduce

N/A

### SDK Version

Latest

### Python Version

3.14

### Is this a regression?

No

### Last Working Version

_No response_

### Additional Context

_No response_

zhongkechen:
*Issue #, if available:*

fix: #469 

*Description of changes:*

Type checking is not working after we migrated to monorepo and the ci workflow succeeded with the type checker actually failing. 

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
@@ -7,7 +7,7 @@ import os
 import re
 from datetime import datetime, UTC
 
-from opentelemetry.sdk.trace import IdGenerator, RandomIdGenerator
+from opentelemetry.sdk.trace import RandomIdGenerator
 
 HASH_LENGTH = 16
 HASHED_ID_PATTERN = re.compile(r"^[0-9a-f]{16}$")
@@ -65,7 +65,7 @@ def operation_id_to_span_id(operation_id: str) -> int:
     return int(hashed_operation_id, 16)
 
 
-class DeterministicIdGenerator(IdGenerator):
+class DeterministicIdGenerator(RandomIdGenerator):
     """An ID generator that produces deterministic span IDs when a pending
     operation ID is set, and random IDs otherwise.
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ requires-python = ">=3.11"
 workspace.members = ["packages/*"]
 
 [tool.hatch.envs.test]
+detached = true
 workspace.members = ["packages/*"]
 dependencies = [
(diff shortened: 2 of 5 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
```diff
@@ -65,7 +65,7 @@ def operation_id_to_span_id(operation_id: str) -> int:
     return int(hashed_operation_id, 16)
 
 
-class DeterministicIdGenerator(IdGenerator):
+class DeterministicIdGenerator(RandomIdGenerator):
```
zhongkechen, line 68: Not sure if this is a correct fix:
```
packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py:101: error: Incompatible types in assignment (expression has type "DeterministicIdGenerator", variable has type "RandomIdGenerator")  [assignment]
```

wangyb-A: will change the type of this to IdGenerator fix this error?

https://github.com/aws/aws-durable-execution-sdk-python/blob/main/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py#L98

SilanHe:
good catch, the output from the Github action is not visible at all locally actually so I completely missed this

pyproject.toml
```diff
@@ -7,6 +7,7 @@ requires-python = ">=3.11"
 workspace.members = ["packages/*"]
 
 [tool.hatch.envs.test]
+detached = true
```
SilanHe, line 10: interesting, I didn't know about this setting.

[detached = true](https://hatch.pypa.io/1.8/config/environment/overview/#detached-environments)

 In this case, I guess we could technically add this for all envs in the pyproject.toml?

SilanHe (approved):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
wangyb-A: nvm the error is on provider type

wangyb-A (approved):

PR #484, opened 2026-06-18
feat(sdk): align create_linear_retry_strategy with create_retry_strategy (parity with JS #652)

### LINKED ISSUE #483: feat(sdk): align create_linear_retry_strategy with create_retry_strategy (parity with JS #652)
## Summary

Bring `create_linear_retry_strategy` to parity with `create_retry_strategy` (and with the JS SDK), by switching it to a config-object API and adding the capabilities it is currently missing: `max_delay` capping, jitter, and `retryable_errors` / `retryable_error_types` filtering. Share the common jitter / error-matching / delay-finalization logic between the two helpers so their behavior cannot drift.

This mirrors the JS change in aws/aws-durable-execution-sdk-js#652.

## Current state (Python)

`create_retry_strategy` already uses a config dataclass with `max_delay`, `jitter_strategy` (default `FULL`), `retryable_errors`, and `retryable_error_types`.

`create_linear_retry_strategy` lags — it is still positional with none of those capabilities:

```python
def create_linear_retry_strategy(
    max_attempts: int = 6,
    initial_delay: Duration | None = None,
    increment: Duration | None = None,
) -> Callable[[Exception, int], RetryDecision]:
    # "No jitter is applied and there is no upper cap on the delay"
```

So today the only way to get jitter, a max-delay cap, or error filtering with linear backoff is to hand-roll a strategy.

## Proposed changes (mirroring JS #652)

- Introduce a `LinearRetryStrategyConfig` (Duration-based) with: `max_attempts`, `initial_delay`, `increment`, `max_delay`, `jitter_strategy`, `retryable_errors`, `retryable_error_types`.
- Defaults aligned with `create_retry_strategy`: `jitter_strategy=FULL`, `max_delay=5 min`, retry-all when neither error filter is set.
- Extract the shared logic used by both helpers (jitter application, retryable-error resolution + matching, and final delay clamp to an integer >= 1 second) into a common module, so the linear and exponential strategies share one implementation.
- Delay calculation: `base = min(initial_delay + increment * (attempts_made - 1), max_delay)`, then apply jitter, then clamp.
- Pin `RetryPresets.linear()` to `jitter=NONE` so its documented deterministic 1s, 2s, 3s, 4s, 5s delays are preserved.

## Breaking change

Same shape as JS #652:

- Signature changes from positional `(max_attempts, initial_delay, increment)` to a config object.
- Default jitter changes from none to `FULL`, so a bare `create_linear_retry_strategy()` now yields jittered delays rather than the deterministic 1s-5s. (`RetryPresets.linear()` stays deterministic via `jitter=NONE`.)

## Reference

- JS PR: https://github.com/aws/aws-durable-execution-sdk-js/pull/652
- Python `create_retry_strategy` is already config-shaped and is the model to follow for the linear helper.

SAY-5:
Closes #483.

### What
`create_linear_retry_strategy` now takes a `LinearRetryStrategyConfig` (Duration-based) and gains `max_delay` capping, jitter (default `FULL`), and `retryable_errors` / `retryable_error_types` filtering, matching `create_retry_strategy`. The shared jitter, error-resolution/matching, and final delay clamp logic is extracted into module-level helpers used by both helpers so they cannot drift.

### Why
`create_retry_strategy` was already config-shaped with those capabilities; the linear helper lagged behind and forced callers to hand-roll a strategy for jitter, a delay cap, or error filtering. This mirrors the JS change in aws/aws-durable-execution-sdk-js#652.

### Notes
This is a breaking change, as called out in the issue: the signature moves from positional `(max_attempts, initial_delay, increment)` to a config object, and a bare `create_linear_retry_strategy()` now applies `FULL` jitter. `RetryPresets.linear()` is pinned to `jitter=NONE` so its documented deterministic 1s-5s delays are preserved.

### Testing
Updated the existing linear tests to the config API and added coverage for the new `max_delay` cap, jitter application, and message/type error filtering. `hatch run test:cov` passes locally for `tests/retries_test.py`.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/retries.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/retries.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import math
 import re
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Generic, TypeVar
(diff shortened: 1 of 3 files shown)


packages/aws-durable-execution-sdk-python/uv.lock
yaythomas, line 1: please exclude from commit :-)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/retries.py
```diff
@@ -71,84 +71,135 @@ def max_delay_seconds(self) -> int:
...
+        else pattern in str(error)
+        for pattern in retryable_errors
+    )
+    is_retryable_error_type: bool = any(
+        isinstance(error, error_type) for error_type in retryable_error_types
+    )
+    return is_retryable_error_message or is_retryable_error_type
+
+
+def _finalize_delay_seconds(base_delay: float, jitter_strategy: JitterStrategy) -> int:
```
yaythomas: this could move to JitterStrategy and then it's nicely encapsulated in the Jitter object...

```
   # on JitterStrategy
   def finalize_delay(self, base_delay: float) -> int:
       return max(1, math.ceil(self.apply_jitter(base_delay)))
```

yaythomas (changes requested):

yaythomas (approved):

Issue #262, opened 2026-01-19
[Bug]: Add TIMED_OUT terminal state


yaythomas:
### Expected Behavior

On logging replay, a TIMED_OUT state should be marked as complete to transition to "logging enabled" status.

### Actual Behavior

Currently, only the following terminal statuses are included:
https://github.com/aws/aws-durable-execution-sdk-python/blob/afd40835f66a4d2267d4397031cb115e6556f65c/src/aws_durable_execution_sdk_python/state.py#L312-L317

### Steps to Reproduce

1. An operation with a timeout - i.e `invoke` or `callback` times out
2. Logging should enable on replay after the timeout

### SDK Version

1.0

### Python Version

3.14

### Is this a regression?

No

PR #488, opened 2026-06-24
feat: context aware replay status

wangyb-A:
*Issue #, if available:* first piece of #389

## Summary

Track replay status per-context instead of as a single execution-wide flag.

## Changes

- **`context.py`**: `DurableContext` owns its replay status, seeded from its
  creator and refined per-operation via a `_replay_aware()` context manager.
  Added public `context.is_replaying()`.
- **`logger.py`**: `Logger` consults a per-context `is_replaying` callable
  instead of holding `ExecutionState`.
- **`state.py`**: removed the global replay machinery (`track_replay`,
  `is_replaying`, `_visited_operations`); added `has_prior_operations()` for
  execution-level first-invocation detection.
- **`execution.py`**: seeds the root context's status; uses
  `has_prior_operations()` for the plugin `is_first_invocation` flag.
- **`concurrency/executor.py`**: branches track their own status; removed the
  per-branch `track_replay` call.

## Replay boundary behavior

`_replay_aware()` flips REPLAY→NEW at three points:
- **before** an op with no checkpoint (brand-new code),
- **after** a non-terminal op (the resume point — its own logs stay deduped),
- **after** a completed op when the next op doesn't exist yet (so bare logs
  immediately after a `wait` are treated as new, not suppressed).

The third case fixes a regression where a log right after a `wait` was
silently dropped on replay.

## Testing
Logging from examples:

### Invocation 1 (requestId 8b1e9590) — first run, suspends at parent wait
| message                        | is_replaying | operationName |
|--------------------------------|--------------|---------------|
| Workflow started (before wait) | false        | —             |
| Preparing item                 | —            | prepare       |
| Prepared, about to wait        | false        | —             |

### Invocation 2 (requestId a46ca6bd) — parent wait fired
| message                                    | is_replaying / child_is_replaying |
|--------------------------------------------|-----------------------------------|
| Resumed after wait                         | false                             |
| Auditing in child context (before child wait) | false (child)                 |
*(before-wait lines from invocation 1 are NOT repeated — de-duplicated)*

### Invocation 3 (requestId 22d99bc2) — child wait fired
| message                                   | child_is_replaying | operationName |
|-------------------------------------------|--------------------|---------------|
| Resumed in child context (after child wait) | false            | —             |
| Finalizing item                           | —                  | finalize      |
| Workflow completed                        | —                  | —             |
*(all parent + child before-wait lines NOT repeated — de-duplicated)*

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
@@ -489,7 +489,6 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
                 is_virtual=is_virtual,
             ),
         )
-        child_context.state.track_replay(operation_id=operation_id)
         return result
 
     def replay(self, execution_state: ExecutionState, executor_context: DurableContext):
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
@@ -475,6 +475,27 @@ class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
             name=name,
         )
 
+        # The branch/iteration container op is resolved here via child_handler,
+        # bypassing context.run_in_child_context and therefore the parent's
+        # `_replay_aware`. Replicate the two things `_replay_aware` would have
+        # done for this container op while replaying:
+        #   1. Existence flip: a brand-new branch (no checkpoint) is new work,
+        #      so the child must start in NEW rather than inheriting REPLAY —
+        #      otherwise logs before the branch's first inner op are wrongly
+        #      de-duplicated during a map/parallel replay.
+        #   2. Replay hook: a branch that already has a checkpoint was observed
+        #      in a prior invocation, so emit the plugin replay hook (once).
+        # Virtual (FLAT) branches do not checkpoint themselves, so neither
+        # applies; their inner operations still self-correct via `_replay_aware`.
+        if not is_virtual and child_context.is_replaying():
+            branch_checkpoint = child_context.state.get_checkpoint_result(operation_id)
+            if not branch_checkpoint.is_existent():
+                child_context._set_replay_status_new()  # noqa: SLF001
+            elif branch_checkpoint.operation is not None:
+                child_context.state.emit_operation_replay_hook(
+                    branch_checkpoint.operation
+                )
+
         def run_in_child_handler() -> ResultType:
             return self.execute_item(child_context, executable)
 
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
@@ -481,32 +481,62 @@ class DurableContext(DurableContextProtocol):
         ).is_existent()
 
     @contextmanager
-    def _replay_aware(self):
+    def _replay_aware(self, *, executes_user_code: bool = False):
         """Wrap a single operation with replay-boundary detection.
 
-        The boundary has three parts:
+        Args:
+            executes_user_code: True for operations that invoke a user-provided
+                function on (re-)entry — `step` and `wait_for_condition`'s check
+                (and, transitively, `wait_for_callback`'s submitter, which runs
+                as a step). When such an operation actually runs the user
+                function it is, by definition, doing new work (a cached
+                SUCCEEDED operation returns its checkpoint without invoking the
+                function). So a non-terminal user-code operation flips the
+                context to NEW *before* its body runs, including on retries. For
+                all other operations (default False), a non-terminal next
+                operation is a pure resume point with no user body, so we keep
+                replay status through it and flip afterwards.
+
+        The boundary has these parts:
 
         - Existence flip (before the op): if we are replaying and the next
           operation has no checkpoint at all, it is brand-new code, so flip to
           NEW immediately so the operation and its logs count as new.
-        - Deferred status flip (after the op): if we are replaying and the next
-          operation exists but is NOT terminal, that operation is the resume
-          point. We keep replay status through the operation and flip to NEW
-          afterwards, so the resuming operation's own logs stay de-duplicated
-          but subsequent code counts as new.
+        - User-code flip (before the op): if `executes_user_code` and the next
+          operation is non-terminal (brand-new OR retrying/re-executing), the
+          user function is about to run, so flip to NEW before it.
+        - Deferred status flip (after the op): for non-user-code operations, if
+          we are replaying and the next operation exists but is NOT terminal,
+          that operation is the resume point. We keep replay status through the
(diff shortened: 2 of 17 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -420,6 +449,75 @@ def _create_step_id(self) -> str:
...
+          but subsequent code counts as new.
+        - Post-op existence flip (after the op): if we are still replaying once
+          the operation completes and the *following* operation does not exist
+          yet, we have reached the replay boundary.
+        """
+        was_replaying: bool = self.is_replaying()
+        # Status: exists but not terminal -> defer flip until after the op runs.
+        flip_after: bool = (
+            was_replaying
+            and self._next_operation_exists()
```
zhongkechen: should we check updatedOperationIds field instead?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -420,6 +449,75 @@ def _create_step_id(self) -> str:
...
+            and not self._next_operation_is_terminal_checkpoint()
```
zhongkechen: So we flip replay status after a Step operation if Step is going to be retried. And the logger in StepContext will have the same replay status as its parent context, which is in REPLAY in this case, suppressing the logs from the step when it's retried. Is this the desired behavior?

wangyb-A: I think we should not supress the log when the step / user code is retried. Resolve in new revision

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
wangyb-A: Planning to use updatedOperationIds with plugin change in a following up pr

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -481,32 +481,62 @@ def _next_operation_exists(self) -> bool:
         ).is_existent()
 
     @contextmanager
-    def _replay_aware(self):
+    def _replay_aware(self, *, executes_user_code: bool = False):
```
zhongkechen: is this true for context operations?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -481,32 +481,62 @@ def _next_operation_exists(self) -> bool:
...
+            and not next_terminal
         )
-        # Existence: brand-new next op -> flip before the op runs.
-        if was_replaying and not self._next_operation_exists():
+        # Before-the-op flips:
+        # - brand-new next op (no checkpoint): always flip to NEW.
+        # - user-code op that is non-terminal (brand-new or retrying): the user
+        #   function is about to run real work, so flip to NEW before it.
+        if was_replaying and (
+            not next_exists or (executes_user_code and not next_terminal)
```
zhongkechen: why don't we always flip before the op?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
wangyb-A: it should be false, 
For example, if there's a wait in the child operation. If we flip the status to NEW before we enter the child context body, the log before wait will be printed, but it shouldn't

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
wangyb-A: please see the [reply](https://github.com/aws/aws-durable-execution-sdk-python/pull/488#discussion_r3483492206)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
zhongkechen: So this is only true for Steps? Confusing as the name `executes_user_code` implies any user code include context/step/submitter/checker/etc

zhongkechen:
Tests failing

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -420,6 +449,91 @@ def _create_step_id(self) -> str:
...
+        # Only peek when replaying; avoids unnecessary checkpoint lookups (and
+        # any step-id side effects) on the common non-replay path.
+        next_checkpoint: CheckpointedResult | None = (
+            self._peek_next_checkpoint() if was_replaying else None
+        )
+        next_exists: bool = (
+            next_checkpoint.is_existent() if next_checkpoint is not None else False
+        )
+        next_terminal: bool = next_checkpoint is not None and (
+            next_checkpoint.is_succeeded() or next_checkpoint.is_failed()
```
yaythomas: The old state.track_replay treated the completed set as {SUCCEEDED, FAILED, CANCELLED, STOPPED, TIMED_OUT}, and issue #262 ([Bug]: Add TIMED_OUT terminal state) explicitly added TIMED_OUT so logging re-enables on replay after an invoke/callback timeout. 

Now that only SUCCEEDED/FAILED are checked, does this change behaviour for TIMED_OUT/CANCELLED/STOPPED?

My worry is that flip_after treats a non-terminal-but-existing op as "a pure resume point with no user body" (per the docstring), and assumes nothing completed after it.

if this is a concern, maybe adding an `is_terminal`
```
# on CheckpointedResult in state.py
def is_terminal(self) -> bool:
    """Return True if the checkpointed operation is in any terminal status."""
    op = self.operation
    if not op:
        return False
    return op.status in _TERMINAL_OPERATION_STATUSES  # SUCCEEDED, FAILED, TIMED_OUT, CANCELLED, STOPPED
```

and then here:

```
# in _replay_aware
next_terminal: bool = next_checkpoint is not None and next_checkpoint.is_terminal()
```

wangyb-A: add in next rev :D

wangyb-A:
need to merge the change of #496

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
zhongkechen, line 502: Do we need wrap this call with `self._replay_aware()` like the other operations in context?

zhongkechen (approved):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
wangyb-A: no, the replay status is already checked and flipped above.
executor call child_handler directly (skip context) so we cannot directly reuse replay_aware annotation

PR #492, opened 2026-06-25
fix: don't surface suspend as a FAILED outcome

### LINKED ISSUE #491: OTEL plugin marks child-context spans as faults when they suspend (timed wait misclassified as FAILED)
## Summary

In the experimental OpenTelemetry plugin (the durable SDK `plugins` parameter is still provisional and emits a `FutureWarning`), the instrumentation records a false fault whenever a durable execution suspends inside a child context. For example, a child context that calls `context.wait(...)` produces a `child-context` span that X-Ray marks as a fault, with a recorded exception named `TimedSuspendExecution`. The execution itself behaves correctly. A suspend is normal durable control flow, not a failure, so it should not produce a fault.

## Impact

Every child context that contains a wait (or any other suspend) emits a false fault into the trace. This inflates error and fault counts in CloudWatch and X-Ray and pollutes the service map. Concurrent branches (map and parallel) run as child contexts, so they are affected the same way. Top-level waits and plain steps are not affected.

## Reproduction

1. Deploy a durable function with the OTEL plugin, the ADOT layer, and X-Ray active tracing.
2. Use a handler with a child context that waits:

```python
@durable_with_child_context
def child_workflow(ctx):
    ctx.step(do_work(), name="child-step-1")
    ctx.wait(Duration.from_seconds(5), name="child-wait")
    return ctx.step(do_more(), name="child-step-2")

@durable_execution(plugins=[OtelPlugin()])
def handler(event, context):
    return context.run_in_child_context(child_workflow(), name="child-context")
```

3. Invoke once (a qualified version is required for durable functions) and open the trace in CloudWatch Traces.

**Observed:** the `child-context` span is marked `fault: true` with a recorded exception `TimedSuspendExecution`.
**Expected:** the `child-context` span closes cleanly with no fault, the same way the top-level invocation and wait spans do.

## Root cause

The defect is in the core SDK instrumentation plumbing, not in the OTEL plugin. The plugin correctly renders whatever outcome it is handed.

1. `state.py` `wrap_user_function` catches the suspend and reports it to plugins as an error whose type is the concrete subclass name:

```python
except SuspendExecution as e:
    self._plugin_executor.on_user_function_end(
        start_info,
        ErrorObject(type=type(e).__name__, ...),  # "TimedSuspendExecution"
    )
    raise
```

2. `plugin.py` `UserFunctionOutcome.from_error` classifies the outcome by matching the base class name only:

```python
elif error.type == SuspendExecution.__name__:  # "SuspendExecution"
    return cls(cls.PENDING)
else:
    return cls(cls.FAILED)
```

`"TimedSuspendExecution" != "SuspendExecution"`, so the outcome falls through to `FAILED`. Since `TimedSuspendExecution` is the type that every timed `wait` raises, the PENDING branch is never reached for the common case. The OTEL plugin then maps `FAILED` to `set_status(ERROR)` plus `record_exception`, which X-Ray shows as a fault.

The top-level path avoids this because a top-level suspend is handled at the invocation level and resolves to `InvocationStatus.PENDING`. Only the `wrap_user_function` path (child contexts, branches) hits the misclassification.

## Why it is name-based

`ErrorObject` is the serialized form of an error and carries only a `type` name string, so `from_error` cannot do an `isinstance` check against the class hierarchy. Matching the base class name alone does not recognize subclasses such as `TimedSuspendExecution`.

## Suggested fix

Decide the outcome where the live exception still exists (the `except SuspendExecution` clause), rather than reconstructing class identity from a name string downstream. A suspend should be reported with a `PENDING` outcome and no error. This is an internal-only change. The public plugin interface already exposes `UserFunctionEndInfo.outcome`, and `PENDING` already exists, so no public API change is needed. One concrete shape:

- Add an internal `PluginExecutor.on_user_function_suspend(start_info)` that dispatches a `UserFunctionEndInfo` with `outcome=PENDING` and `error=None` through the existing public `on_user_function_end`.
- Route the `except SuspendExecution` branch in `wrap_user_function` to it.
- Simplify `from_error` to `None -> SUCCEEDED` else `FAILED`, removing the fragile name match.

## Affected versions

The plumbing dates to the plugin interface (PR #371). The behavior is present on current `main`.

yaythomas:
Closes #491

## Problem

In the experimental OpenTelemetry plugin, a durable execution that suspended inside a child context (for example a child context that called `context.wait(...)`) produced a false fault on the `child-context` span in X-Ray, with a recorded exception named `TimedSuspendExecution`. The execution itself behaved correctly; only the instrumentation labeling was wrong.

## Root cause

The core SDK reported a suspend to plugins as an `ErrorObject` carrying the concrete exception name (`TimedSuspendExecution`). `UserFunctionOutcome.from_error` matched only the base class name (`SuspendExecution`), so the suspend fell through to `FAILED` instead of being recognized. The OTEL plugin faithfully rendered `FAILED` as a span error.

## Change

- `state.py`: the `except SuspendExecution` branch now just re-raises. It no longer notifies plugins of the suspend.
- The OTEL plugin's existing `on_invocation_end` sweep already iterates remaining open operation spans and ends them cleanly with no status set. That sweep now also handles the suspended context span, which ends as a non-fault span with no recorded exception. This matches how the JS SDK's OTEL plugin handles the same case.
- The OTEL plugin's `on_user_function_end` PENDING branch is removed (it can no longer be reached, since suspends no longer dispatch through the end hook).
- `UserFunctionOutcome` is reduced to `SUCCEEDED` and `FAILED`. `from_error` is simplified to `None → SUCCEEDED`, else `FAILED`. The fragile name-match against `SuspendExecution.__name__` is gone.
- `durable.attempt.number` and `durable.attempt.outcome` are no longer emitted on `CONTEXT` spans. These per-attempt fields are meaningful for `STEP` (each retry is an attempt) but not for `CONTEXT`. Dropping them aligns the emitted attributes across SDKs. `STEP` spans are unchanged.

No public plugin API change: `UserFunctionOutcome` retains the two values plugins already used in practice; the `PluginExecutor` end hook is unchanged. The `from __future__ import annotations` cleanup in `plugin.py` (previously a separate commit) is folded in.

## Testing

- New regression `test_wrap_user_function_suspend_does_not_fire_end_hook` asserts the end hook is never invoked when a wrapped function raises `TimedSuspendExecution`. On `main`, the end hook fires with `outcome=FAILED`, so this test would fail; with this change, `captured` stays empty.
- New OTEL test `test_context_span_omits_attempt_attributes` asserts a CONTEXT span carries neither `durable.attempt.number` nor `durable.attempt.outcome`.
- `TestUserFunctionOutcomeValues` pins the enum membership to `{SUCCEEDED, FAILED}` to catch accidental regressions.
- Core SDK suite: 1275 passed. OTEL package suite: 49 passed. Typecheck clean (75 source files). `hatch fmt --check` clean across both packages.

## Verified on a deployed durable function

The fix has been deployed to a test Lambda (us-west-2) with a child context that contains a wait. Three invocations of one execution, fresh X-Ray trace: all operation spans (including the suspended `child-context`, its `child-wait`, and the continuation `child-context` on the resuming invocation) close cleanly with no fault and no recorded exception. JS-SDK demo function in the same account produces the equivalent non-fault span tree for cross-SDK confirmation.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import contextlib
 import datetime
 import functools
@@ -7,7 +9,6 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Callable, MutableMapping
 
-from aws_durable_execution_sdk_python.exceptions import SuspendExecution
 from aws_durable_execution_sdk_python.identifier import OperationIdentifier
 from aws_durable_execution_sdk_python.lambda_service import (
     DurableExecutionInvocationOutput,
@@ -51,16 +52,12 @@ class OperationEndInfo(OperationInfo):
 class UserFunctionOutcome(Enum):
     SUCCEEDED = "SUCCEEDED"
     FAILED = "FAILED"
-    PENDING = "PENDING"
 
     @classmethod
-    def from_error(cls, error: ErrorObject | None) -> "UserFunctionOutcome":
+    def from_error(cls, error: ErrorObject | None) -> UserFunctionOutcome:
         if error is None:
             return cls(cls.SUCCEEDED)
-        elif error.type == SuspendExecution.__name__:
-            return cls(cls.PENDING)
-        else:
-            return cls(cls.FAILED)
+        return cls(cls.FAILED)
 
 
 @dataclass(frozen=True)
@@ -86,7 +83,7 @@ class UserFunctionEndInfo(OperationInfo):
     @classmethod
     def from_start_info(
         cls, start_info: UserFunctionStartInfo, error: ErrorObject | None
-    ) -> "UserFunctionEndInfo":
+    ) -> UserFunctionEndInfo:
         return UserFunctionEndInfo(
             operation_id=start_info.operation_id,
             operation_type=start_info.operation_type,
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
@@ -453,11 +453,8 @@ class OtelPlugin(DurableInstrumentationPlugin):
                     else "Unknown error"
                 )
             )
-        elif info.outcome is UserFunctionOutcome.SUCCEEDED:
-            span.set_status(StatusCode.OK)
         else:
-            # PENDING
-            span.set_status(StatusCode.UNSET, "PENDING")
+            span.set_status(StatusCode.OK)
 
         end_timestamp = info.end_time
         if end_timestamp is not None and end_timestamp == info.start_time:
@@ -493,9 +490,13 @@ class OtelPlugin(DurableInstrumentationPlugin):
             attributes["durable.operation.type"] = info.operation_type.value
         if hasattr(info, "name") and info.name is not None:
             attributes["durable.operation.name"] = info.name
-        if hasattr(info, "attempt") and info.attempt is not None:
-            attributes["durable.attempt.number"] = info.attempt
-        if hasattr(info, "outcome") and info.outcome is not None:
-            attributes["durable.attempt.outcome"] = info.outcome.value
+        # Per-attempt fields are meaningful for STEP (each attempt is retried)
+        # but not for CONTEXT (a context is entered once per invocation, not
+        # retried). Omit them on CONTEXT spans for cross-SDK consistency.
+        if getattr(info, "operation_type", None) is not OperationType.CONTEXT:
+            if hasattr(info, "attempt") and info.attempt is not None:
+                attributes["durable.attempt.number"] = info.attempt
+            if hasattr(info, "outcome") and info.outcome is not None:
+                attributes["durable.attempt.outcome"] = info.outcome.value
 
         return attributes
--- a/packages/aws-durable-execution-sdk-python-otel/tests/test_plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/tests/test_plugin.py
@@ -269,6 +269,51 @@ def test_user_function_callbacks_emit_attempt_span_attributes():
     )
 
 
+def test_context_span_omits_attempt_attributes():
+    """CONTEXT operations do not carry per-attempt attributes.
+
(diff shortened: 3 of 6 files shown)


SilanHe (dismissed):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -101,6 +99,24 @@ def from_start_info(
...
+        return cls(
+            operation_id=start_info.operation_id,
+            operation_type=start_info.operation_type,
+            sub_type=start_info.sub_type,
+            name=start_info.name,
+            parent_id=start_info.parent_id,
+            start_time=start_info.start_time,
+            is_replay_children=start_info.is_replay_children,
+            attempt=start_info.attempt,
+            outcome=UserFunctionOutcome.PENDING,
```
ayushiahjolia: On SuspendExecution, SDK calls CheckpointDurableExecution with STARTED, so we might want to keep this Started, and later it will emit completed from list of operations from backend.

yaythomas:
Updated this PR with a different fix approach after review.

The previous version explicitly signaled the suspend to plugins with a new `UserFunctionOutcome.PENDING` value and an `on_user_function_suspend` hook. Instead, this will now align with the JS SDK's mechanism instead: don't fire the end hook on suspend at all; let the OTEL plugin's existing `on_invocation_end` sweep close any remaining open operation spans cleanly.

This drops the public `PENDING` outcome, simplifies the plumbing, and matches JS exactly.

Same observable fix (no fault on the suspended child-context span). New approach is verified end-to-end on a deployed test Lambda, all operation spans close non-fault, JS demo produces the equivalent span tree.


Expectation | Result
-- | --
Execution completes correctly | {"a":"a","b":"b","child":"c1-c2"} ✓
All spans closed | 46/46, 0 unclosed ✓
No fault on any plugin span (incl. suspended child-context) | NONE ✓
CONTEXT spans have NO durable.attempt.number | NONE ✓
CONTEXT spans have NO durable.attempt.outcome | NONE ✓
STEP spans STILL have durable.attempt.number | step-a, step-b, child-step-1, child-step-2 ✓
STEP spans STILL have durable.attempt.outcome | same four ✓

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
yaythomas: instead of STARTED, we will follow JS and not emit this record.

SilanHe (approved):

PR #493, opened 2026-06-26
ci: publish durable execution lambda layer (EXPERIMENTAL)

zhongkechen:
## Summary
- add a GitHub workflow to build and publish public Lambda layer versions for the SDK + OTel plugin
- add a repo script for building runtime/architecture-specific layer zips
- document manual and release-driven layer publishing

## Testing
- hatch fmt --check .github/scripts packages/aws-durable-execution-sdk-python-otel
- hatch run test:all .github/scripts/tests/test_build_lambda_layer.py .github/scripts/tests/test_parse_sdk_branch.py packages/aws-durable-execution-sdk-python-otel/tests
- hatch run types:check
- hatch build (OTel package)
- python3 .github/scripts/check_dist_legal_files.py packages/aws-durable-execution-sdk-python-otel
- workflow YAML parse check

### DIFF
--- a/.github/workflows/lambda-layer-publish.yml
+++ b/.github/workflows/lambda-layer-publish.yml
@@ -1,6 +1,13 @@
 name: Publish Lambda Layer
 
 on:
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/scripts/build_lambda_layer.py"
+      - ".github/workflows/lambda-layer-publish.yml"
+      - "packages/aws-durable-execution-sdk-python/**"
+      - "packages/aws-durable-execution-sdk-python-otel/**"
   release:
     types: [published]
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


.github/workflows/lambda-layer-publish.yml
```diff
@@ -0,0 +1,152 @@
+name: Publish Lambda Layer
+
+on:
+  pull_request:
```
yaythomas, line 4: this just for testing? 

rather than triggering on release?

.github/workflows/lambda-layer-publish.yml
```diff
@@ -0,0 +1,152 @@
...
+  workflow_dispatch:
+    inputs:
+      regions:
+        description: "Comma-separated AWS regions to publish to"
+        required: false
+        default: "us-east-1"
+
+permissions:
+  contents: read
+  id-token: write
```
yaythomas, line 22: it doesn't really matter all that much given there's only 1 job in here currently, but generally the pattern we follow is `read` at the top level and write only on the jobs that need it.

zhongkechen:
Layer is for internal testing only

Issue #479, opened 2026-06-17
Add FileSystem serdes


dhegberg:
Add FileSystem serdes for parity with JavaScript:
https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/utils/serdes/filesystem-serdes.ts

PR #494, opened 2026-06-26
feat(serdes): Add filesystem serdes for large payload storage

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/479

*Description of changes:*
- Adds `create_filesystem_serdes()` - a SerDes implementation that stores serialized values on a durable filesystem (Amazon S3 Files or Amazon EFS mounted to Lambda). This keeps checkpoint sizes small regardless of payload
size, with optional inline preview for observability.

Features - 

- **ALWAYS mode**: Every value written to file; checkpoint stores file pointer
- **OVERFLOW mode**: Inline if < 256KB, overflow to file if larger
- **URI path encoding**: Human-readable paths derived from ARN (`<functionName>/<executionName>/<invocationId>/<operationId>.json`)
- **HASH path encoding**: Fixed-length SHA-256 paths for safety
- **Preview**: Optional inline preview in checkpoint envelope for console/API visibility without reading the full file
- **ARN parsing**: Compact directory structure from durable execution ARNs

Tests -
- Added unit tests, e2e tests with replay and integration tests.

Next PRs - 
- Cloud integration tests (requires S3 Files or EFS mount infrastructure)
- Documentation update in `aws-durable-execution-docs` repo

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
@@ -1,0 +1,265 @@
+"""Filesystem-based serialization for durable functions.
+
+This module provides a SerDes implementation that stores serialized values on a
+durable filesystem (Amazon S3 Files or Amazon EFS mounted to Lambda).
+
+WARNING: Do NOT use with Lambda's ephemeral /tmp storage. Lambda's /tmp is local
+to a single execution environment and is not shared across invocations. On replay,
+a different environment may be used and the file will not be found.
+
+Use only with a durable, shared filesystem such as:
+- Amazon S3 Files — mount an S3 bucket as a filesystem via the Lambda console or IaC
+- Amazon EFS — mount an EFS file system to your Lambda function
+
+Key Features:
+- Two storage modes: ALWAYS (write every value to file) and OVERFLOW (inline if
+  small, overflow to file if exceeds checkpoint size limit)
+- Two path encoding modes: URI (human-readable) and HASH (fixed-length, safe)
+- Optional preview support for storing compact summaries inline in checkpoints
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+from dataclasses import dataclass
+from enum import StrEnum
+from typing import Any, Protocol
+from urllib.parse import quote
+
+from aws_durable_execution_sdk_python.constants import CHECKPOINT_SIZE_LIMIT_BYTES
+from aws_durable_execution_sdk_python.preview import build_preview
+from aws_durable_execution_sdk_python.serdes import (
+    EXTENDED_TYPES_SERDES,
+    SerDes,
+    SerDesContext,
+)
+from aws_durable_execution_sdk_python.types import DurableExecutionArn
+
+# Subtract 1KB headroom for the envelope wrapper and other checkpoint metadata
+_OVERFLOW_THRESHOLD_BYTES = CHECKPOINT_SIZE_LIMIT_BYTES - 1024
+
+
+class FileSystemSerDesMode(StrEnum):
+    """Controls when data is written to the filesystem.
+
+    - ALWAYS: Every value is written to a file; the checkpoint stores only a
+      file pointer. Best for consistently large payloads or when you want
+      predictable checkpoint sizes.
+
+    - OVERFLOW: Data is written inline (as JSON) unless it exceeds the durable
+      function checkpoint size limit (~256KB), in which case it overflows to a
+      file. Best for mixed workloads where most payloads are small.
+    """
+
+    ALWAYS = "ALWAYS"
+    OVERFLOW = "OVERFLOW"
+
+
+class FileSystemPathEncoding(StrEnum):
+    """Controls how the durable execution ARN and operation ID are turned into
+    on-disk directory and file names.
+
+    - URI: The per-execution directory is a compact, human-navigable path built
+      from the ARN's function name, execution name and invocation id
+      (<functionName>/<executionName>/<invocationId>); the file name is the
+      operation ID percent-encoded. Names stay readable, but a very long
+      operation ID may exceed the filesystem's per-name length limit (commonly
+      255 bytes). If the ARN does not match the expected durable-execution shape,
+      the whole ARN is percent-encoded into a single directory segment instead.
+
+    - HASH: The ARN (directory) and operation ID (file name) are each replaced
+      by their SHA-256 hex digest. Names are a fixed length (64 chars) and always
+      filesystem-safe regardless of the characters or length of the original
+      value, at the cost of no longer being human-readable.
+    """
+
+    URI = "URI"
+    HASH = "HASH"
+
+
+class GeneratePreview(Protocol):
+    """Protocol for preview generation functions."""
+
+    def __call__(self, value: Any) -> dict[str, Any] | None: ...
+
+
+@dataclass(frozen=True)
+class FileSystemSerDesConfig:
+    """Configuration options for FileSystemSerDes.
+
+    Attributes:
+        storage_mode: Controls when data is written to the filesystem.
+            Default: FileSystemSerDesMode.ALWAYS
+        path_encoding: Controls how the ARN (directory) and operation ID (file
+            name) are encoded into path segments. Default: FileSystemPathEncoding.URI
+        generate_preview: Optional function that generates a preview object from
+            the value. When provided, the preview is stored inline in the
+            checkpoint envelope alongside the file pointer, making data visible
+            in the console and API without reading the full file.
+        serdes: SerDes instance used to serialize/deserialize values written to
+            the filesystem. Default: ExtendedTypeSerDes (supports datetime,
+            Decimal, UUID, bytes, tuple, and all primitive types).
+    """
+
+    storage_mode: FileSystemSerDesMode = FileSystemSerDesMode.ALWAYS
+    path_encoding: FileSystemPathEncoding = FileSystemPathEncoding.URI
+    generate_preview: GeneratePreview | None = None
+    serdes: SerDes[Any] = EXTENDED_TYPES_SERDES
+
+
+def _encode_segment(value: str, encoding: FileSystemPathEncoding) -> str:
+    """Encode a path segment (ARN or operation ID) into a filesystem-safe name.
+
+    For URI encoding, uses percent-encoding of all characters except unreserved
+    (RFC 3986). For HASH encoding, uses SHA-256 hex digest.
+    """
+    if encoding is FileSystemPathEncoding.HASH:
+        return hashlib.sha256(value.encode()).hexdigest()
+    return quote(value, safe="")
+
+
+def _resolve_execution_dir(
+    base_path: str,
+    arn: str,
+    path_encoding: FileSystemPathEncoding,
+) -> str:
+    """Resolve the per-execution directory under base_path.
+
+    In URI mode, derives a compact human-navigable path from the execution's
+    function name, execution name and invocation id. If the ARN doesn't match
+    the expected shape, the whole ARN is URI-encoded into a single segment.
+
+    In HASH mode, the whole ARN is hashed into a single fixed-length segment.
+    """
+    if path_encoding is FileSystemPathEncoding.URI:
+        parsed = DurableExecutionArn.from_arn(arn)
+        if parsed:
+            return os.path.join(
+                base_path,
+                parsed.function_name,
+                parsed.execution_name,
+                parsed.invocation_id,
+            )
+    return os.path.join(base_path, _encode_segment(arn, path_encoding))
+
+
+def _write_to_file(
+    base_path: str,
+    value: Any,
+    context: SerDesContext,
+    path_encoding: FileSystemPathEncoding,
+    value_serdes: SerDes[Any],
+) -> str:
+    """Write value as serialized string to a file and return the file path."""
+    dir_path: str = _resolve_execution_dir(
+        base_path, context.durable_execution_arn, path_encoding
+    )
+    os.makedirs(dir_path, exist_ok=True)
+    file_name: str = f"{_encode_segment(context.operation_id, path_encoding)}.json"
+    file_path: str = os.path.join(dir_path, file_name)
+    serialized: str = value_serdes.serialize(value, context)
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write(serialized)
+    return file_path
+
+
+class FileSystemSerDes(SerDes[Any]):
+    """SerDes that stores values on a durable filesystem.
+
+    WARNING: Do NOT use with Lambda's ephemeral /tmp storage.
+    Lambda's /tmp filesystem is local to a single execution environment and is
+    not shared across invocations or function instances. On replay, a different
+    execution environment may be used and the file will not be found, causing
+    deserialization to fail.
+
+    Use only with a durable, shared filesystem such as:
+    - Amazon S3 Files — mount an S3 bucket as a filesystem via the Lambda
+      console or IaC
+    - Amazon EFS — mount an EFS file system to your Lambda function
+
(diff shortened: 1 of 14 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+
+def _encode_segment(value: str, encoding: FileSystemPathEncoding) -> str:
+    """Encode a path segment (ARN or operation ID) into a filesystem-safe name.
+
+    For URI encoding, uses percent-encoding of all characters except unreserved
+    (RFC 3986). For HASH encoding, uses SHA-256 hex digest.
+    """
+    if encoding == FileSystemPathEncoding.HASH:
+        return hashlib.sha256(value.encode()).hexdigest()
+    from urllib.parse import quote
```
yaythomas: inline import. imports at the top pls :-)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+
+    return quote(value, safe="")
+
+
+def _parse_durable_execution_arn(
```
yaythomas: would be nice to have a `DurableExecutionArn` frozen dataclass with a `from_arn`, maybe living inside `types`?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+    )
+    os.makedirs(dir_path, exist_ok=True)
+    file_name = f"{_encode_segment(context.operation_id, path_encoding)}.json"
+    file_path = os.path.join(dir_path, file_name)
+    with open(file_path, "w", encoding="utf-8") as f:
+        json.dump(value, f)
+    return file_path
+
+
+class _FileSystemSerDes(SerDes[Any]):
```
yaythomas: this can be frozen, probably?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+    "PreviewField",
+    "PreviewMode",
+    "build_preview",
+    "create_filesystem_serdes",
+]
+
+logger = logging.getLogger(__name__)
+
+# Checkpoint size limit in bytes (256KB)
+_CHECKPOINT_SIZE_LIMIT_BYTES = 256 * 1024
```
yaythomas: there's already a CHECKPOINT_SIZE_LIMIT in operation/child.py. A shared home e.g. config.py, could be nice

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
```diff
@@ -0,0 +1,199 @@
...
+                collect(item, path_prefix)
+            return
+
+        for key in obj:
+            if key in dangerous_keys:
+                continue
+            if "." in str(key):
+                continue
+
+            path = f"{path_prefix}.{key}" if path_prefix else str(key)
```
yaythomas: local type declaration, pls 

https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#data-structures--typing

> Declare a type definition wherever you declare a variable, even within a function scope and even where it's implied

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
```diff
@@ -0,0 +1,199 @@
...
+            masked = _is_matched(path, config.mask)
+            excluded = _is_matched(path, config.exclude)
+            visible = not excluded and (
```
yaythomas: local types, please. 

maybe just me, but maybe this logic reads easier this way round? what's your view?  
```
visible: boolean = not excluded and (
    masked
    or config.mode == PreviewMode.INCLUDE_ALL
    or _is_matched(path, config.include)
)
```

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+)
+
+
+def _encode_segment(value: str, encoding: FileSystemPathEncoding) -> str:
+    """Encode a path segment (ARN or operation ID) into a filesystem-safe name.
+
+    For URI encoding, uses percent-encoding of all characters except unreserved
+    (RFC 3986). For HASH encoding, uses SHA-256 hex digest.
+    """
+    if encoding == FileSystemPathEncoding.HASH:
```
yaythomas: `is` for enums, rather than ==

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+class GeneratePreview(Protocol):
+    """Protocol for preview generation functions."""
+
+    def __call__(self, value: Any) -> dict[str, Any] | None: ...
+
+
+@dataclass(frozen=True)
+class FileSystemSerdesConfig:
```
yaythomas: `FileSystemSerdesMode`/`FileSystemSerdesConfig`/ use JS "Serdes" casing, but the base class and the private `_FileSystemSerDes` use the repo's SerDes (capital D, as in `JsonSerDes`/`ExtendedTypeSerDes`). Thus far the repo's established public spelling is SerDes.

Might be an idea to add a serdes field to `FileSystemSerdesConfig`, defaulting to the extended types serdes:

```
 config=FileSystemSerdesConfig(
        serdes=EXTENDED_TYPES_SERDES)
```

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import os
+import re
+import sys
```
yaythomas: this used?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+    "FileSystemSerdesConfig",
+    "FileSystemSerdesMode",
+    "PreviewConfig",
+    "PreviewField",
+    "PreviewMode",
+    "build_preview",
+    "create_filesystem_serdes",
+]
+
+logger = logging.getLogger(__name__)
```
yaythomas: this used?

and if not, how come ruff didn't notice, I wonder? 

the package's pyproject.toml has [tool.ruff.lint] select = ["TID252"], does this means ruff is configured to enable only TID252 (the absolute-imports rule) and all default checks, including F401, are turned off?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
```diff
@@ -0,0 +1,199 @@
...
+        config: Preview configuration.
+
+    Returns:
+        A nested dict representing the preview, or None if no fields are visible
+        or value is not a dict/object.
+    """
+    if not isinstance(value, dict):
+        return None
+
+    dangerous_keys = {"__proto__", "constructor", "prototype"}
```
yaythomas: do these matter in python? in js these are sensitive because in JS these keys have special meaning, but not so in python?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if isinstance(obj, list):
+            for item in obj:
+                collect(item, path_prefix)
+            return
+
+        for key in obj:
+            if key in dangerous_keys:
+                continue
+            if "." in str(key):
```
yaythomas, line 142: This existing if "."  guard stays, it is meaningful in python (dot in a key would collide with the dot-path notation used everywhere else in this module in preview.)

Although we should document clearly that `.` carries special meaning in a key name if someone is using preview.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+        if self._storage_mode == FileSystemSerdesMode.ALWAYS:
+            file_path = _write_to_file(
+                self._base_path, value, serdes_context, self._path_encoding
+            )
+            envelope: dict[str, Any] = {"file": file_path}
+            if self._generate_preview:
+                preview = self._generate_preview(value)
+                if preview:
+                    envelope["preview"] = preview
+            return json.dumps(envelope)
```
yaythomas, line 230: this should probably use the `ExtendedTypeSerDes` instead? Else values that work today with the default serdes for backend checkpointing, datetime, Decimal, UUID, bytes, will raise TypeError, and tuple silently becomes list.

The tests don't surface this because they only round-trip dict/list/primitive payloads.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
```diff
@@ -0,0 +1,320 @@
...
+    FieldMatchMode,
+    PreviewConfig,
+    PreviewField,
+    PreviewMode,
+    build_preview,
+)
+from aws_durable_execution_sdk_python.serdes import SerDes, SerDesContext
+
+
+# Re-export preview types so users can import everything from filesystem_serdes
```
yaythomas: Exporting preview from here, while convenient for FileSystemSerdes consumers, will not make sense for future different consumers of preview.

The preview docstring says:
> The preview system is designed to be used with any SerDes implementation — not just the filesystem serdes. For example, you could use build_preview with a custom DynamoDB-backed serdes or any other external storage serdes.

i.e recommend removing the __all__ re-export. This is more a JS idea (to export from a module), in Python everything's public by default.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
ayushiahjolia: Can't use `@dataclass(frozen=True)` here because `_FileSystemSerDes` inherits from `SerDes[Any]`. Either we can add `__slots__` or keep it as is, `_` at the beginning signals it's private.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
ayushiahjolia: will create a separate constants file, config.py has all public types

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
ayushiahjolia: Done, added at some other places as well.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
ayushiahjolia: Fixing name and adding field as well.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
ayushiahjolia: Not used and yes, select = ["TID252"] means ruff only enables that single rule. F401 (unused imports) and all other default checks are off. The CI's hatch fmt --check only enforces formatting (via ruff format) and that one lint rule.

I can followup with enabling F401 or all pyflakes checks if we want.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/preview.py
ayushiahjolia: Removed, added by agent to follow everything JS does :)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/filesystem_serdes.py
yaythomas: Should be do-able, here's an example, unless I'm missing somehing?
```python
"""Minimal demo: @dataclass(frozen=True) inheriting from an ABC + Generic[T].

Run: python3 /tmp/frozen_serdes_repro.py
"""
from abc import ABC, abstractmethod
from dataclasses import FrozenInstanceError, dataclass
from typing import Any, Generic, TypeVar

T = TypeVar("T")


class SerDes(ABC, Generic[T]):
    @abstractmethod
    def serialize(self, value: T) -> str: ...
    @abstractmethod
    def deserialize(self, data: str) -> T: ...


@dataclass(frozen=True)
class FileSystemSerDes(SerDes[Any]):
    base_path: str
    mode: str = "ALWAYS"

    def serialize(self, value: Any) -> str:
        return f"{self.base_path}/{value}"

    def deserialize(self, data: str) -> Any:
        return data


# 1. Construction works.
fs = FileSystemSerDes(base_path="/mnt/s3")
print("1. constructed:", fs)

# 2. Frozen: assignment raises.
try:
    fs.base_path = "/elsewhere"
except FrozenInstanceError as e:
    print("2. frozen confirmed:", e)

# 3. Frozen dataclasses are hashable (auto __hash__).
print("3. hashable:", hash(fs) is not None)

# 4. Abstract method enforcement still works (omitting deserialize fails).
@dataclass(frozen=True)
class Incomplete(SerDes[Any]):
    base_path: str
    def serialize(self, value: Any) -> str:
        return ""
    # deserialize NOT implemented

try:
    Incomplete(base_path="/x")
except TypeError as e:
    print("4. ABC enforced:", e)

```

Now, saying that, I just noticed the existing SerDes here https://github.com/aws/aws-durable-execution-sdk-python/blob/main/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/serdes.py#L371 don't use `frozen`, so might as well keep to the pattern and not freeze it :-) 

Is this marked underscore private for a reason? The others are public:
- JsonSerDes — public, instantiate directly.
- ExtendedTypeSerDes — public, instantiate directly.
- PassThroughSerDes — public, instantiate directly

zhongkechen: I'm afraid adding dataclass to non-`data` classes like `SerDes` could lead to more confusion than benefits.

yaythomas: agree, great points :-) 

remaining to check, is this marked _ private for a reason?

ayushiahjolia: yeah, I was using another factory method create_filesystem_serdes as main method, so we didn't need to make this public. But I removed it now and made this one public.

yaythomas (approved):

PR #497, opened 2026-06-30
chore(lint): Enable Pyflakes rules across all packages

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/479

*Description of changes:*
Enables the Pyflakes (`F`) rule category in ruff across all packages. reviously only `TID252` (ban relative imports) was enforced, meaning unused imports, undefined names, and unused variables went undetected.

Testing - 
- `ci-checks.sh` passes (tests, coverage, typecheck, fmt for all packages)
- `ruff check --select F,TID252` reports zero violations

### DIFF
--- a/packages/aws-durable-execution-sdk-python-examples/cli.py
+++ b/packages/aws-durable-execution-sdk-python-examples/cli.py
@@ -1,0 +1,502 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import logging
+import os
+import shutil
+import subprocess
+import sys
+import time
+import zipfile
+from pathlib import Path
+
+
+# Configure logging
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+logger = logging.getLogger(__name__)
+
+
+try:
+    import boto3
+
+    from aws_durable_execution_sdk_python.lambda_service import LambdaClient  # noqa: F401
+except ImportError:
+    sys.exit(1)
+
+
+def load_catalog():
+    """Load examples catalog."""
+    catalog_path = Path(__file__).parent / "examples-catalog.json"
+    with open(catalog_path) as f:
+        return json.load(f)
+
+
+def build_examples():
+    """Build examples with SDK dependencies."""
+
+    build_dir = Path(__file__).parent / "build"
+    src_dir = Path(__file__).parent / "src"
+    packages_dir = Path(__file__).parent.parent
+
+    logger.info("Building examples...")
+
+    # Clean and create build directory
+    if build_dir.exists():
+        logger.info("Cleaning existing build directory")
+        shutil.rmtree(build_dir)
+    build_dir.mkdir()
+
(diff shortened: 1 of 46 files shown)


packages/aws-durable-execution-sdk-python-examples/cli.py
```diff
@@ -20,7 +20,7 @@
 try:
     import boto3
 
-    from aws_durable_execution_sdk_python.lambda_service import LambdaClient
+    from aws_durable_execution_sdk_python.lambda_service import LambdaClient  # noqa: F401
```
nvasiu, line 23: `# noqa: F401` exempts this line from the unused imports rule, I assume because we want to keep `LambdaClient` available for users importing this?

How did we decide which imports to exempt from this rule? Just wondering if this PR automatically removed anything that we wanted to keep.

ayushiahjolia: 1.  LambdaClient is there as a fail-fast check. It's in the try/except ImportError block so the CLI dies early if the SDK isn't installed, rather than blowing up later.

2. I asked Kiro to update rules in pyproject.toml, which I think enabled rules and removed all unnecessary imports. There was one import that was not directly used in file but was used at runtime, I found it after pushing changes to this PR and added noqa for it (mock-path resolution for the os one in serialization.py).

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
```diff
@@ -22,7 +22,7 @@ target-version = "py311"
 
 [tool.ruff.lint]
 preview = true
-select = ["TID252"]  # Enforce absolute imports (ban relative imports)
+select = ["F", "TID252"]  # Pyflakes (unused imports, undefined names, etc.) + absolute imports
```
zhongkechen: Unused imports in Python might be actually used because the imported symbols may not be used in the current module, instead be used by other modules importing this module

yaythomas:
d2383533b "update ruff config" added: +select = ["TID252"] # Enforce absolute imports (ban relative imports)

This is the regression point. Adding TID252 as the sole entry overrode the default and silently switched off F (and the E4/E7/E9 subset). From here, unused imports/vars accumulated undetected.

The PR could optionally also restore E4/E7/E9 (the rest of the old default) via select = ["E4","E7","E9","F","TID252"], since those were dropped in the same commit.

And a minor pre-existing inconsistency: the testing package sets preview = false while the others set preview = true. Not necessarily in scope for this PR :-)

packages/aws-durable-execution-sdk-python-examples/pyproject.toml
ayushiahjolia: Hmm I relied on tests to verfify this, I found one variable which was used at runtime, but no other issues.

nvasiu (approved):

Issue #405, opened 2026-05-30
[Bug]: unstable map test case


zhongkechen:
### Expected Behavior

unit tests should always succeed

### Actual Behavior

Some unit tests fail frequently

- https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26666529143/job/78600975717?pr=371
- https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/26665947867/job/78599228507

```
_________________________ test_map_with_min_successful _________________________

durable_runner = <test.conftest.TestRunnerAdapter object at 0x7fb315db76d0>

    @pytest.mark.example
    @pytest.mark.durable_execution(
        handler=map_with_min_successful.handler,
        lambda_function_name="Map with Min Successful",
    )
    def test_map_with_min_successful(durable_runner):
        """Test map with min_successful threshold."""
        with durable_runner:
            result = durable_runner.run(input="test", timeout=10)
    
        assert result.status is InvocationStatus.SUCCEEDED
    
        result_data = deserialize_operation_payload(result.result)
    
        # With min_successful=6, operation completes after reaching 6 successes
        # Due to concurrency (max_concurrency=5), some items may complete before check
        # Items 1-6 succeed, item 10 succeeds, items 7-9 fail
        # Depending on timing, we get 6 or 7 successes
        assert result_data["success_count"] >= 6
        assert result_data["success_count"] <= 7
    
        # Operation stops once min_successful is reached
        # Items 7-9 (which would fail) are never processed
        assert result_data["failure_count"] == 0
        assert result_data["total_count"] == 10
    
        # Verify we got the expected successful results
        # Items 1-6 always succeed (2, 4, 6, 8, 10, 12)
        # Item 10 might also succeed (20) depending on timing
        assert len(result_data["results"]) == result_data["success_count"]
        for result_val in result_data["results"]:
            assert result_val % 2 == 0  # All results should be even (item * 2)
            assert result_val >= 2 and result_val <= 20  # Range: items 1-10 * 2
            assert result_val not in [14, 16, 18]  # Items 7-9 should not be present
    
        # Completion reason should be MIN_SUCCESSFUL_REACHED
        assert result_data["completion_reason"] == "MIN_SUCCESSFUL_REACHED"
    
        # Get the map operation
        map_op = result.get_context("map_min_successful")
        assert map_op is not None
        assert map_op.status is OperationStatus.SUCCEEDED
    
[...]

PR #501, opened 2026-06-30
fix(testing): scale local durable timers

zhongkechen:
## Summary
- Add a shared `DURABLE_EXECUTION_TIME_SCALE` helper for local durable test timers.
- Apply the scale to wait timers, step retry timers, callback timeouts, and callback heartbeat timeouts.
- Keep wait operation history faithful to the modeled durable duration: only the local scheduler delay is scaled, while `WaitStartedDetails.ScheduledEndTimestamp` remains unscaled.
- Default local example tests to `0.05` scale so modeled durable delays do not spend unnecessary wall-clock time in CI.
- Keep a 5s floor for long callback timeout/heartbeat timers so externally-driven callback tests do not race unrealistically short deadlines.
- Let heartbeat-focused examples opt out with `time_scale="1.0"` because they intentionally test real heartbeat timing.

## Context
These commits were split out from PR #498. They are not part of the root-cause fix for the scheduler hang or the Python 3.13/3.14 callback race; those remain in PR #498.

This PR keeps the timer scaling change separate so it can be reviewed as a test-duration improvement for local durable timers.

## Changes
- Add `aws_durable_execution_sdk_python_testing.time_scale`.
- Reuse the helper in wait and step retry processors instead of parsing `DURABLE_EXECUTION_TIME_SCALE` inline.
- Scale wait scheduling without scaling the wait history event timestamp.
- Scale callback timeout and heartbeat scheduling in the local test executor.
- Preserve short callback deadlines by using `max(scaled_delay, min(original_delay, 5s))` for long callback timers.
- Add test coverage for scaled wait scheduling with unscaled history, step retry scaling, callback timeout scaling, callback heartbeat scaling, and short-deadline preservation.

## Verification
- `hatch fmt --check`
- `hatch run types:check`
- `git diff --check`
- `hatch run test:all packages/aws-durable-execution-sdk-python-testing/tests/checkpoint/processors/wait_test.py packages/aws-durable-execution-sdk-python-testing/tests/checkpoint/processors/step_test.py packages/aws-durable-execution-sdk-python-testing/tests/executor_test.py -q` - 138 passed
- Previous heartbeat example verification: `hatch run test:all packages/aws-durable-execution-sdk-python-examples/test/callback/test_callback_heartbeat.py packages/aws-durable-execution-sdk-python-examples/test/wait_for_callback/test_wait_for_callback_heartbeat.py -q` - 2 passed


SilanHe:
I'm not against this change since you're already done the work but barring more comments, I think it would be better to just port the "skip time" functionality from the other testing libraries

zhongkechen:
> I'm not against this change since you're already done the work but barring more comments, I think it would be better to just port the "skip time" functionality from the other testing libraries

Time scaling is already used in waits. I just extended it to step retries and other operations that have delays. It is different from Java i think because the way python test cases interact with the execution is different. In Java  local test runner, the runner stops execution after each invocation and waits for user interaction while pythons runner runs execution concurrently when user interacts with the execution.

yaythomas:
There's a pre-existing issue here (not your fault!) where Wait timers leak scaled time into the observable history, when arguably the event history should contain the un-scaled time. 

```
scaled_wait_seconds = scale_delay(wait_seconds)
scheduled_end_timestamp = datetime.now(UTC) + timedelta(seconds=scaled_wait_seconds)
wait_details = WaitDetails(scheduled_end_timestamp=scheduled_end_timestamp)
```

The Callback timeout / heartbeat you're adding here _don't_ have this issue.

I have a big refactor coming that re-arranges how the test Executor works and that makes it compatible with all the existing JS example tests (featuring concurrency). I've split it into 2 PRs for easier reviewing, but each is still pretty big 😓. 

I still have to rebase these coming PRs on the recent changes in this repo, which is not going to be a lot of fun. 

If you'd like, I can pick up this PR here and incorporate it after I got those 2 PRs in? It would make my re-basing life simpler, and I will happily pick up the great work you did here and do the merging legwork afterwards?

zhongkechen:
>It would make my re-basing life simpler

create a draft PR for your changes and I'll be happy to help you merge

zhongkechen:
Fixed the history event for wait operation in a new commit: https://github.com/aws/aws-durable-execution-sdk-python/pull/501/commits/6f1fb2d83355f637519b8a938282fc27bce10162

zhongkechen:
Tests failed due to this issue: https://github.com/aws/aws-durable-execution-sdk-python/issues/405

The fIx of the issue is in another PR: https://github.com/aws/aws-durable-execution-sdk-python/pull/505

yaythomas:
After the 1st of the two big coming PRs landed https://github.com/aws/aws-durable-execution-sdk-python/pull/515 ( #516 still pending review/merge) I started looking at rebasing this current PR to do the merging legwork as promised, but then...

It turns out that #515's timer restructuring, together with the per-execution ExecutionWorker in #516, make @SilanHe's suggestion to port the "skip time" behaviour from the JS/Java runners easier... 

I put an idea to skip time in this new dispensation up as #523 (stacked on the pending #516) so we can look at it concretely... I think it might be the case that we could go straight to skip_time rather than shipping fractional scaling as an interim step?

A difference in that PR I'd like to draw attention to, is that the current PR #501 also scales callback timeout/heartbeat timers, with a 5s floor, whereas #523 I provisionally skip wait and step-retry timers but leaves callback timeout/heartbeat timers on the real clock. I'm thinking those are externally driven, so skipping them instantly races the test's own send_callback? So the question is whether we fold callbacks into skip (with a floor), or keep them real, or handle them separately?

Keen to hear your thoughts @zhongkechen  and @SilanHe! Implementation in #523 matches what the JS testing framework does.

PR #502, opened 2026-06-30
feat(otel): emit logical step spans

zhongkechen:
## Summary
- create logical STEP operation spans from operation start/end callbacks
- keep per-attempt spans from user-function callbacks as children of the STEP span
- update log correlation and plugin tests for active attempt spans

## Tests
- hatch fmt
- hatch run dev-otel:test
- python3 .github/scripts/lintcommit.py --range origin/main..HEAD

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
@@ -152,12 +152,18 @@ class OtelPlugin(DurableInstrumentationPlugin):
         with self._operation_spans_lock:
             return self._operation_spans.get(operation_id)
 
+    @staticmethod
+    def _attempt_span_key(info: UserFunctionStartInfo | UserFunctionEndInfo) -> str:
+        """Return the registry key for a STEP attempt span."""
+        return f"{info.operation_id}:attempt:{info.attempt or 1}"
+
     def get_current_span_context(self) -> SpanContext | None:
         """Return the span context to use for log correlation.
 
         Resolution order:
-        1. The span attached to the OTel thread-local context. Inside a step or
-           child context this is the active operation span (attached in
+        1. The span attached to the OTel thread-local context. Inside a step
+           this is the active attempt span, and inside a child context this is
+           the active context span (attached in
            on_user_function_start), and between operations it is the enclosing
            operation span (restored in on_user_function_end).
         2. The invocation span from the plugin registry. This is the path used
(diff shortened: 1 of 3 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -411,18 +428,30 @@ def on_user_function_start(self, info: UserFunctionStartInfo) -> None:
...
         span = self._start_span(
             operation_id=info.operation_id,
             name=span_name,
             attributes=attributes,
             start_time=info.start_time,
             parent_span=parent_span,
-            existed=info.attempt != 1,
+            existed=info.attempt != 1 and info.operation_type is not OperationType.STEP,
+            span_key=span_key,
+            deterministic_span_id=info.operation_type is not OperationType.STEP,
```
SilanHe, line 456: hmm so the approach is to manage the operation span within on_operation_start and on_operation_end for STEP operation but for CONTEXT to continue creating it in the on_user_function_start, on_user_function_end.

SilanHe (dismissed):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
zhongkechen: yes, we will set the bottom level span as the current span so that the spans from user function will have the correct parent. So this has to be done in onUserFunctionStart

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
```diff
@@ -363,8 +379,8 @@ def on_operation_end(self, info: OperationEndInfo) -> None:
         operation span before being ended.
         """
         logger.debug("Durable operation ended: %s", info)
-        if info.operation_type in [OperationType.CONTEXT, OperationType.STEP]:
-            # Context and Step operations are tracked using on_user_function_end
+        if info.operation_type is OperationType.CONTEXT:
+            # Context operations are tracked using on_user_function_end.
```
SilanHe, line 383: is this comment always true? I think we recently merged a change where on_invocation_end does close spans for CONTEXT on invocation end.

SilanHe (approved):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
zhongkechen: I think it's still true unless this is removed from on_user_function_end:

https://github.com/aws/aws-durable-execution-sdk-python/blob/ddfcc9a10bb331e7114fccdd14b599789623bb07/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py#L399-L415

PR #504, opened 2026-07-02
feat: support updated operation IDs in replay

zhongkechen:
Summary:
- Add UpdatedOperationIds to DurableExecutionInvocationInput and pass it into ExecutionState.
- Emit plugin operation update/end hooks for operations changed while suspended instead of replay start/end hooks.
- Teach the local testing backend to carry updated operation IDs across resumed invocations.

Validation:
- hatch fmt --check
- hatch run dev-core:typecheck
- hatch run dev-testing:typecheck
- hatch run dev-core:test packages/aws-durable-execution-sdk-python/tests/execution_test.py packages/aws-durable-execution-sdk-python/tests/context_test.py
- hatch run dev-testing:test packages/aws-durable-execution-sdk-python-testing/tests/execution_test.py packages/aws-durable-execution-sdk-python-testing/tests/invoker_test.py
- hatch run dev-otel:test

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
@@ -509,12 +509,18 @@ class DurableContext(DurableContextProtocol):
             next_exists
             and next_checkpoint.operation.operation_type is OperationType.STEP
         )
-
         # While replaying, an operation that already has a checkpoint was
-        # observed in a prior invocation — notify plugins (once per op). State
+        # observed in a prior invocation. If the backend says the operation
+        # changed since the last invocation, notify plugins as an update rather
+        # than replayed history; otherwise notify as replayed history. State
         # owns the dedup; the context owns the "only while replaying" gate.
         if was_replaying and next_exists:
-            self.state.emit_operation_replay_hook(next_checkpoint.operation)
+            if self.state.is_operation_updated_since_last_invocation(
+                next_checkpoint.operation.operation_id
+            ):
+                self.state.emit_operation_update_hook(next_checkpoint.operation)
+            else:
+                self.state.emit_operation_replay_hook(next_checkpoint.operation)
         # Deferred flip applies only to non-step resume points. For step ops we
         # flip before instead, so don't defer.
         flip_after: bool = (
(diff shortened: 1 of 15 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -509,12 +509,17 @@ def _replay_aware(self):
...
+        # changed since the last invocation, notify plugins as an update rather
+        # than replayed history; otherwise notify as replayed history. State
         # owns the dedup; the context owns the "only while replaying" gate.
         if was_replaying and next_exists:
-            self.state.emit_operation_replay_hook(next_checkpoint.operation)
+            was_updated = self.state.emit_operation_update_hook(
+                next_checkpoint.operation
+            )
+            if not was_updated:
+                self.state.emit_operation_replay_hook(next_checkpoint.operation)
```
SilanHe: if on_operation_start and on_operation_end is emitted on every invocation, how does that affect the Otel plugin's spans?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -454,6 +464,32 @@ def emit_operation_replay_hook(self, operation: Operation) -> None:
...
+        """
+        if not self.is_operation_updated_since_last_invocation(operation.operation_id):
+            return False
+        if operation.operation_type is OperationType.EXECUTION:
+            return True
+
+        with self._updated_operation_hooks_lock:
+            if operation.operation_id in self._updated_operation_hooks:
+                return True
+            self._updated_operation_hooks.add(operation.operation_id)
```
SilanHe: why do we need to update this `_updated_operation_hooks`?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
zhongkechen: replay is true so existed is true. The spans on replay will be created with links to the spans in the previous invocation.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
zhongkechen: Yeah, deduplication logic here is redundant. Removed.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
SilanHe: I guess I'm asking whether all spans from previous replays will be recreated for each invocation or if they will only be present for invocations where they are relevant

wangyb-A (approved):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
zhongkechen: Only in invocations they are relevant

PR #506, opened 2026-07-03
feat: add operation change plugin hook

zhongkechen:
## Summary
- add OperationChangeInfo and on_operation_change to the Python plugin interface
- emit operation change hooks after checkpoint responses merge status changes
- enrich OperationInfo with result, error, attempt, and end_time for parity with the JS SDK
- add plugin executor and execution state coverage

## Verification
- hatch run dev-core:test packages/aws-durable-execution-sdk-python/tests/plugin_test.py packages/aws-durable-execution-sdk-python/tests/state_test.py -q
- hatch run dev-otel:test -q
- hatch run dev-core:typecheck
- hatch fmt --check
- hatch run test:all -q

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
@@ -5,9 +5,10 @@ import datetime
 import functools
 import logging
 from concurrent.futures import ThreadPoolExecutor
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
-from typing import Any, Callable, MutableMapping
+from collections.abc import Mapping, Sequence
+from typing import Any, Callable, MutableMapping, cast
 
 from aws_durable_execution_sdk_python.identifier import OperationIdentifier
 from aws_durable_execution_sdk_python.lambda_service import (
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -46,8 +101,15 @@ class OperationStartInfo(OperationInfo):
...
 @dataclass(frozen=True)
 class OperationEndInfo(OperationInfo):
-    end_time: datetime.datetime | None
-    error: ErrorObject | None
+    pass
+
+
+@dataclass(frozen=True)
+class OperationChangeInfo:
+    request_id: str | None
```
SilanHe: Sorry about this, I was just double checking on OperationChangeInfo in JS. Both the otel plugin and the insights plugin don't actually use this request_id field. I've decided to remove it since it's only there because OperationChangeInfo inherits from InvocationBaseInfo in JS.

[invocationInfo](https://github.com/aws/aws-durable-execution-sdk-js/blob/main/packages/aws-durable-execution-sdk-js/src/types/plugin.ts#L154)

This is kind of confusing, it's my fault for letting this change get merged on my behalf. I'm planning a few small interface changes which should clear up relationships between the interfaces/data classes without introducing breaking changes for any of the plugins.

In conclusion, my recommendation is to remove this field from OperationChangeInfo.

SilanHe: Here's my change for the JS sdk: https://github.com/aws/aws-durable-execution-sdk-js/pull/693

zhongkechen: Removed. Must have followed the old contract

SilanHe (approved):

PR #512, opened 2026-07-06
fix: retry transient GetDurableExecution not found

zhongkechen:
## Summary
- retry cloud runner status polling when GetDurableExecution returns ResourceNotFoundException
- keep non-retryable ClientError behavior unchanged
- add a regression test for transient ResourceNotFoundException followed by success

## Why this is needed
The cloud runner can invoke a durable Lambda and immediately begin polling GetDurableExecution. That API is eventually consistent: right after invoke, the durable execution ARN may be returned before the execution record is visible to GetDurableExecution. In that short window the service can return ResourceNotFoundException even though the execution is valid and will become readable shortly.

Without retrying this transient error, cloud tests can fail before the workflow has a chance to start. Retrying within the existing polling loop matches the runner's wait semantics and the existing wait_for_callback behavior for GetDurableExecutionHistory.

## Verification
- hatch run dev-testing:test packages/aws-durable-execution-sdk-python-testing/tests/runner_test.py::test_cloud_runner_wait_for_completion_get_execution_resource_not_found_retries packages/aws-durable-execution-sdk-python-testing/tests/runner_test.py::test_cloud_runner_wait_for_completion_get_execution_failure
- hatch fmt --check

### DIFF
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
@@ -1054,6 +1054,17 @@ class DurableFunctionCloudTestRunner:
                     DurableExecutionArn=execution_arn
                 )
                 execution = GetDurableExecutionResponse.from_dict(execution_dict)
+            except ClientError as e:
+                error_code = e.response.get("Error", {}).get("Code")
+                # Retry while the asynchronous invoke's durable execution record is
+                # still becoming visible to GetDurableExecution.
+                if error_code == "ResourceNotFoundException":
+                    logger.info("Execution status not available yet; retrying")
+                    time.sleep(self.poll_interval)
+                    continue
+
+                msg = f"Failed to get execution status: {e}"
+                raise DurableFunctionsTestError(msg) from e
             except Exception as e:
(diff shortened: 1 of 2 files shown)


zhongkechen:
The first attempt of the CI run failed: https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/28814987052/job/85452004870?pr=512

The actual failing assertion was:

```text
FAILED packages/aws-durable-execution-sdk-python-examples/test/parallel/test_parallel_with_named_branches.py::test_parallel_with_named_branches
AssertionError: assert 'fetch-orders' == 'fetch-user-data'
```

So this was not caused by the `GetDurableExecution` retry change. It looks like an existing ordering flake in the parallel example test.

That test expects `parallel_op.child_operations` to be in the same order as the branch declaration:

```python
assert child_names[0] == "fetch-user-data"
assert child_names[1] == "fetch-orders"
```

But the run observed `fetch-orders` first. Since the example uses `context.parallel(..., max_concurrency=3)`, child operations can be recorded in execution/checkpoint order rather than declaration order. On this Python 3.12 attempt, the first two parallel branches completed or were observed in the opposite order.

The important bit: the workflow result had already matched the expected result list, and only the child operation ordering assertion failed. So this is a test brittleness/concurrency ordering issue, unrelated to PR #512’s cloud runner retry behavior.

yaythomas (approved):

Issue #462, opened 2026-06-10
[Testing] Enhance get_execution_state with paging


dhegberg:
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/34

---

## Context
Implement proper paging logic in checkpoint processor

## Technical Requirements
- Implement proper paging logic in checkpoint processor
- Add next_marker handling for large operation lists
- Add pagination metadata to responses

## Implementation Details
- **File**: `src/aws_durable_execution_sdk_python_testing/checkpoint/processor.py:98`

## Acceptance Criteria
- [ ] Proper paging logic implemented in checkpoint processor
- [ ] Next_marker handling added for large operation lists
- [ ] Pagination metadata added to responses

Issue #460, opened 2026-06-10
[Testing] Complete get_durable_execution_history implementation


dhegberg:
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/36

---

## Context
Finish implementation in CLI module

## Technical Requirements
- Finish implementation in CLI module
- Convert operations to chronological event history

## Implementation Details
- **File**: `src/aws_durable_execution_sdk_python_testing/cli.py:378`

## Acceptance Criteria
- [ ] Implementation finished in CLI module
- [ ] Operations converted to chronological event history

Issue #439, opened 2026-06-10
[Testing] [Bug]: Wait timing optimization


dhegberg:
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/183

---

### Expected Behavior

```
const resultPromise = context.step(async () => {
    await new Promise((resolve) => setTimeout(resolve, 7000));
    return "step completed";
});
await context.wait("wait-1", { seconds: 10 });
const result = await resultPromise; // Re-invokes after 10s, rather than 3s
```

For this example real backend reinvokes after 3s.

### Actual Behavior

The function is re-invoked after the full wait duration of 10s instead of the remaining time (3s). 

### Steps to Reproduce

See above

### SDK Version

1.

### Python Version

3.13

### Is this a regression?

No

Issue #438, opened 2026-06-10
[Testing] [Bug]: Return updated operations on empty checkpoint


dhegberg:
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/184

---

### Expected Behavior

When creating a Checkpoint with an empty `updates` array, such that
```
 CheckpointDurableExecutionRequest(DurableExecutionArn=<<my-arn>>, CheckpointToken=<<my-token>>, Updates=[])

```

The response should contain `NewExecutionState`, where

### If NO new events occurred:
1. CheckpointToken: New token with incremented sequence
2. NewExecutionState.Operations: Empty array []
3. NewExecutionState.NextMarker: null

### If new events occurred (timer fired, etc.):
1. CheckpointToken: New token with incremented sequence
2. NewExecutionState.Operations: Array with those operations
3. NewExecutionState.NextMarker: null

The response CAN return an empty NewExecutionState.Operations array when there are no updates AND no new backend events.

The response should ALWAYS includes NewExecutionState - it never omits it.

### Actual Behavior

The local-runner has a short-circuit only to pass back the updated checkpoint token and set            new_execution_state=None if `updates` are empty. This is wrong.

```
        if updates:
            checkpoint_output = self._checkpoint_processor.process_checkpoint(
                checkpoint_token=checkpoint_token,
                updates=updates,
                client_token=client_token,
            )

            new_execution_state = None
            if checkpoint_output.new_execution_state:
                new_execution_state = CheckpointUpdatedExecutionState(
                    operations=checkpoint_output.new_execution_state.operations,
                    next_marker=checkpoint_output.new_execution_state.next_marker,
                )

            return CheckpointDurableExecutionResponse(
                checkpoint_token=checkpoint_output.checkpoint_token,
                new_execution_state=new_execution_state,
            )

        # Save execution state after generating new token
        new_checkpoint_token = execution.get_new_checkpoint_token()
        self._store.update(execution)

        return CheckpointDurableExecutionResponse(
            checkpoint_token=new_checkpoint_token,
            new_execution_state=None,
        )
```

### Steps to Reproduce

Create empty checkpoint call

### SDK Version

1

### Python Version

3.13

### Is this a regression?

No

PR #515, opened 2026-07-07
feat(testing): invoker state and checkpoint stamping

### LINKED ISSUE #464: [Testing] Add single instance execution control
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/28

---

## Context
Implement execution queue mechanism in executor

## Technical Requirements
- Implement execution queue mechanism in executor
- Add execution state tracking to prevent concurrent lambda invocations
- Implement queuing logic for multiple execution requests
- Add proper cleanup when executions complete

## Acceptance Criteria
- [ ] Execution queue mechanism implemented in executor
- [ ] Execution state tracking added to prevent concurrent lambda invocations
- [ ] Queuing logic implemented for multiple execution requests
- [ ] Proper cleanup added when executions complete

yaythomas:
## What

Align the local test runner's data-plane responses with the real service, validated against the JS example suite as a cross-language conformance check.

## Changes

**Invoker state and checkpoint stamping:**
- Track invocation state (PRE_INVOKE/INVOKING/COMPLETED) and gate checkpoints so at most one handler invocation is in flight per execution.
- Record per-update timestamps and stamp checkpoint operations with a monotonic sequence so delta computation and pagination share one snapshot.
- Offload handler invocation off the scheduler event loop so concurrent map/parallel branches no longer serialize on a single thread.
- Return a locked snapshot from in-process state reads.

**History generation (GetDurableExecutionHistory):**
- Emit one history event per checkpoint update so step retries report the correct StepStarted/StepFailed counts and attempt numbers.
- Reconstruct events from recorded update transitions using the real operation for details, with a fallback for operations completed outside the checkpoint path.

**Fidelity fixes:**
- Store the RETRY payload as the step result so waitForCondition state replays, and map a RETRY without error to StepSucceeded.
- Set callback timeout errorType to Callback.Timeout/Callback.Heartbeat with the correct message strings.
- Introduce `CheckpointCore.match_cached` as the single source of truth for idempotent checkpoint replay detection (both web and in-process paths delegate to it).

**Invocation timeout:**
- Add execution_timeout and invocation_timeout to the runner constructor and CLI. A timed-out invocation leaves the step STARTED and replays.

## Testing

- 1311 passed, 95.67% coverage (gate 95%)
- Typecheck clean (102 files)
- Format clean
- Cross-language conformance: JS example suite failures reduced from 27 to 7 (remaining are CHAINED_INVOKE feature gap and one upstream JS SDK bug)

## Stack

This is PR 1 of 2. PR 2 (per-execution worker serialization https://github.com/aws/aws-durable-execution-sdk-python/pull/516) builds on top.

Closes #464, #460, #462, #438, #439


packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/checkpoint/validators/valid_actions_by_operation_type.py
```diff
@@ -33,7 +33,10 @@
 
 
 class ValidActionsByOperationTypeValidator:
```
yaythomas: ooof, this belongs elsewhere.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/checkpoint/validators/transitions.py
```diff
@@ -1,66 +0,0 @@
-"""Validator for valid actions by operation type."""
-
-from __future__ import annotations
```
yaythomas, line 3: moved to CheckpointValidator

packages/aws-durable-execution-sdk-python-examples/test/conftest.py
```diff
@@ -107,14 +107,14 @@ def run(
...
-        return self._runner.run(input=input, timeout=timeout)
+        return self._runner.run(input=input)
 
     def run_async(
         self,
         input: str | None = None,  # noqa: A002
         timeout: int = 60,
     ) -> str:
-        return self._runner.run_async(input=input, timeout=timeout)
+        return self._runner.run_async(input=input)
```
zhongkechen: timeout parameter is still passed but not used any more. Not sure if this is intentional.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
zhongkechen: max_page_bytes parameter is not specified here.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/invoker.py
```diff
@@ -119,7 +119,8 @@ def create_invocation_input(
             checkpoint_token=execution.get_new_checkpoint_token(),
             initial_execution_state=InitialExecutionState(
                 operations=execution.operations,
-                next_marker="",
+                # the relevant invariant: null when the page fits in one shot, not "".
+                next_marker=None,  # type: ignore[arg-type]
```
zhongkechen: Why is this always None?

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
```diff
@@ -779,78 +1318,220 @@ def _validate_invocation_response_and_store(
...
+                        str(uuid.uuid4()),
+                    )
+                    self._store.save(execution)
+                    self._invocation_state[execution_arn] = InvocationState.PRE_INVOKE
                 error_obj = ErrorObject.from_message(
-                    message=f"Function not found: {execution.start_input.function_name}"
+                    message=f"Function timed out after {self._invocation_timeout_seconds} seconds"
                 )
-                self._fail_workflow(execution_arn, error_obj)
+                self._retry_invocation(execution, error_obj)
```
zhongkechen, line 1522: I'm afraid we haven't killed the current invocation when we start a new one. What happens if the current invocation is still running and still doing some actual work like creating durable operations?

packages/aws-durable-execution-sdk-python-examples/test/conftest.py
yaythomas: nice catch! 

added `execution_timeout` to run() for symmetry with `run_async()`.

the conftest adapter now maps `timeout` to `execution_timeout`.

I'm keeping the old timeout there but with the deprecation notice for the moment, plan is to remove on next major version bump.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
yaythomas: added the arg into CheckpointProcessor + both invokers + Executor, collapsed the duplicated 5_000_000 constant into one canonical 5 * 1024 * 1024 size limit.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/invoker.py
yaythomas: because I failed to wire up the pager for initial invokes 😬 

added: the invoke input now pages via OperationPaginatorState.page() with next_marker = marker or ""

Added 2 round-trip tests.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
yaythomas: wow, sharp eyes 🦅 , and a tricky, subtle one to notice too! thank you! 🙇

We don't kill the in-flight invocation (we can't force-stop it in-process the way the platform stops a timed-out sandbox).

But what we can do and what I missed in this first revision, is make a superseded invocation's subsequent/orphaned checkpoint calls harmless.

 I added a per-invocation identity (invocation_id) to `CheckpointToken`, minted via `Execution.begin_new_invocation()` when a new invocation is dispatched.

Once that happens, a checkpoint carrying the old invocation's token is rejected with InvalidParameterValueException. So now orphaned work will just be abandoned and run to completion in the background.

Added a regression test covering the superseded-checkpoint rejection.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
```diff
@@ -350,55 +436,40 @@ def get_execution_state(
...
+                self._invocation_state.get(execution_arn, InvocationState.PRE_INVOKE)
+                is not InvocationState.INVOKING
+            ):
+                msg = "Invalid checkpoint token"
+                raise InvalidParameterValueException(msg)
 
-        # Get operations (excluding the initial EXECUTION operation for state)
-        operations = execution.get_assertable_operations()
+            # the relevant invariant token check.
+            if not self._is_current_token(execution, checkpoint_token):
```
zhongkechen, line 465: Would this cause web handler to fail?

https://github.com/aws/aws-durable-execution-sdk-python/blob/99b482e1dcbed69a15b71c71c5beb6012ad72e2e/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/web/handlers.py#L456

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
```diff
@@ -556,56 +791,355 @@ def checkpoint_execution(
...
+            new_token_sequence = execution.advance_token_sequence()
 
-        if updates:
-            checkpoint_output = self._checkpoint_processor.process_checkpoint(
-                checkpoint_token=checkpoint_token,
-                updates=updates,
-                client_token=client_token,
+            paginator = OperationPaginatorState.pin(execution)
+            delta = paginator.unseen_operations()
+            response_ops, _ = paginator.page_subset(
```
zhongkechen: What would happen to the next page?

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
yaythomas: ah, missed the emulator. adding.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
yaythomas: the remaining ops aren't lost: advance_handler_seen only moves the watermark to cover the ops actually returned, so the rest come back in the next checkpoint's delta. covered by `test_delta_truncation_advances_watermark_partially`.

yaythomas: actually hang on, I'm going to remove paging here.

zhongkechen (approved):

PR #517, opened 2026-07-07
ci: deploy examples with SAM stack

### LINKED ISSUE #518: Migrate Python SDK CI tests with SAM CLI
- Migrate the test deployment mechanism from direct AWS CLI/SDK Lambda calls to SAM CLI (sam build + sam deploy).\
- Validate that all integration examples deploy and pass, cleanup stale functions and update document if required.

zhongkechen:
## Summary
- Replace the old per-function example deployment experience with a SAM-stack based flow: developers now build once, generate a template, and deploy the full examples stack.
- Remove the single-function deploy CLI path so manual and CI workflows match how we actually test examples in cloud.
- Remove the checked-in generated SAM template and update workflow; developers no longer need to maintain or review generated template churn.
- Keep local example tests simple with `hatch run dev-examples:test`, while cloud tests use the deployed stack and an automatically generated function-name map.
- Make each cloud-test matrix leg deploy Lambda functions with the matching Python runtime, so the `3.11`, `3.12`, `3.13`, and `3.14` jobs actually exercise cloud resources on that runtime.
- Add GitHub Actions job summaries for cloud test results so each matrix leg reports test totals and individual test-case outcomes without requiring log spelunking.
- Add optional Lambda endpoint override support for preprod testing without hard-coding an endpoint into the workflow.

## Workflow changes
- Add `.github/workflows/cloud-tests.yml` with workflow display name `Cloud tests`.
- Run the workflow for every PR targeting `main` and via `workflow_dispatch`.
- Rename the cloud test job to `example-tests`.
- Run cloud tests across Python `3.11`, `3.12`, `3.13`, and `3.14`.
- Use Python-version prefixes for deployed resources: `Py311`, `Py312`, `Py313`, and `Py314`.
- Prefix stack names as `${PYTHON_PREFIX}-python-examples` and Lambda function names as `${PYTHON_PREFIX}-`.
- Add a concurrency group per Python version, `cloud-tests-${{ matrix.python-prefix }}`, with `cancel-in-progress: false`, so workflows for the same Python version do not overlap.
- Configure AWS credentials with `secrets.TEST_ROLE_ARN`.
- Verify the assumed account matches `secrets.TEST_ACCOUNT_ID`.
- Deploy all examples with SAM using `sam build` and `sam deploy`.
- Pass `PythonRuntime=python${{ matrix.python-version }}` to SAM so each matrix leg deploys functions on the same Python version used by the test runner.
- Use `secrets.TEST_LAMBDA_EXECUTION_ROLE_ARN` as the pre-existing Lambda execution role for every generated function.
- Add optional Lambda endpoint override support through the manual `lambda_endpoint` input or the `TEST_LAMBDA_ENDPOINT` repository/org variable.
- Pass the optional Lambda endpoint override to both SAM (`LambdaEndpoint`) and pytest (`LAMBDA_ENDPOINT`) so deployed functions and test clients target the same endpoint.
- Only export `LAMBDA_ENDPOINT` to later workflow steps when an override is configured, so normal PR runs do not pass an empty endpoint to boto3.
- Generate `PYTEST_FUNCTION_NAME_MAP` after deployment and pass it to the test runner.
- Run the integration tests concurrently with `pytest-xdist` using `-n 8 --dist loadfile`.
- Write pytest JUnit XML to `.test-results/examples.xml`.
- Add an `always()` summary step that reports total, passed, failed, errored, and skipped test counts to `$GITHUB_STEP_SUMMARY`.
- Include failed, errored, and skipped test cases in the visible summary and all test cases in a collapsible details section.
- Report a clear summary message when pytest fails before writing the JUnit report.
- Leave deployed stacks in place after cloud test runs so failed runs can be inspected and debugged.

## Removed workflows and legacy deployment path
- Delete `.github/workflows/deploy-examples.yml`.
- Delete `.github/workflows/integration-tests.yml`.
- Delete `.github/workflows/update-sam-template.yml`.
- Delete the checked-in `packages/aws-durable-execution-sdk-python-examples/template.yaml`.
- Delete the old `packages/aws-durable-execution-sdk-python-examples/cli.py`.
- Remove Hatch scripts for the deleted example CLI commands: `cli`, `bootstrap`, `deploy`, `invoke`, `get`, `history`, `policy`, and `list`.

## SAM/template generation
- Update `generate_sam_template.py` to build `template.generated.json` from `examples-catalog.json`.
- Add reusable helpers for loading the catalog, converting handler names to CloudFormation logical IDs, and building the template object.
- Add a `--output` option to write the generated template to a custom path.
- Add `FunctionNamePrefix` and `LambdaExecutionRoleArn` template parameters.
- Add a `PythonRuntime` template parameter with default `python3.13` and allowed values for `python3.11`, `python3.12`, `python3.13`, and `python3.14`.
- Set `Globals.Function.Runtime` from the `PythonRuntime` parameter instead of hard-coding `python3.13`.
- Set each generated function's `FunctionName` from `FunctionNamePrefix` plus the logical ID.
- Set each generated function's `Role` from the `LambdaExecutionRoleArn` parameter.
- Preserve per-example `DurableConfig`, `LoggingConfig`, `Layers`, `Tracing`, and environment settings from the catalog.
- Keep `LambdaEndpoint` as a SAM parameter with the standard Lambda endpoint default and wire it to `AWS_ENDPOINT_URL_LAMBDA` for generated functions.
- Stop creating an IAM role in the generated template.

## Build/test tooling
- Add `scripts/build_examples.py` to build a shared example bundle for SAM packaging.
- The build script cleans the examples `build/` directory, installs runtime packages into it, and copies example source files into the Lambda bundle.
- Add `scripts/generate_function_name_map.py` to create the JSON mapping from example names to qualified `$LATEST` function names.
- Add `scripts/write_junit_summary.py` to convert pytest JUnit XML into Markdown for GitHub Actions summaries.
- Update `hatch run examples:build` to use `build_examples.py`.
- Update `hatch run examples:generate-sam-template` to generate `template.generated.json`.
- Update `hatch run examples:clean` to remove the examples build directory, generated template, and root `.aws-sam` build output.
- Add `pytest-xdist` to the shared test Hatch environment for concurrent cloud test execution.

## Cloud test runner changes
- Replace the single-function cloud-test environment variables with `PYTEST_FUNCTION_NAME_MAP`.
- Update cloud test lookup to parse the JSON function map and match test marker names case-insensitively.
- Skip cloud tests whose example name is not present in `PYTEST_FUNCTION_NAME_MAP`.
- Keep support for optional `LAMBDA_ENDPOINT` in the cloud runner.
- Treat a blank `LAMBDA_ENDPOINT` value as unset to avoid boto3 `Invalid endpoint` errors.

## Known follow-up
- The retained-stack cloud retry example flake is intentionally left for a future PR; this PR does not change `StepContext` or the retry examples to address it.

## Documentation and ignore updates
- Update `CONTRIBUTING.md` to document the SAM build/deploy flow instead of the old example CLI deploy flow.
- Update the examples test README to describe deploying the full SAM stack before cloud tests.
- Document the optional `PythonRuntime` SAM parameter in manual deploy examples.
- Document SAM remote execution as the replacement path for invoking deployed examples, inspecting durable executions, stopping executions that go wrong, and managing reusable remote test events after the examples stack is deployed.
- Link to the SAM remote execution command reference and show `sam remote invoke`, `sam remote execution get`, `sam remote execution history`, `sam remote execution stop`, `sam remote test-event put`, and `sam remote test-event list`.
- Document `PYTEST_FUNCTION_NAME_MAP` as the required cloud-mode test mapping.
- Point CI documentation to `cloud-tests.yml`.
- Ignore root `.aws-sam/` output and generated `template.generated.json` files.

## Testing
Latest follow-up validation after reverting the flaky retry test fix:
- `hatch fmt --check`
- `PYTHONPYCACHEPREFIX=/private/tmp/codex-pycache python3 -m py_compile packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/types.py packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py packages/aws-durable-execution-sdk-python-examples/src/step/step_with_retry.py packages/aws-durable-execution-sdk-python-examples/src/step/steps_with_retry.py`
- `git diff --check --cached`
- `hatch run dev-examples:pytest packages/aws-durable-execution-sdk-python-examples/test/step/test_step_with_retry.py packages/aws-durable-execution-sdk-python-examples/test/step/test_steps_with_retry.py`

- `hatch fmt --check`
- `python3 -m py_compile packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py packages/aws-durable-execution-sdk-python-examples/scripts/generate_function_name_map.py packages/aws-durable-execution-sdk-python-examples/scripts/build_examples.py packages/aws-durable-execution-sdk-python-examples/scripts/write_junit_summary.py packages/aws-durable-execution-sdk-python-examples/test/conftest.py`
- `hatch run examples:generate-sam-template`
- Generated-template assertion: 61 functions use `LambdaExecutionRoleArn` and no `DurableFunctionRole` resource is present.
- Generated-template assertion: `Globals.Function.Runtime` resolves from `PythonRuntime`, whose default remains `python3.13`.
- `python3 packages/aws-durable-execution-sdk-python-examples/scripts/generate_function_name_map.py --function-name-prefix PythonPr517-`
- `sam validate --template-file packages/aws-durable-execution-sdk-python-examples/template.generated.json --region us-west-2`
- `sam build --template-file packages/aws-durable-execution-sdk-python-examples/template.generated.json`
- `hatch run test:examples-integration --help`
- `hatch run dev-examples:test -k test_hello_world --junitxml=/private/tmp/codex-examples-junit.xml`
- `hatch run dev-examples:test`
- `hatch run dev-core:test`
- `python3 packages/aws-durable-execution-sdk-python-examples/scripts/write_junit_summary.py /private/tmp/codex-examples-junit.xml --title "Cloud test summary (Python 3.14)" --output /private/tmp/codex-examples-summary.md`
- `python3 packages/aws-durable-execution-sdk-python-examples/scripts/write_junit_summary.py /private/tmp/missing-junit.xml --title "Cloud test summary (Python 3.14)" --output /private/tmp/codex-missing-summary.md`
- YAML parse check for `.github/workflows/cloud-tests.yml`
- `git diff --check`

Previously also ran:
- `hatch run dev-examples:test -k test_hello_world`
- `hatch run examples:build`

Did not run an actual AWS deploy locally.


.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
+name: Cloud tests
+
+on:
+  pull_request:
```
yaythomas, line 4: no `paths:...`

The old deploy-examples.yml had a paths filter so it only ran when
- packages/aws-durable-execution-sdk-python/src/**
- packages/aws-durable-execution-sdk-python-examples/**,

or the workflow itself changed.

And the old integration-tests.yml was Python 3.13 only.

This new workflow runs on every PR to main regardless of path, on all four Python versions. Not the end of the world, but it could result in unnecessary deployments/actions running.

.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+      env:
+        LAMBDA_EXECUTION_ROLE_ARN: ${{ secrets.TEST_LAMBDA_EXECUTION_ROLE_ARN }}
+        LAMBDA_ENDPOINT_OVERRIDE: ${{ github.event.inputs.lambda_endpoint || vars.TEST_LAMBDA_ENDPOINT }}
+        PYTHON_PREFIX: ${{ matrix.python-prefix }}
+        PYTHON_RUNTIME: python${{ matrix.python-version }}
+      run: |
+        STACK_NAME="${PYTHON_PREFIX}-python-examples"
+        FUNCTION_NAME_PREFIX="${PYTHON_PREFIX}-"
+
+        STACK_NAME=$(echo "$STACK_NAME" | tr -cd '[:alnum:]-' | cut -c1-128)
```
yaythomas: some of dynamic scripting here we could possibly do in env via format? 

The stack name is only keyed on Python version, so all PRs share the same four stacks.

group + `cancel-in-progress: false` will make it queue rather than clobber existing for simultaneous PRs.

but this means the next PR for that Python version will overwrite everything.

If we want per-PR isolation, then "PR-<number>-Python or something"), the small change is to compose the names in env: instead of the run: shell step,
using the pull_request context.

```
    jobs:
      example-tests:
        env:
          # PR runs get PrN; workflow_dispatch and pushes get RunN.
          PR_TAG: ${{ github.event.pull_request.number && format('Pr{0}', github.event.pull_request.number) || format('Run{0}', github.run_id) }}
        steps:
          ...
          - name: Deploy all examples with SAM
            env:
              STACK_NAME: ${{ matrix.python-prefix }}-${{ env.PR_TAG }}-python-examples
              FUNCTION_NAME_PREFIX: ${{ matrix.python-prefix }}-${{ env.PR_TAG }}-
              ...
```


but this also has the negative that per-PR stacks accumulate until they get cleaned up.

We could have a a companion `on: pull_request types: [closed]` workflow that runs aws cloudformation delete-stack across the four python-prefix values?

.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+  id-token: write
+  contents: read
+
+jobs:
+  example-tests:
+    name: example-tests (${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    concurrency:
+      group: cloud-tests-${{ matrix.python-prefix }}
+      cancel-in-progress: false
```
yaythomas, line 31: how about per-PR rather than per-runtime?

```
concurrency:
  group: cloud-tests-${{ matrix.python-prefix }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

this also cancels predecessors for the same PR.

packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py
```diff
@@ -29,91 +35,95 @@ def generate_sam_template():
...
-                                        "Resource": "*",
-                                    }
-                                ],
-                            },
-                        }
-                    ],
-                },
-            }
+            },
+            "FunctionNamePrefix": {
```
yaythomas, line 56: nit:

maybe make this one required rather than empty, or set a distinctive default like "PythonExamples-"??

if someone (definitely future me!) runs sam deploy manually without setting FunctionNamePrefix, I'd get functions named HelloWorld, Timer, Step, etc. 

short, generic, easy to collide with other things in an accountThe CI workflow always sets the prefix, so
CI is safe; the risk is a human running the CONTRIBUTING.md recipe.

packages/aws-durable-execution-sdk-python-examples/scripts/build_examples.py
```diff
@@ -0,0 +1,68 @@
...
+            "--upgrade",
+            "--target",
+            str(build_dir),
+            *[str(package) for package in runtime_packages],
+        ],
+        check=True,
+    )
+
+    logger.info("Copying examples from %s", src_dir)
+    for file_path in src_dir.rglob("*"):
```
yaythomas, line 48: The flattening step copies every file under src/ to build/<basename> and the comment already explains why. But if basenames collide, then ordering is rglob("*") order. When two files collide, whichever comes later wins. Not deterministic across filesystems.

e.g src/foo/step.py alongside src/step/step.py.

could raise on collision?

```
    if not file_path.is_file():
        continue
    dst = build_dir / file_path.name
    if dst.exists() and not dst.samefile(file_path):
        raise RuntimeError(
            f"Duplicate basename would flatten to {dst}: {file_path}"
        )
    shutil.copy2(file_path, dst)
```

.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+        PYTEST_FUNCTION_NAME_MAP=$(
+          python packages/aws-durable-execution-sdk-python-examples/scripts/generate_function_name_map.py \
+            --function-name-prefix "$FUNCTION_NAME_PREFIX"
+        )
+
+        echo "PYTEST_FUNCTION_NAME_MAP=$PYTEST_FUNCTION_NAME_MAP" >> $GITHUB_ENV
+
+    - name: Run Integration Tests
+      env:
+        AWS_REGION: ${{ env.AWS_REGION }}
```
yaythomas: nit: already set at top-level

CONTRIBUTING.md
```diff
@@ -94,7 +94,18 @@ hatch build
 
 # Examples deployment (from repo root)
 hatch run examples:build
-hatch run examples:deploy "Hello World"
+hatch run examples:generate-sam-template
```
yaythomas, line 97: thank you so much for this! awesome to have the CONTRIBUTING guide and the test/README get an update alongside the change :-)

yaythomas (commented):
This is a massive improvement on the ad-hoc per-function CLI deploy path to a SAM stack, and it deletes a lot of code. Thank you! 

The refactor of generate_sam_template.py is really cool too!

which is to say review comments are largely nits and I'm pretty much happy to approve :-) 

the broader conceptual question I have is whether we want per-PR, or per-runtime-version stacks. If we want per-runtime, which is what is currently implement, shared per-version stack then only the most recent run is left behind for debugging. (so 1 PR could overwrite another PRs). but if we go per-PR, then we have to think about cleaning up stacks.

I'm not all that opinionated about this either, just flagging this for discussion and happy to go with your preference :-)

packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py
```diff
@@ -29,91 +35,95 @@ def generate_sam_template():
...
+                "Type": "String",
+                "Default": "",
+            },
+            "LambdaExecutionRoleArn": {
```
bchampp, line 60: If we don't specify this, can we fallback on attaching the managed `AWSLambdaBasicDurableExecutionRolePolicy`?

pyproject.toml
```diff
@@ -153,17 +154,9 @@ workspace.members = [
...
 
 [tool.hatch.envs.examples.scripts]
-cli = "python packages/aws-durable-execution-sdk-python-examples/cli.py {args}"
-bootstrap = "python packages/aws-durable-execution-sdk-python-examples/cli.py bootstrap"
 generate-sam-template = "python packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py"
-build = "python packages/aws-durable-execution-sdk-python-examples/cli.py build"
-deploy = "python packages/aws-durable-execution-sdk-python-examples/cli.py deploy {args}"
-invoke = "python packages/aws-durable-execution-sdk-python-examples/cli.py invoke {args}"
-get = "python packages/aws-durable-execution-sdk-python-examples/cli.py get {args}"
-history = "python packages/aws-durable-execution-sdk-python-examples/cli.py history {args}"
```
bchampp, line 163: We have these counterparts in SAM, e.g. `sam remote execution history {arn}`. Consider updating these to point to those commands instead (or at least adding them in the contributing guide)

.github/workflows/cloud-tests.yml
bchampp: +1 - we don't need to run integration tests when the testing SDK changes.

.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
+name: Cloud tests
+
+on:
+  pull_request:
+    branches: ["main"]
+  workflow_dispatch:
+    inputs:
+      lambda_endpoint:
```
bchampp, line 8: Should region be an input here too? Would specifying an endpoint for a different region than the `AWS_REGION` specifies break somewhere?

.github/workflows/cloud-tests.yml
bchampp: +1 - we should be disambiguating all function names (and stacks) with the PR number. That's one of the tenant boundaries we had in the previous approach.

.github/workflows/cloud-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+
+    - name: Run Integration Tests
+      env:
+        AWS_REGION: ${{ env.AWS_REGION }}
+      run: |
+        echo "Running integration tests for all deployed examples"
+        echo "AWS Region: ${AWS_REGION}"
+        echo "Configured function map: ${PYTEST_FUNCTION_NAME_MAP}"
+
+        hatch run test:examples-integration -n 8 --dist loadfile
```
bchampp: At one point we had set this up so that the GitHub runner orchestrated multiple child workflows, one per example, so that it was easy to see specifically which test is failing after a given change. From a developer experience I found this to be nicer because I could quickly see (without clicking into the workflow) what's breaking. What do you think about updating the test runner to fan out like that? (I think it could be noisy with running 1 example per runtime though)

bchampp: Alternatively, if the workflow reported the test summary as a comment in the PR after it ran (and highlights any failing tests), this could work too.

.github/workflows/cloud-tests.yml
bchampp: The per-PR stacks shouldn't be accumulating. We need a clean up job that runs after a given PR is merged. I thought we had something like that already but I think it may have just been for cleaning up the log groups from the function executions. Since we're deploying CloudFormation stacks we'd need to add a new workflow to clean those up too (after a given PR is merged).

.github/workflows/cloud-tests.yml
zhongkechen: I intentionally remove the paths because I found it was not running initially when I updated the workflows and some other files. 

> run integration tests when the testing SDK changes.

I think we need to rerun tests when testing package changes. Integration tests rely on testing package.

.github/workflows/cloud-tests.yml
zhongkechen: Good catch. I'll add region as an input.

.github/workflows/cloud-tests.yml
zhongkechen: > so that it was easy to see specifically which test is failing after a given change

Agreed if the number of test cases is relatively small. When the number grows, I'm afraid it wouldn't be a big difference looking at the long list of jobs vs looking at the logs directly.

packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py
zhongkechen: The new role that github assumes doesn't have the permission to create new roles. We use a predefined role everywhere.

pyproject.toml
zhongkechen: I'll put the commands from the doc below in the contributing guide
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-remote-execution.html

.github/workflows/cloud-tests.yml
zhongkechen: Fact: now we have 244 functions per workflow run

packages/aws-durable-execution-sdk-python-examples/scripts/build_examples.py
zhongkechen: This is how it's working now. We can follow up separately if it's a concern.

https://github.com/aws/aws-durable-execution-sdk-python/blob/11a45212911b209d44fce8b2137265a5b259ab1e/packages/aws-durable-execution-sdk-python-examples/cli.py#L76-L79

.github/workflows/cloud-tests.yml
zhongkechen: I intentionally set `cancel-in-progress: false` so that a new deployment wouldn't be triggered when an old one is still ongoing.

.github/workflows/cloud-tests.yml
zhongkechen: fixed.

.github/workflows/cloud-tests.yml
zhongkechen: moved shell variables to env

zhongkechen: As discussed offline, we'll leave per-PR stack as a future improvement if we find it useful.

.github/workflows/cloud-tests.yml
zhongkechen: added a summary of test cases, see example here: https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/28965835771

packages/aws-durable-execution-sdk-python-examples/scripts/generate_sam_template.py
zhongkechen: Done.

zhongkechen:
We decided to leave this open for future considerations.

**Per runtime stacks**

Pros:
- Keeps the AWS footprint bounded: one examples stack per Python runtime instead of one set per active PR/runtime.
- Has simpler lifecycle management, since there are only stable runtime stacks to update rather than many short-lived PR stacks to garbage collect.
- Updating existing Lambda functions is usually faster than creating new functions from scratch, so repeated runs should generally spend less time in deploy.
- Reduces quota pressure around Lambda functions, CloudFormation stacks, SAM packaging artifacts, and related account limits.
- The current concurrency group per runtime prevents two jobs from mutating the same runtime stack at the same time.

Cons:
- A newer PR run can overwrite the deployed resources from an older PR run for the same runtime.
- Only the most recent run per runtime is left behind, so debugging a previous failed PR run is harder.
- Logs, durable executions, and function state all share stable function names, which can make forensic debugging noisier.
- Per-runtime concurrency can become a bottleneck during busy periods because all PRs serialize on the same `Py311`/`Py312`/`Py313`/`Py314` stack.

**Per PR stacks**

Pros:
- Better isolation: each PR gets its own deployed resources, so runs do not clobber each other.
- Failed runs are easier to inspect because another PR will not replace the stack before someone looks at it.
- Different PRs can run cloud tests in parallel without competing for the same runtime stack.
- AWS artifact names can directly identify the PR and runtime, e.g. `Pr517-Py313-*`.

Cons:
- Resource usage multiplies quickly because this workflow deploys the full examples stack across four runtimes.
- Needs explicit cleanup, ideally on `pull_request.closed` plus a scheduled TTL cleanup for abandoned or failed workflows.
- Higher risk of hitting account quotas during busy periods.
- Same-PR reruns still need serialization or deterministic replacement behavior, otherwise multiple pushes to the same PR can leave overlapping stacks.

yaythomas (approved):
:shipit:

PR #523, opened 2026-07-09
feat(testing): add skip-time virtual clock

### LINKED ISSUE #437: [Testing] [Feature]: Add option to skip time in test runner
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/188

---

### What would you like?

We're missing an option in the unit testing runner to pass a flag to allow developers to override wait states in their unit tests. We *do* support this with an environment variable: `DURABLE_EXECUTION_TIME_SCALE` but it should be a parameter to the test runner too.

### Possible Implementation

We should add a `skip_time` flag to the test runners `.run()` method, or have it be enabled globally on the test runner object.

No

### Does this require an RFC?

No

### Additional Context

_No response_

yaythomas:
*Description of changes:*

- Add Clock protocol with RealClock and SkipClock implementations
- Give each ExecutionWorker its own clock via the registry factory
- Drive wait and step-retry timers off the clock so modeled delays complete instantly under skip while history keeps real durations
- Default DurableFunctionTestRunner to skip_time=True; pass skip_time=False for real wall-clock timing
- Remove the DURABLE_EXECUTION_TIME_SCALE env-var scaling

*Issue #, if available:*
closes #437 

### DIFF
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/checkpoint/processors/wait.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/checkpoint/processors/wait.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-import logging
-import os
-from datetime import UTC, datetime, timedelta
+from datetime import datetime, timedelta
 from typing import TYPE_CHECKING
 
 from aws_durable_execution_sdk_python.lambda_service import (
@@ -36,6 +34,7 @@ class WaitProcessor(OperationProcessor):
         current_op: Operation | None,
         notifier: ExecutionNotifier,
         execution_arn: str,
+        now: datetime,
     ) -> Operation:
         """Process WAIT operation update with scheduler integration for timers."""
         match update.action:
@@ -43,13 +42,8 @@ class WaitProcessor(OperationProcessor):
                 wait_seconds = (
                     update.wait_options.wait_seconds if update.wait_options else 0
                 )
-                time_scale = float(os.getenv("DURABLE_EXECUTION_TIME_SCALE", "1.0"))
-                logging.info("Using DURABLE_EXECUTION_TIME_SCALE: %f", time_scale)
-                scaled_wait_seconds = wait_seconds * time_scale
 
-                scheduled_end_timestamp = datetime.now(UTC) + timedelta(
-                    seconds=scaled_wait_seconds
-                )
+                scheduled_end_timestamp = now + timedelta(seconds=wait_seconds)
 
                 # Create WaitDetails with scheduled timestamp
                 wait_details = WaitDetails(
@@ -63,7 +57,7 @@ class WaitProcessor(OperationProcessor):
                     status=OperationStatus.STARTED,
                     parent_id=update.parent_id,
                     name=update.name,
-                    start_timestamp=datetime.now(UTC),
+                    start_timestamp=now,
                     end_timestamp=None,
                     sub_type=update.sub_type,
                     execution_details=None,
@@ -87,6 +81,7 @@ class WaitProcessor(OperationProcessor):
                     update=update,
                     current_operation=current_op,
                     status=OperationStatus.CANCELLED,
+                    now=now,
                 )
             case _:
                 msg: str = "Invalid action for WAIT operation."
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
@@ -36,6 +36,7 @@ from aws_durable_execution_sdk_python_testing.checkpoint.processor import (
     DEFAULT_MAX_INVOCATION_PAGE_BYTES,
 )
 from aws_durable_execution_sdk_python_testing.client import InMemoryServiceClient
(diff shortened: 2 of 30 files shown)


packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/checkpoint/processors/wait.py
```diff
@@ -43,12 +41,9 @@ def process(
...
                     update.wait_options.wait_seconds if update.wait_options else 0
                 )
-                time_scale = float(os.getenv("DURABLE_EXECUTION_TIME_SCALE", "1.0"))
-                logging.info("Using DURABLE_EXECUTION_TIME_SCALE: %f", time_scale)
-                scaled_wait_seconds = wait_seconds * time_scale
 
                 scheduled_end_timestamp = datetime.now(UTC) + timedelta(
-                    seconds=scaled_wait_seconds
+                    seconds=wait_seconds
                 )
```
zhongkechen: It is still using the real time here, which causes the operation after the wait operation to have a start time stamp earlier than the scheduled end time of wait. This confuses the timeline of the history. Not sure if this is the desired behavior.

yaythomas: not desired :-) 

Fixed: made clock.now() the single source of "now" for stamping too. Threaded now from the execution's clock through CheckpointCore to each processor, and wired a clock resolver into both runner constructors. 

history is now monotonic (start + accumulated modeled durations), matching the JS @sinonjs/fake-timers behavior.

Added a stepA → wait(60) → stepB test asserting monotonicity and faithful duration

"RealClock.now() == datetime.now(UTC), so real-time runs are byte-identical; only skip mode changes.

zhongkechen (dismissed):
LGTM

yaythomas:
rebased on main, no other change

[edit:] ooh hang on CI failures 😕 , investigating

zhongkechen (dismissed):

zhongkechen:
seems `wait_for_callback(...)` failed before the test ever sends a callback.

yaythomas:
changed: 

1. Advancing SkipClock: now() = real time + accumulated skip, arm() adds the gap — monotonic, distinct timestamps, matching fake-timers shouldAdvanceTime: true.
2. Per-runner clock: registry and worker are clock-free again; one Clock is shared by the Executor, CheckpointProcessor, and both runner types.
3. All history stamp sites moved to the clock: execution start/end, wait and callback completions, invocation start/end, the history fallback, and the stop response now reuses the stamped time. events_to_operations uses the event's own timestamp instead of wall-clock. Test-only fallbacks go through real_now().
4. TID251 ruff ban on datetime.now outside clock.py — verified it fires on a synthetic violation.

Verification: 1353 testing-package tests pass, full workspace run 2883 passed, the 4 previously failing callback tests pass locally and in CI, mypy clean (119 files), hatch fmt --check clean. 

3.13 integration test failure failing only on test_otel_logger_example_spans_in_xray, which is known pre-existing OTel X-Ray flk.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
zhongkechen: DurableChildContextTestRunner doesn't support the new `skip_time` parameter

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
```diff
@@ -619,11 +627,13 @@ def __init__(
             store if store is not None else InMemoryExecutionStore()
         )
         self.poll_interval = poll_interval
+        self._clock: Clock = SkipClock() if skip_time else RealClock()
```
zhongkechen, line 630: Would a shared SkipClock() between multiple executions cause issues?

zhongkechen (dismissed):

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
yaythomas: adding it in, ty!

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
yaythomas: The only cross-execution effect is acceleration.

If execution A arms a skip past execution B's wait horizon, B's wait becomes due on its next wake instead of at its own arm. Since a single monotonic clock serves everything, it can only accelerate, never reorder.

What it can't do:

1. Prematurely fire externally driven timers. Callback timeout/heartbeat, execution timeout, and invocation timeout are real call_later delays on the scheduler, so A skipping an hour of virtual time doesn't age B's 30s callback timeout.
2. Distort history. Modeled durations come from scheduled_end - start, both stamped from the same clock, so they stay faithful regardless of who skipped what.
3. The alternative (per-execution clocks) is worse: executions share one store and one history surface, so independently advancing clocks put cross-execution timestamps in different clock domains — the same class of bug as the CI failure this revision fixed, one level up.

zhongkechen (approved):

PR #533, opened 2026-07-14
feat: limit in-flight operations in map/parallel

### LINKED ISSUE #279: [Feature]: `max_concurrency` in map/parallel doesn't throttle async operations like `invoke`
### What would you like?

`max_concurrency` in map/parallel doesn't throttle async operations like `invoke`

## Problem Statement

When using `context.map()` or `context.parallel()` with `max_concurrency` set, you'd expect it to limit the number of concurrent operations in progress. However, the current implementation only limits the number of active threads, not the number of in-flight async operations.

This means when map/parallel iterations contain async operations like `context.invoke()`, `context.wait()`, or `context.wait_for_condition()` - all items can be started nearly simultaneously regardless of the `max_concurrency` setting. While this does optimize processing, it does seem surprising given the ` max_concurrency` semantic.

### How it works today

1. All tasks are submitted to a `ThreadPoolExecutor` immediately
2. The thread pool has `max_workers = max_concurrency` threads
3. When a task calls an async operation like `context.invoke()`:
   - It creates a checkpoint
   - Raises `SuspendExecution` 
   - The thread completes and returns to the pool
   - The next queued task starts immediately
4. Result: All async operations are initiated as fast as threads can cycle through items (typically milliseconds per item)

### Example

```python
# User expects: Only 5 Lambda invocations in-flight at once
# Actual behavior: All 100 Lambda invocations start within ~1 second
result = context.map(
    items=range(100),
    func=lambda ctx, item, idx, items: ctx.invoke("MyFunction", item),
    config=MapConfig(max_concurrency=5)
)
```

With `max_concurrency=5` and 100 items:
- Tasks 0-4 start in 5 threads
- Each calls `invoke()` and suspends (releases thread)
- Tasks 5-9 start immediately in the freed threads
- This continues rapidly until all 100 invocations are started
- All 100 Lambda functions are invoked concurrently

## Expected Behavior

`max_concurrency` should limit the number of **in-flight operations** (including suspended ones), not just active threads.

### Desired behavior

1. Start `max_concurrency` tasks initially
2. When a task **completes** (reaches SUCCEEDED or FAILED state), start the next pending task
3. Suspended tasks remain "in-flight" and count against the concurrency limit
4. If all in-flight tasks are suspended, the parent suspends too

### Example with expected behavior

```python
result = context.map(
    items=range(100),
    func=lambda ctx, item, idx, items: ctx.invoke("MyFunction", item),
    config=MapConfig(max_concurrency=5)
)
```

With `max_concurrency=5` and 100 items:
- Tasks 0-4 start and invoke Lambda functions
- Parent suspends waiting for responses
- When Lambda 0 completes, task 5 starts
- When Lambda 1 completes, task 6 starts
- At most 5 Lambda invocations are in-flight at any time

### Current workarounds

Manually pre-batch their arrays:

```python
# Manual batching to achieve max 5 concurrent invocations
def batch_items(items, batch_size):
    return [items[i:i + batch_size] for i in range(0, len(items), batch_size)]

batches = batch_items(items, max_concurrency=5)
for batch in batches:
    for i, item in enumerate(batch):
        context.invoke(f"batch_{i}", "MyFunction", item)
```

This is not great.

## Related Discussion

See [Discussion #278](https://github.com/aws/aws-durable-execution-sdk-python/discussions/278) for the initial  report.

## Proposed Solution

Implement dynamic task submission where:

1. Track "in-flight" tasks (PENDING + RUNNING + SUSPENDED states)
2. Only submit new tasks when `in_flight_count < max_concurrency`
3. Decrement `in_flight_count` only when tasks reach terminal states (COMPLETED or FAILED)
4. Suspended tasks remain in-flight and block new submissions

## Acceptance Criteria

- [ ] `max_concurrency` limits in-flight operations, not just active threads
- [ ] Suspended tasks count against the concurrency limit
- [ ] New tasks only start when existing tasks complete (not when they suspend)
- [ ] All existing tests pass
- [ ] New tests verify the corrected concurrency behavior
- [ ] Documentation updated to clarify concurrency semantics
- [ ] Consider backward compatibility strategy

### Code locations

- `src/aws_durable_execution_sdk_python/concurrency/executor.py`: `ConcurrentExecutor.execute()`
- `src/aws_durable_execution_sdk_python/operation/map.py`: Map operation
- `src/aws_durable_execution_sdk_python/operation/parallel.py`: Parallel operation

### Key insight

The issue stems from submitting all tasks upfront:
```python
futures = [submit_task(exe_state) for exe_state in self.executables_with_state]
```

Replace this with dynamic submission based on in-flight count.

### Possible Implementation

_No response_

No

### Does this require an RFC?

No

### Additional Context

_No response_

yaythomas:
*Description of changes:*

**BREAKING CHANGE** -- see the behavior changes section; config validation, result shape, and the large-result summary payload format change in this PR.
`max_concurrency` previously sized the thread pool only, so branches that suspended (invoke, wait, callbacks) released their thread and the next item started immediately. All items were initiated at once regardless of the configured limit.

Rework the concurrency executor around a single coordinator loop:

- The calling thread owns all branch state and schedules work; worker threads only run branches and report outcomes on a queue. No locks remain in the module.
- `max_concurrency` now bounds in-flight branches. A suspended branch keeps its slot until it reaches a terminal state, matching the JS and Java SDKs. When every in-flight branch is suspended, the parent suspends with the earliest resume timestamp, or indefinitely.
- Branches start in index order and operation ids derive from the branch index, so scheduling stays deterministic across invocations.
- Timed suspends resume in-process via the event queue timeout, replacing the `TimerScheduler` background thread.
- Consolidate completion logic (thresholds and reason inference) into `CompletionPolicy`, replacing `ExecutionCounters` and the duplicated logic in `BatchResult`.
- Stop scheduling when a branch is orphaned: an ancestor completed, so all further checkpoints under it would be rejected.
- Record the completion decision in an SDK-owned envelope and obey it on replay. When a map/parallel result exceeds the checkpoint size limit, the SDK now writes the parent payload itself: `type`, `totalCount`, `completionReason`, the started-branch index set (`startedIndexes`, or `completedIndexes` when smaller), and stats fields (`startedCount`, `successCount`, `failureCount`, `status`). Replay reads the record and reconstructs the exact live result — same items, same statuses, same reason — instead of re-deriving from child checkpoints. A configured `summary_generator` no longer produces the payload: its output is stored verbatim under the envelope's `summary` key for observability and is never read by the SDK, so custom generators get exact replay and cannot displace the record. An exception raised by a custom generator propagates and fails the operation, matching the JS SDK. The envelope is checkpointed as provided; a payload exceeding the checkpoint size limit fails the operation with the service's payload validation error (terminal map failure, no retry loop -- checkpoint API 4xx errors are classified non-retryable). The envelope is written and parsed with plain JSON, independent of any configured serdes. Payloads without the record (checkpoints written by earlier SDK versions) fall back to checkpoint derivation, preserving old behavior. This also makes FLAT-nesting results reconstructable on replay, which checkpoint derivation cannot discriminate.

**Behavior changes**

- `BatchResult` omits never-started branches (matching the JS SDK). Previously every item appeared as STARTED even when an early completion meant it never ran; `total_count` now reflects branches that actually started.
- Configs with only `min_successful` set no longer fail fast on the first failure. The old tolerance check ignored `min_successful` when deciding whether criteria existed, so a single failure stopped scheduling; scheduling now continues past failures until `min_successful` is reached or all items finish, matching the JS and Java SDKs.
- Config bounds are validated at construction, matching the Java SDK: `max_concurrency` and `min_successful` must be at least 1, `tolerated_failure_count` non-negative, `tolerated_failure_percentage` within 0–100. Previously `max_concurrency=0` silently meant unlimited and `min_successful=0` silently meant all items.
- `min_successful` greater than the item count now raises `ValidationError` at the operation call, matching the Java SDK. Previously it silently degraded to "complete when everything finishes".
- `CompletionConfig.all_completed()` now tolerates all failures (`tolerated_failure_percentage=100`), matching its documentation and the Java SDK. Previously its all-`None` fields selected the fail-fast default, so a single failure stopped scheduling -- the opposite of the factory's name.
- Large map/parallel results now checkpoint a summary at all. The summary generator was previously attached to the iteration child contexts rather than the map/parallel context itself, so the parent payload of a large result was an empty string and the execution history carried no summary.
- A branch whose own item result exceeds the checkpoint size limit no longer fails. The default summary generator was previously applied to the raw item value, which raised and silently failed the branch. Large item results now checkpoint an empty payload with replay-children set and re-execute on replay, matching the JS SDK.
- The `MapSummaryGenerator` and `ParallelSummaryGenerator` classes are removed (not exported from the package root; reachable only by deep import, referenced by no examples or docs, and never invoked on the released large-result path). The envelope writes their fields directly.
- The large-result summary payload is the envelope for every execution. The default fields are a superset of the previous default summary (same keys and meanings, plus the index set, and `startedCount` now on map as well as parallel). Output from a custom `summary_generator` appears under the `summary` key instead of replacing the whole payload.

**Testing**

- Unit: 1424 tests, typecheck and format clean, examples suite green.
- Fatal errors from branches (for example a background checkpoint failure) propagate to the caller instead of being recorded as branch failures, and failed FLAT branches carry the same error type live and on replay.
- Cloud conformance (in-repo `aws-durable-execution-sdk-python-conformance-tests`, deployed against this branch's SDK): all 9 suites pass with 145/145 validations — map 20/20 (including 9-16 MapLargeResult, which exercises the summarized-replay path, and the early-stop cases for fail-fast, min-successful, and failure-tolerance semantics), parallel 22/22, callback 19/19, child 17/17, invoke 16/16, step 20/20, wait 5/5, wait_for_callback 15/15, wait_for_condition 11/11.

*Issue #, if available:*
Closes #279

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
@@ -4,21 +4,24 @@ from __future__ import annotations
 
 import heapq
 import logging
-import threading
+import queue
 import time
-from abc import ABC, abstractmethod
-from concurrent.futures import Future, ThreadPoolExecutor
-from typing import TYPE_CHECKING, Generic, Self, TypeVar
+from collections import deque
+from concurrent.futures import ThreadPoolExecutor
+from typing import TYPE_CHECKING, Generic, TypeVar, cast
 
 from aws_durable_execution_sdk_python.concurrency.models import (
     BatchItem,
     BatchItemStatus,
     BatchResult,
+    Branch,
+    BranchEvent,
+    BranchEventKind,
     BranchStatus,
+    CompletionPolicy,
+    CompletionReason,
+    CompletionRecord,
     Executable,
-    ExecutableWithState,
-    ExecutionCounters,
-    SuspendResult,
 )
 from aws_durable_execution_sdk_python.config import (
     ChildConfig,
@@ -26,24 +29,29 @@ from aws_durable_execution_sdk_python.config import (
 )
 from aws_durable_execution_sdk_python.exceptions import (
     DurableOperationError,
+    InvalidStateError,
     OrphanedChildException,
     SuspendExecution,
     TimedSuspendExecution,
 )
-from aws_durable_execution_sdk_python.identifier import OperationIdentifier
+from aws_durable_execution_sdk_python.identifier import (
+    OperationIdentifier,
+    OperationIdNamespace,
+)
 from aws_durable_execution_sdk_python.lambda_service import ErrorObject
 from aws_durable_execution_sdk_python.operation.child import child_handler
 
 
 if TYPE_CHECKING:
     from collections.abc import Callable
-
     from aws_durable_execution_sdk_python.config import CompletionConfig
     from aws_durable_execution_sdk_python.context import DurableContext
     from aws_durable_execution_sdk_python.lambda_service import OperationSubType
     from aws_durable_execution_sdk_python.serdes import SerDes
-    from aws_durable_execution_sdk_python.state import ExecutionState
-    from aws_durable_execution_sdk_python.types import SummaryGenerator
+    from aws_durable_execution_sdk_python.state import (
+        CheckpointedResult,
+        ExecutionState,
+    )
 
 
 logger = logging.getLogger(__name__)
@@ -55,99 +63,44 @@ CallableType = TypeVar("CallableType")
 ResultType = TypeVar("ResultType")
 
 
-# region concurrency logic
-class TimerScheduler:
-    """Manage timed suspend tasks with a background timer thread."""
+def _branch_error_object(err: Exception) -> ErrorObject:
+    """Convert a failed branch's error for the batch result.
 
-    def __init__(
-        self, resubmit_callback: Callable[[list[ExecutableWithState]], None]
-    ) -> None:
-        self.resubmit_callback = resubmit_callback
-        self._pending_resumes: list[tuple[float, int, ExecutableWithState]] = []
-        self._lock = threading.Lock()
-        self._schedule_counter = 0
-        self._shutdown = threading.Event()
-        self._timer_thread = threading.Thread(target=self._timer_loop, daemon=True)
-        self._timer_thread.start()
-
-    def __enter__(self) -> Self:
-        return self
-
-    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
-        self.shutdown()
-
-    def schedule_resume(
-        self, exe_state: ExecutableWithState, resume_time: float
-    ) -> None:
-        """Schedule a task to resume at the specified time.
-
-        Uses a counter as a tie-breaker to ensure FIFO ordering when multiple
-        tasks have the same resume_time, preventing TypeError from comparing
-        ExecutableWithState objects.
-        """
-        with self._lock:
-            heapq.heappush(
-                self._pending_resumes,
-                (resume_time, self._schedule_counter, exe_state),
-            )
-            self._schedule_counter += 1
-
-    def shutdown(self) -> None:
-        """Shutdown the timer thread and cancel all pending resumes."""
-        self._shutdown.set()
-        self._timer_thread.join(timeout=1.0)
-        with self._lock:
-            self._pending_resumes.clear()
-
-    def _timer_loop(self) -> None:
-        """Background thread that processes timed resumes."""
-        while not self._shutdown.is_set():
-            next_resume_time = None
-
-            with self._lock:
-                if self._pending_resumes:
-                    next_resume_time = self._pending_resumes[0][0]
-
-            if next_resume_time is None:
-                # No pending resumes, wait a bit and check again
-                self._shutdown.wait(timeout=0.1)
-                continue
-
-            current_time = time.time()
-            if current_time >= next_resume_time:
-                # Drain every due resume under the lock, transitioning each to
-                # PENDING atomically with the pop. Keeping pop+reset_to_pending
-                # together is required: should_execution_suspend reads branch
-                # status without this lock, so an item that is removed from the
-                # heap but still SUSPENDED_WITH_TIMEOUT could trigger a spurious
-                # parent suspend.
-                ready: list[ExecutableWithState] = []
-                with self._lock:
-                    while (
-                        self._pending_resumes
-                        and self._pending_resumes[0][0] <= current_time
-                    ):
-                        _, _, exe_state = heapq.heappop(self._pending_resumes)
-                        if exe_state.can_resume:
-                            exe_state.reset_to_pending()
-                            ready.append(exe_state)
-                # Resubmit outside the lock. Only the heap pop and the PENDING
-                # transition need the lock. The checkpoint refresh is a blocking
-                # network call and the submit hands work to the pool, so running
-                # them off the lock keeps timed resumes from serializing behind
-                # the network round trip and keeps the timer thread from
-                # re-entering this non-reentrant lock when a submitted future
-                # completes inline and its done-callback calls schedule_resume.
-                if ready:
-                    self.resubmit_callback(ready)
-            else:
-                # Wait until next resume time
-                wait_time = min(next_resume_time - current_time, 0.1)
-                self._shutdown.wait(timeout=wait_time)
+    Records the raw escaping type so live results, branch FAIL checkpoints,
+    and replay reconstruction all carry the same discriminator;
+    ``from_exception`` on the ChildContextError wrapper would instead
+    record the wrapper class name.
+    """
+    if isinstance(err, DurableOperationError):
+        return ErrorObject(
+            message=err.message,
+            type=err.error_type,
+            data=err.data,
+            stack_trace=err.stack_trace,
+        )
+    return ErrorObject.from_exception(err)
 
 
-class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
-    """Execute durable operations concurrently. This contains the execution logic for Map and Parallel."""
+class ConcurrentExecutor(Generic[CallableType, ResultType]):
+    """Execute durable operations concurrently. This contains the execution logic for Map and Parallel.
+
+    Scheduling model: a single coordinator loop runs on the calling thread
+    and owns all branch state. Worker threads run branches and report each
+    outcome as a :class:`BranchEvent` on a queue; they never mutate shared
+    state, so the module needs no locks.
+
+    ``max_concurrency`` bounds in-flight branches, not threads. A branch
+    that suspends (e.g. awaiting an invoke result or callback) keeps its
+    concurrency slot until it reaches a terminal state. New branches start
+    only when a slot frees up. When every in-flight branch is suspended and
+    no slot is available, the parent suspends too: with the earliest resume
+    timestamp when one exists, indefinitely otherwise.
+
+    Branches are always started in index order and operation ids derive
+    from the branch index, so scheduling is deterministic across
+    invocations. On re-invocation the previously started branches are
+    admitted first and replay from their checkpoints.
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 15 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
```diff
@@ -115,36 +111,26 @@ def map_handler(
...
-    #
-    # See TypeScript reference: aws-durable-execution-sdk-js/src/handlers/map-handler/map-handler.ts (~line 79)
+    map_config: MapConfig = config or MapConfig()
+    min_successful: int | None = map_config.completion_config.min_successful
+    if min_successful is not None and min_successful > len(items):
+        msg = (
+            f"min_successful cannot be greater than total items: "
+            f"{min_successful} > {len(items)}"
+        )
+        raise ValidationError(msg)
```
zhongkechen: Should ValidationError be checkpointed?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -522,19 +468,101 @@ def run_in_child_handler() -> ResultType:
...
+        """Re-derive one branch recorded as terminal.
+
+        Non-virtual branches have a terminal checkpoint: terminal branch
+        events are only emitted after the synchronous SUCCEED/FAIL
+        checkpoint call returned. Virtual (FLAT) branches never checkpoint
+        themselves, so re-executing the branch body over its inner
+        operations' checkpoints discriminates success from failure.
         """
-        Replay rather than re-run children.
+        operation_id: str = executor_context._create_step_id_for_logical_step(  # noqa: SLF001
```
zhongkechen: can we avoid calling internal method of executor_context?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
zhongkechen, line 88: Why don't we directly bind `item` to `func` here and put iteration name into `Executable` so that the base class doesn't have to look back into subclasses (MapExecutor/ParallelExecutor) for these fields.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -55,99 +60,44 @@
...
 
 
 class ConcurrentExecutor(ABC, Generic[CallableType, ResultType]):
-    """Execute durable operations concurrently. This contains the execution logic for Map and Parallel."""
+    """Execute durable operations concurrently. This contains the execution logic for Map and Parallel.
+
+    Scheduling model: a single coordinator loop runs on the calling thread
+    and owns all branch state. Worker threads run branches and report each
+    outcome as a :class:`BranchEvent` on a queue; they never mutate shared
+    state, so the module needs no locks.
```
zhongkechen, line 90: nice queue based lock free design

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -220,238 +144,260 @@ def get_iteration_name(self, index: int) -> str:
...
+                        # A dropped event would leave the counters stale and
+                        # hang the coordinator, so fail loudly instead.
+                        msg = f"Unhandled branch event: {event}"
+                        raise InvalidStateError(msg)
+        finally:
+            # Shutdown without waiting for running threads for early return
+            # when completion criteria are met (e.g., min_successful).
+            # Running threads continue in the background and raise
+            # OrphanedChildException on their next attempt to checkpoint.
+            pool.shutdown(wait=False, cancel_futures=True)
```
zhongkechen, line 289: Would this cause the threads and resources to accumulate if users heavily use early return with long running tasks

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
yaythomas: The raise occurs inside map_handler, which runs inside child_handler's try block, so it is checkpointed and the map context FAILs terminally with error_type: ValidationError, surfacing as a catchable ChildContextError.

That's deterministic (replay re-raises the recorded error), loud, and visible in history. And the current placement means the error behaves exactly like every other failure inside a map.

The alternative is to validate in context.py before the child context starts, so no STARTED-then-FAILED context appears in history, which is closer to matching Java's constructor validation + throw `IllegalArgumentException`?

So the diffs is 
a) ChildContextError wrapping ValidatorError, with a STARTED/FAILED checkpoint for the map (current)
vs
b) a bare exception at execution level?

There's prob an argument to make for either. I instinctively would've thought a STARTED/FAILED on the map makes sense since this can only rear its head at runtime rather than design time (like a compiler err), but happy to reconsider if you think otherwise?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
yaythomas: love it, implementing

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
yaythomas: we can and we should :-) change inbound

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
yaythomas: It would, sort of... but there's not much we can do about it

On early completion the pool is abandoned (`shutdown(wait=False)`). Any checkpoint attempt by an orphaned branch raises `OrphanedChildException` from a local state check (_parent_done in state.py, no network round trip), unwinding the thread immediately.

So a branch survives early completion only for the span of user code between SDK calls, and only checkpoint-free compute/sleep bodies run to natural completion.

Accumulation is possible within one invocation if the handler runs successive early-completing maps over long checkpoint-less branches. ThreadPoolExecutor has no kill and cancel_futures=True only stops queued work. This is bounded by the invocation lifetime (it has to be). Same behavior as before this PR.

The alternatives don't help. Python has no thread cancellation. A cancellation token doesn't work because the running code is customer code, and the SDK already checks at every operation boundary, which is every cancellation point we have. A shared bounded pool across maps would let zombie branches hold slots and silently shrink max_concurrency for later maps, trading inert threads for stalls. 

For comparison, JS parks abandoned branches at their first checkpoint attempt, and Java runs them to natural completion on an unbounded cached thread pool with only the final checkpoint suppressed (isOperationCompleted guard in handleChildContextSuccess). Python's unwind-at-first-checkpoint is the most aggressive reclamation of the three.

The practical guidance for consumers is the existing one that long-running branch work should checkpoint via steps, and branches doing that exit promptly.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
yaythomas: I'm going to do some prep work for #573 to service this while I'm at it...

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
zhongkechen: I'm afraid checkpointing validation error isn't consistent with the other operations.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
zhongkechen: In Java, the SDK waits for the threads to complete before returning, even when they are orphaned threads.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/map.py
yaythomas: oh, ha, yes, sorry, I didn't catch where you were going there, wait and wait_for_condition are the others and yes, you're right, they are using bare exception at execution level. Amended!

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
yaythomas: oooh, you mean to move the wait into the _current_ invocation, so the current invocation has to wait until they finish?

Whereas this was abandoning the thread and ending the invocation, which optimizes for _cold_ invocations (less wait time because we abandoned early), but the price we pay for that is on a warm invoke the thread might still be around in the environment and accumulate there.

okay, amended, so joining the threads before returning. Branches that checkpoint still unwind at their first attempt after abandonment, so the join only bites checkpoint-free branches.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -828,6 +838,18 @@ def stop_checkpointing(self) -> None:
...
+        """Register a branch thread pool for joining at invocation end.
+
+        Concurrency coordinators register their pool on creation. close()
+        joins every registered pool before stopping the checkpoint batcher,
+        so branches abandoned by early completion finish (or unwind on
+        OrphanedChildException at their next checkpoint attempt) before the
+        invocation returns.
+        """
+        with self._branch_pools_lock:
+            self._branch_pools.append(pool)
```
zhongkechen, line 851: The max_concurrency is no longer bound by the max threads in the pool. Maybe we can consider share the same pool for all concurrency operations.

zhongkechen (approved):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
yaythomas: The pool is still created with max_workers=max_in_flight, where max_in_flight = max_concurrency or len(executables))? And submission is gated on in_flight < max_in_flight.

To make sure I'm understanding, you're proposing one effectively-unbounded cached pool? 

While considering this, Python’s ThreadPoolExecutor does not reap idle threads, so the pool retains its high-water thread count until invocation end, and we'd lose isolation. Maybe debugging thread names and dumps get harder too?

Issue #120, opened 2025-11-04
Consolidate error handling


yaythomas:
## overview
Currently exceptions from a step raise as a `CallableRuntimeError`. 

https://github.com/aws/aws-durable-execution-sdk-python/blob/3aea45710cc8da739f6d6ddb3b4120002b35f81d/src/aws_durable_execution_sdk_python/state.py#L188C1-L201C53

A recent refactor in Ts introduced some different/new error types.

**ChildContextError Consistency:**
- Always wrap errors in ChildContextError in executeChildContext
- Preserve original error message and pass error as cause
- Handle both Error objects and non-Error values consistently
- Add test for child context error  with no Error field

**StepError Consistency:**
- Wrap generic errors in StepError at handler level
- Preserve DurableOperationError instances and specific error types
- Add test to verify Error objects are wrapped with correct message/cause

**StepInterruptedError Handling:**
- Pass StepInterruptedError to retry strategy for decision making
- Wrap StepInterruptedError in StepError when retry is exhausted
- Update tests to verify proper error wrapping and retry flow

per https://github.com/aws/aws-durable-execution-sdk-js/pull/212

## ac
1. Add ChiildContextError + implement in ChildContext handler
2. Add StepError + implement in Step handler (replacing CallabeRuntimeError)
3. StepInterruptedError raises as a StepError when retries exhausted

Issue #297, opened 2026-02-07
[Bug]: error_type on CallableRuntimeError is lost when using ctx.parallel() — double-wrapped by ErrorObject.from_exception()


Iamrodos:
### Expected Behavior

When a custom exception is raised inside a `ctx.parallel()` branch, I'd expect `error_type` on the resulting `CallableRuntimeError` (from `batch_result.throw_if_error()`) to be my original exception's class name — the same way it works with `ctx.run_in_child_context()`, where `error_type` correctly preserves the original class name.

```python
try:
    result.throw_if_error()
except CallableRuntimeError as e:
    print(e.error_type)  # Expected: 'PermanentFailure'
```

### Actual Behavior

`error_type` always comes back as `'CallableRuntimeError'` instead of my original exception type. It looks like the exception goes through `ErrorObject.from_exception()` twice:

**First wrap** (in `ChildOperationExecutor.execute()`, `operation/child.py`):
- My code raises `PermanentFailure("Invalid input data")`
- `ErrorObject.from_exception(e)` → `ErrorObject(type="PermanentFailure")` — correct
- `to_callable_runtime_error()` → `CallableRuntimeError(error_type="PermanentFailure")` — correct

**Second wrap** (in `ConcurrentExecutor._create_result()`, `concurrency/executor.py`):
- `ErrorObject.from_exception(executable.error)` is called on the already-wrapped `CallableRuntimeError`
- `type(exception).__name__` is now `CallableRuntimeError`, so the original `error_type="PermanentFailure"` is lost

I'm wondering if I'm doing something wrong, or if this is a gap in how exceptions flow through parallel branches?

### Steps to Reproduce

1. Define a custom exception and a durable function that uses `ctx.parallel()`:

```python
class PermanentFailure(Exception):
    pass

def my_handler(ctx):
    def branch1(child_ctx):
        raise PermanentFailure("Invalid input data")

    result = ctx.parallel([branch1])
    result.throw_if_error()
```

2. Catch the `CallableRuntimeError` and inspect `error_type`:

```python
try:
    result.throw_if_error()
except CallableRuntimeError as e:
    print(e.error_type)  # Returns 'CallableRuntimeError', not 'PermanentFailure'
```

3. Compare with `ctx.run_in_child_context()` which correctly preserves `error_type='PermanentFailure'` (only one wrap).

### SDK Version

1.3.0

### Python Version

3.14

### Is this a regression?

No

### Additional Context

[...]

PR #540, opened 2026-07-14
feat(errors): typed per-operation error hierarchy

### LINKED ISSUE #546: `wait_for_condition` and `run_in_child_context` persist a terminal FAIL checkpoint for a retryable InvocationError
### Problem
When an InvocationError escapes a wait_for_condition check (or run_in_child_context body), execute() persists a FAIL checkpoint (is_sync=True) and then re-raises to trigger a Lambda retry. But the committed FAIL makes the retry useless: on replay, check_result_status() sees is_failed() → raise_callable_error() → CallableRuntimeError → handler returns FAILED. The check/body never re-runs, and the error type silently changes from InvocationError to CallableRuntimeError.

### Contrast
step handles this correctly: a retryable error goes through retry_handler, which writes a RETRY checkpoint (non-terminal, re-executes) and only writes FAIL once retries are exhausted.

```
def check(state, ctx):
    raise InvocationError("transient failure")
context.wait_for_condition(check=check, config=WaitForConditionConfig(...))
```

First run: FAIL persisted, InvocationError bubbles, Lambda retries. 
Retry: short-circuits to CallableRuntimeError, check never re-runs.

### Proposed fix
Re-raise InvocationError before writing the FAIL checkpoint in both wait_for_condition.execute() and child.execute() (including is_virtual). Keep writing FAIL for non-retryable errors (e.g. BotoClientError with is_retryable() == False).

ayushiahjolia:
*Issue #, if available:* 
https://github.com/aws/aws-durable-execution-sdk-python/issues/120
https://github.com/aws/aws-durable-execution-sdk-python/issues/546

*Description of changes:*

Replaces the single catch-all `CallableRuntimeError` with a typed, per-operation error hierarchy. A failed step, invoke, child context, or wait_for_condition now surfaces as a distinct, catchable exception type. It also lands the `from_exception`/`throw_if_error` changes those types require. The `CallbackError` rework and the nested-boundary regression test are intentionally deferred to follow-up PRs.

#### Public API

- Exports the typed operation errors and their base from the package root: `StepError`, `InvokeError`, `ChildContextError`, `WaitForConditionError`, and `DurableOperationError`, plus `ExecutionError`. Users can now `from aws_durable_execution_sdk_python import StepError` (etc.) and catch on the specific type. `CallbackError` and `SerDesError` are intentionally not exported yet (see follow-ups).

#### Breaking changes

- Operation failures now raise the typed subclasses (StepError, InvokeError, ChildContextError, WaitForConditionError) instead of CallableRuntimeError (removed).
- Catch on the exception class to identify the operation. The error_type attribute carries the original Python exception name (e.g. "ValueError"); the original message is on .message, and the original error is attached as __cause__ (the live exception on first run, a reconstructed stand-in on replay). At the wire/checkpoint level, an escaping wrapper is recorded under its class-name discriminator (e.g. "StepError") one level up, while the operation's own FAIL checkpoint records the raw error type.
- wait_for_condition check failures are now wrapped in WaitForConditionError rather than re-raised raw.
- wait_for_callback/callback failures now surface as CallbackError (previously CallableRuntimeError); CallbackError remains in the termination tree until the follow-up.
- An exhausted interrupted step fails immediately as StepError rather than round-tripping through a Lambda retry.

#### Fixes #546 - non-retryable failures no longer leave a STARTED op in a FAILED execution

- In child context and wait_for_condition, a non-retryable `InvocationError` (e.g. a `BotoClientError` classified EXECUTION) now falls through to the FAIL checkpoint before being wrapped, so the operation isn't left STARTED inside a FAILED execution. A retryable `InvocationError` still re-raises so the backend can retry, matching the JS and Java SDKs (which resolve retry/suspend before the branch boundary and terminally FAIL everything else).

#### Testing

- Updated/added unit tests across exceptions, lambda_service, state, concurrency, and the four operation handlers.
- Added coverage for: registry + from_error_fields (including fallback to the base DurableOperationError for unknown/user discriminators), from_exception field preservation across a single operation boundary, raise_operation_error, wait_for_condition control-flow errors propagating unwrapped, non-retryable InvocationError wrapping in child and wait_for_condition (#546), and the execution result recording the StepError discriminator while the step's own FAIL checkpoint records the raw error type.
- Added an integration test and an example; the example also verifies the caught error is identical across a wait/replay boundary.

#### Follow-ups

- Move `CallbackError` under `DurableOperationError` with graded subclasses (`CallbackExternalError`, `CallbackTimeoutError`, `CallbackSubmitterError`) and drop `CALLBACK_ERROR` metadata.
- Regression test for `data` surviving ≥2 nested `run_in_child_context` boundaries

### DIFF
--- a/packages/aws-durable-execution-sdk-python-examples/test/wait_for_callback/test_wait_for_callback_failure.py
+++ b/packages/aws-durable-execution-sdk-python-examples/test/wait_for_callback/test_wait_for_callback_failure.py
@@ -21,7 +21,11 @@ def test_wait_for_callback_failure(durable_runner):
 
     assert result.status is InvocationStatus.FAILED
     assert isinstance(result.error, ErrorObject)
+    # The callback failure raises CallbackError inside the child context that
+    # wait_for_callback runs in; it is wrapped as ChildContextError so first run
+    # and replay surface an identical type. The original CallbackError type is
+    # preserved on the error's error_type/__cause__.
     assert result.error.to_dict() == {
         "ErrorMessage": "my callback error",
-        "ErrorType": "CallableRuntimeError",
+        "ErrorType": "ChildContextError",
     }
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/__init__.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/__init__.py
@@ -18,9 +18,15 @@ from aws_durable_execution_sdk_python.context import (
 
 # Most common exceptions - users need to handle these exceptions
 from aws_durable_execution_sdk_python.exceptions import (
+    ChildContextError,
     DurableExecutionsError,
+    DurableOperationError,
+    ExecutionError,
     InvocationError,
+    InvokeError,
+    StepError,
     ValidationError,
+    WaitForConditionError,
 )
 
 # Core decorator - used in every durable function
@@ -33,12 +39,18 @@ from aws_durable_execution_sdk_python.types import StepContext
 
 __all__ = [
     "BatchResult",
+    "ChildContextError",
     "DurableContext",
     "DurableExecutionsError",
+    "DurableOperationError",
+    "ExecutionError",
     "InvocationError",
+    "InvokeError",
     "ParallelBranch",
     "StepContext",
+    "StepError",
     "ValidationError",
+    "WaitForConditionError",
     "WithRetryConfig",
     "__version__",
     "durable_execution",
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
@@ -11,6 +11,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Generic, TypeVar
 
 from aws_durable_execution_sdk_python.exceptions import (
+    ChildContextError,
     InvalidStateError,
     SuspendExecution,
 )
@@ -260,8 +261,12 @@ class BatchResult(Generic[R], BatchResultProtocol[R]):  # noqa: PYI059
             (item.error for item in self.all if item.status is BatchItemStatus.FAILED),
             None,
         )
-        if first_error:
-            raise first_error.to_callable_runtime_error()
+        if first_error is not None:
+            # Each item runs in a child context, so a failed item surfaces as a
+            # ChildContextError wrapping the escaping error (reconstructed as its
+            # typed form where possible, with __cause__ set). Remaining errors
+            # stay in get_errors().
+            first_error.raise_as_operation_error(ChildContextError)
 
     def get_results(self) -> list[R]:
         return [
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
@@ -6,7 +6,6 @@ Avoid any non-stdlib references in this module, it is at the bottom of the depen
 from __future__ import annotations
 
 import time
-from dataclasses import dataclass
 from enum import Enum
 from typing import TYPE_CHECKING, Self, TypedDict
 
@@ -266,25 +265,88 @@ class InvalidStateError(DurableExecutionsError):
     """Raised when an operation is attempted on an object in an invalid state."""
 
 
-class UserlandError(DurableExecutionsError):
-    """Failure in user-land - i.e code passed into durable executions from the caller."""
+class DurableOperationError(DurableExecutionsError):
+    """Base class for typed, per-operation failures.
 
+    Wraps a failure that escaped a Durable Function operation (step, invoke,
+    child context, wait_for_condition). The concrete class identifies the
+    operation kind (so callers can ``except StepError``); the escaping error is
+    preserved as ``__cause__`` on the first run and reconstructed on replay.
 
-class CallableRuntimeError(UserlandError):
-    """This error wraps any failure from inside the callable code that you pass to a Durable Function operation."""
+    Attributes:
+        message: Human-readable failure message.
+        error_type: Type name of the error that escaped the operation (the
+            original/inner error, e.g. ``"ValueError"``), not the operation
+            class. Defaults to the class name only when no escaping error is
+            known.
+        data: Optional serialized error payload, preserved across operation
+            boundaries.
+        stack_trace: Optional stack trace lines captured from the origin.
+    """
 
     def __init__(
         self,
-        message: str | None,
-        error_type: str | None,
-        data: str | None,
-        stack_trace: list[str] | None,
+        message: str | None = None,
+        error_type: str | None = None,
+        data: str | None = None,
+        stack_trace: list[str] | None = None,
     ) -> None:
         super().__init__(message)
-        self.message = message
-        self.error_type = error_type
-        self.data = data
-        self.stack_trace = stack_trace
+        self.message: str | None = message
+        self.error_type: str = error_type or type(self).__name__
+        self.data: str | None = data
+        self.stack_trace: list[str] | None = stack_trace
+
+    @classmethod
+    def from_error_fields(
+        cls,
+        error_type: str | None,
+        message: str | None,
+        data: str | None,
+        stack_trace: list[str] | None,
+    ) -> DurableOperationError:
+        """Rebuild the correct subclass from serialized checkpoint error fields.
+
+        Looks the discriminator up in the reconstruction registry, falling back
+        to the base :class:`DurableOperationError` when ``error_type`` is unknown
+        (e.g. a downstream error surfaced by an async invoke/callback checkpoint).
+        """
+        target_cls: type[DurableOperationError] = _DURABLE_OPERATION_ERROR_REGISTRY.get(
+            error_type or "", DurableOperationError
+        )
+        return target_cls(
+            message=message,
+            error_type=error_type,
+            data=data,
+            stack_trace=stack_trace,
+        )
+
+
+class StepError(DurableOperationError):
+    """Raised when a step operation fails."""
+
+
+class InvokeError(DurableOperationError):
+    """Raised when a durable invoke operation fails."""
+
+
+class ChildContextError(DurableOperationError):
+    """Raised when a child context (run_in_child_context, map, parallel) fails."""
+
+
+class WaitForConditionError(DurableOperationError):
+    """Raised when a wait_for_condition operation fails."""
+
+
+# Reconstruction registry for replay: only the SDK's own operation-error types.
+# Any other discriminator falls back to DurableOperationError in
+# from_error_fields, so we never call a constructor the SDK doesn't control.
+_DURABLE_OPERATION_ERROR_REGISTRY: dict[str, type[DurableOperationError]] = {
+    "StepError": StepError,
+    "InvokeError": InvokeError,
+    "ChildContextError": ChildContextError,
+    "WaitForConditionError": WaitForConditionError,
+}
 
 
 class StepInterruptedError(InvocationError):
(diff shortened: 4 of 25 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
```diff
@@ -239,6 +241,24 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
...
+            if isinstance(e, (InvocationError, ExecutionError)):
+                if not self.is_virtual:
+                    self.state.create_checkpoint(
+                        operation_update=OperationUpdate.create_context_fail(
+                            identifier=self.operation_identifier,
+                            error=ErrorObject.from_exception(e),
+                            sub_type=self.sub_type,
+                        )
+                    )
+                raise
```
zhongkechen, line 246: Not replay safe if ExecutionError is thrown

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -235,6 +236,17 @@ def from_dict(cls, data: MutableMapping[str, Any]) -> ErrorObject:
 
     @classmethod
     def from_exception(cls, exception: Exception) -> ErrorObject:
+        # The wire ErrorType is always the class name (the discriminant): a raw
+        # error records its own type; a typed wrapper records its operation kind.
+        # For a DurableOperationError also preserve its own data and stack_trace
+        # so the inner info survives across a single operation boundary.
+        if isinstance(exception, DurableOperationError):
```
zhongkechen, line 243: Issue #297 again?

ayushiahjolia: yes I am going to fix it separately

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
ayushiahjolia: Good catch. first-run raises raw ExecutionError but replay gives ChildContextError. I'll wrap ExecutionError as ChildContextError so first run and replay match, and while I am here, I'll do the same in wait_for_condition.

ayushiahjolia: Bug for InvocationError - https://github.com/aws/aws-durable-execution-sdk-python/issues/546

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -266,25 +270,81 @@ class InvalidStateError(DurableExecutionsError):
...
+    ) -> None:
+        super().__init__(message)
+        self.message: str | None = message
+        self.error_type: str = error_type or type(self).__name__
+        self.data: str | None = data
+        self.stack_trace: list[str] | None = stack_trace
+
+    def __init_subclass__(cls, **kwargs: object) -> None:
+        super().__init_subclass__(**kwargs)
+        _DURABLE_OPERATION_ERROR_REGISTRY[cls.__name__] = cls
```
yaythomas: `__init_subclass__` registers every subclass anyone defines, including user-space code, so the wire-reconstruction path ends up calling constructors the SDK doesn't control?

what happens if this raises out of the Step?
```
class CustomError(StepError):
    def __init__(self, my_arg: str):
        super().__init__(f"arb: {my_arg}")
```

1. The step fails, and the FAIL checkpoint records ErrorType: "CustomError" (the checkpoint always stores the raw escaping type.
2. On retry-exhaustion (first run) or replay, the SDK reconstructs: registry lookup finds `CustomError`, calls it with `message=..., error_type=..., data=..., stack_trace=....`
3. But __init__ only accepts my_arg, so

```
TypeError: PaymentDeclinedError.__init__() got an unexpected keyword argument 'message'
```


It might be an idea to close the registry, so instead of __init_subclass__ with an explicit dict literal of the SDK's own classes, built once after the class definitions, without metaclass magic. i.e something like

```
@classmethod
def from_error_fields(cls, error_type, message, data, stack_trace) -> DurableOperationError:
    target_cls: type[DurableOperationError] = _DURABLE_OPERATION_ERROR_REGISTRY.get(
        error_type or "", DurableOperationError
    )
    return target_cls(message=message, error_type=error_type, data=data, stack_trace=stack_trace)
```

For comparison, the JS reference implementation demotes any unknown type to `StepError`, so a JS user subclassing `StepError` gets the same fallback.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
yaythomas: thanks! will #297 stay open for tracking?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -222,20 +222,29 @@ def is_replay_children(self) -> bool:
             return False
         return op.context_details.replay_children if op.context_details else False
 
-    def raise_callable_error(self, msg: str | None = None) -> None:
+    def raise_operation_error(
+        self,
+        operation_error_cls: type[DurableOperationError],
+        msg: str | None = None,
```
yaythomas, line 228: NoReturn, rather than None?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/__init__.py
```diff
@@ -18,9 +18,14 @@
...
 # Most common exceptions - users need to handle these exceptions
 from aws_durable_execution_sdk_python.exceptions import (
+    ChildContextError,
     DurableExecutionsError,
+    DurableOperationError,
     InvocationError,
+    InvokeError,
+    StepError,
     ValidationError,
+    WaitForConditionError,
```
yaythomas, line 29: CallbackError not here?

saying that, we could argue only to keep the "big" or notable err types here in the root export. InvocationError/ExecutionError/DurableOperationError/ValidationError? Not necessarily convinced of this, mind, you could equally argue these should be top-level exports too 😅

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
```diff
@@ -260,8 +262,17 @@ def throw_if_error(self) -> None:
...
             None,
         )
-        if first_error:
-            raise first_error.to_callable_runtime_error()
+        if first_error is not None:
+            # Each item runs in a child context, so a failed item surfaces as a
+            # ChildContextError wrapping the escaping error (reconstructed as its
+            # typed form where possible). Remaining errors stay in get_errors().
+            cause: DurableOperationError = first_error.to_durable_operation_error()
+            raise ChildContextError(
```
yaythomas: doesn't this duplicate what `first_error.raise_as_operation_error(ChildContextError)` does?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
yaythomas: Right now a non-retryable BotoClientError skips the FAIL checkpoint and the context op stays STARTED in a FAILED execution. Same in wait_for_condition.py. 

Gate prob needs to be `isinstance(e, InvocationError) and e.is_retryable()` per #546? If so, please add the check and a non-retryable test case.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
```diff
@@ -260,8 +262,17 @@ def throw_if_error(self) -> None:
             (item.error for item in self.all if item.status is BatchItemStatus.FAILED),
             None,
         )
-        if first_error:
-            raise first_error.to_callable_runtime_error()
+        if first_error is not None:
+            # Each item runs in a child context, so a failed item surfaces as a
+            # ChildContextError wrapping the escaping error (reconstructed as its
+            # typed form where possible). Remaining errors stay in get_errors().
+            cause: DurableOperationError = first_error.to_durable_operation_error()
```
yaythomas: `first_error` here holds a different record live vs replay, I think. 

the live path stores `from_exception(ChildContextError)` (`type="ChildContextError"`, executor.py `_create_result`), replay stores the raw checkpoint error (`type="ValueError"`). 

So the `ChildContextError` raised here carries a different `error_type` across replay.

This is a pre-existing issue, but we might as well take this oppo. to fix this and normalize the live path in `_create_result` to build the `ErrorObject` from the wrapper's fields (`type=e.error_type`) so both paths equal the checkpoint record.

packages/aws-durable-execution-sdk-python-examples/test/wait_for_callback/test_wait_for_callback_failure.py
```diff
@@ -21,7 +21,11 @@ def test_wait_for_callback_failure(durable_runner):
...
     assert result.status is InvocationStatus.FAILED
     assert isinstance(result.error, ErrorObject)
+    # The callback failure raises CallbackError inside the child context that
+    # wait_for_callback runs in; it is wrapped as ChildContextError so first run
+    # and replay surface an identical type. The original CallbackError type is
+    # preserved on the error's error_type/__cause__.
     assert result.error.to_dict() == {
         "ErrorMessage": "my callback error",
-        "ErrorType": "CallableRuntimeError",
+        "ErrorType": "ChildContextError",
```
yaythomas, line 30: I see in PR description callback is deferred for follow up, so jotting this down so I don't forget:

with this PR, `wait_for_callback` failures surface as `ChildContextError`. 

The desired end state is JS's behavior (`wait-for-callback-handler.ts` errorMapper): callback errors pass through typed (`CallbackError`/`CallbackTimeoutError`/`CallbackExternalError`), submitter `StepError` maps to `CallbackSubmitterError`.

So the graded-callback follow-up will change this surface again.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
ayushiahjolia: yes, I'll create a separate PR to fix it.

packages/aws-durable-execution-sdk-python-examples/test/wait_for_callback/test_wait_for_callback_failure.py
ayushiahjolia: yes that's the plan! I kept this PR short intentionally to make it easier to review.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
ayushiahjolia: Done - changed to NoReturn (both branches raise, and it lines up with `ErrorObject.raise_as_operation_error)`.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
ayushiahjolia: Yep, identical - `raise_as_operation_error(ChildContextError)` reconstructs the same wrapper with `__cause__` set via `from_error_fields`, which is exactly what the manual block did.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/__init__.py
ayushiahjolia: Looked at all customer-facing errors, which we should be adding _all_.
Added ExecutionError to the list and keeping CallbackError for the followup along with Callback error changes.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
ayushiahjolia: Fixed non retryable error case along with tests.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
ayushiahjolia: Fixed and added a regression test.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
ayushiahjolia: Fixed. I kept the fallback as DurableOperationError rather than JS's StepError, but happy to switch to StepError for cross-SDK parity.

yaythomas (approved):
nice, thank you so much! 🙇 

:shipit:

Issue #530, opened 2026-07-13
[Bug]: Wait for condition create_wait_strategy failed at runtime


wangyb-A:
### Expected Behavior

create_wait_strategy currently fails at runtime with
```
AttributeError: 'WaitDecision' object has no attribute 'should_continue'
```

Python: https://github.com/aws/aws-durable-execution-sdk-python/blob/main/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/waits.py#L81

WaitDecision is dead code should be removed

WaitForConditionConfig= parameters for theoperation(wait_strategy,initial_state,serdes).
WaitStrategyConfig= parameters for the exponential-backoff strategy

### Actual Behavior

See above

### SDK Version

1

### Python Version

3.14

### Is this a regression?

No

yaythomas:
Defect 1: type mismatch makes the built-in wait strategy unusable

create_wait_strategy returns a function producing WaitDecision (attribute should_wait). wait_for_condition consumes WaitForConditionDecision (attribute should_continue). Wiring the built-in strategy into WaitForConditionConfig fails on the first poll attempt:

```
AttributeError: 'WaitDecision' object has no attribute 'should_continue'
```

Confirmed against the released SDK (1.7.0). JS and Java return the correct decision type.

Python has no working built-in wait strategy for wait_for_condition today.

Defect 2: exhaustion semantics diverge from JS and Java

The max-attempts branch in create_wait_strategy (waits.py L81) returns no_wait(), which wait_for_condition treats as success. JS throws Error and Java throws WaitForConditionFailedException at the same point. Once Defect 1 is fixed, this branch becomes reachable and executions that exhaust polling attempts would complete successfully with partial state. Both defects must be fixed together. Fixing only the type mismatch converts today's loud failure into a silent wrong result.

AdityaAudi:
@wangyb-A @yaythomas I took a look and I can see both defects on the latest main. Wanted to write up what's going on and float an approach before opening a PR.

[...]

PR #541, opened 2026-07-16
fix(waits): repair strategy, fail on exhaustion

AdityaAudi:
*Fixes Issue #530*

**Summary**

The built-in `wait_for_condition` wait strategy has never been usable, and the bug hiding behind it would silently corrupt results once the surface error was fixed. This fixes both, together.

**Root cause**

1. Type mismatch — the built-in strategy can't be used at all.
`create_wait_strategy` returned a `WaitDecision` (attribute `should_wait`), but `wait_for_condition` consumes a `WaitForConditionDecision` (attribute `should_continue`). Passing the built-in strategy into a `WaitForConditionConfig` failed on the very first poll:

```
AttributeError: 'WaitDecision' object has no attribute 'should_continue'
```

Nothing in the tests or examples routed `create_wait_strategy` into a config, the shipped example hand-wrote its own strategy, so this stayed latent. The JS and Java SDKs return the correct decision type.

2. Exhaustion silently succeeds.
When polling attempts run out, the strategy returned a "stop" decision. Once defect #1 is fixed, that maps to `should_continue=False`, which the executor treats as **success**, completing with whatever partial state it happened to hold. JS throws `Error` and Java throws `WaitForConditionFailedException` at this point. Fixing only the type would have converted today's loud crash into a silent wrong result, which is why both are addressed here.

**Breaking change**

Executions that previously exhausted their polling attempts and completed "successfully" with partial state now fail with `WaitForConditionError`. This is the correct behavior and matches the other SDKs. Landing in v2; to be documented in the migration guide.

**Testing**

- New regression guard: the built-in strategy wires into a `WaitForConditionConfig` and returns the consumed type (closes the gap that let defect #1 ship).
- Condition met on the final attempt still succeeds.
- Both exhaustion surfaces covered:
  - **Live raise**: strategy raises → durable `FAIL` checkpoint carrying `error_type == "WaitForConditionError"` → `WaitForConditionError` propagates.
  - **Replay**: the `FAILED` checkpoint short-circuits the next invocation, surfacing as `CallableRuntimeError` with `error_type == "WaitForConditionError"` (typed reconstruction on replay lands with #120; asserting on `error_type` keeps the test stable across that change).
- Full suite: 1381 passed · `ruff` + format clean · `mypy` clean.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
@@ -242,7 +242,7 @@ class WaitForConditionOperationExecutor(OperationExecutor[T]):
             if delay_seconds is not None and delay_seconds < 1:
                 logger.warning(
                     (
-                        "WaitDecision delay_seconds step for id: %s, name: %s,"
+                        "wait_for_condition delay_seconds for id: %s, name: %s,"
                         "is %d < 1. Setting to minimum of 1 seconds."
                     ),
                     self.operation_identifier.operation_id,
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/waits.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/waits.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Generic
 
 from aws_durable_execution_sdk_python.config import Duration, JitterStrategy, T
+from aws_durable_execution_sdk_python.exceptions import WaitForConditionError
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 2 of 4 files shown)


zhongkechen (commented):
Thank you for contributing this fix and for the thorough regression coverage. The implementation looks good.

Before merge, please fix the commit-lint error: the current subject exceeds the repository limit of 50 characters (`invalid subject (must be <=50 chars)`). Please shorten the PR title/commit subject so the lint-commits check passes.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -266,6 +266,15 @@ class InvalidStateError(DurableExecutionsError):
     """Raised when an operation is attempted on an object in an invalid state."""
 
 
+class WaitForConditionError(DurableExecutionsError):
+    """Raised when a wait_for_condition exhausts its polling attempts before the
+    condition is met, matching the exhaustion semantics of the JS and Java SDKs.
+
+    #120 will reparent this under a shared operation-error base, so catch on the
+    concrete type rather than the parent that's about to move.
```
yaythomas: note (none-blocking): reparenting under an operation-error base won't affect anyone catching DurableExecutionsError (the new base will still sit under it).

Today:      WaitForConditionError → DurableExecutionsError
After #120: WaitForConditionError → OperationError → DurableExecutionsError

for future, we have to update this note when #120 lands.

AdityaAudi: Good call, I dropped the #120 note entirely so it can't go stale; the base class already documents the catch contract.

yaythomas:
#120 landed via PR #540, so this will need a rebase with
- drop the exceptions.py and __init__.py hunks entirely (use main's WaitForConditionError).
- Rewrite the replay test to expect WaitForConditionError; remove the CallableRuntimeError import.
- Keep the waits.py fix as-is; the small log-message tweak in operation/wait_for_condition.py is fine.

yaythomas (approved):

yaythomas:
thank you so much @AdityaAudi, great contribution, and thanks for your patience with all the earth moving with the v1 -> v2 moves!

PR #542, opened 2026-07-16
fix(otel): fix continuation span start time

nvasiu:
*Issue #, if available:*

N/A

*Description of changes:*

Update the continuation span's start time to use the operation's start time.

Current behaviour:

When `on_operation_end` continues a span from a previous invocation, it creates a span with `start_time` = `datetime.now()`. But the span's end time is set to the `end_time` recorded in the operation info.

If an operation quickly finishes before the span is started, `end_time` will be before the `datetime.now()` call, resulting in a negative span duration.


SilanHe (dismissed):

SilanHe:
can you merge after the tests pass?

zhongkechen:
Can we have a case assert on the start time?

zhongkechen (dismissed):

zhongkechen (approved):

Issue #407, opened 2026-06-01
[Bug]: Step operation should call ser/des before return the result


wangyb-A:
### Expected Behavior

I would like the the step operatoin to return the deseralized of seralized result on first run. So that the first run and replay will have same result.

### Actual Behavior

The first run will return the raw result. The seralized result was checkpointed. So replay will return the deseralized result.

### Steps to Reproduce

```
"""1-6: Custom serdes (per-step) - transforms result to uppercase."""

from typing import Any

from aws_durable_execution_sdk_python.config import StepConfig
from aws_durable_execution_sdk_python.context import (
    DurableContext,
    StepContext,
    durable_step,
)
from aws_durable_execution_sdk_python.execution import durable_execution
from aws_durable_execution_sdk_python.serdes import SerDes, SerDesContext


class UppercaseSerDes(SerDes[str]):
    """Custom serdes that transforms strings to uppercase on serialization."""

    def serialize(self, value: str, _: SerDesContext) -> str:
        return value.upper()

    def deserialize(self, data: str, _: SerDesContext) -> str:
        return data


@durable_step
def return_input(_step_context: StepContext, value: str) -> str:
    return value


@durable_execution
def handler(event: Any, context: DurableContext) -> str:
    result: str = context.step(
        return_input(event),
        config=StepConfig(serdes=UppercaseSerDes()),
    )
    return result

```

### SDK Version

1.5.0

### Python Version

3.14

### Is this a regression?

No

### Additional Context

Note that this might be a breaking change.

PR #543, opened 2026-07-16
fix(step): Return round-tripped result on first run

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/407

*Description of changes:*
A step serialized its result for the checkpoint but returned the raw in-memory value on the first run, while replay returned the value deserialized from the checkpoint. With a non-identity custom SerDes the first-run and replay results diverged, breaking the durability guarantee that a step yields the same value on every run.

Serialize then deserialize the result before returning it on the first run so the value matches what replay produces.

Added unit, integration, and example coverage for the first-run/replay equality with a non-identity SerDes.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py
@@ -249,7 +249,17 @@ class StepOperationExecutor(OperationExecutor[T]):
                 self.operation_identifier.operation_id,
                 self.operation_identifier.name,
             )
-            return raw_result  # noqa: TRY300
+            # Return the round-tripped value so the first run matches replay,
+            # which reconstructs the result by deserializing the checkpoint.
+            # A None payload is returned as-is, mirroring the replay path.
+            if serialized_result is None:
+                return None  # type: ignore[return-value]
+            return deserialize(  # noqa: TRY300
+                serdes=self.config.serdes,
+                data=serialized_result,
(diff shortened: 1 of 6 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py
```diff
@@ -249,7 +249,17 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
                 self.operation_identifier.operation_id,
                 self.operation_identifier.name,
             )
-            return raw_result  # noqa: TRY300
+            # Return the round-tripped value so the first run matches replay,
+            # which reconstructs the result by deserializing the checkpoint.
+            # A None payload is returned as-is, mirroring the replay path.
+            if serialized_result is None:
+                return None  # type: ignore[return-value]
```
zhongkechen, line 256: curious why we didn't add Optional[T] to the return type

zhongkechen (approved):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py
ayushiahjolia: Making this method return `Optional[T]` fails mypy with an "incompatible override" error. Using Optional[T] properly requires widening the return type across the whole operation chain and context.step(), telling every caller a step might return None even when it can't.

Issue #544, opened 2026-07-16
Fix `wait_for_condition` to return ser/des round-tripped result on first run


ayushiahjolia:
Problem:
wait_for_condition returns inconsistent results between the first run and replay when a custom SerDes is configured. On the first run, execute() returns the raw new_state produced by the check function. On replay, the value is reconstructed via deserialize(serialize(new_state)) from the checkpoint. If the SerDes is not a perfect round-trip (e.g. it normalizes, drops, or transforms fields), callers observe one value on the first execution and a different value after any replay.

Fix:
On the success path in operation/wait_for_condition.py, return the round-tripped value (deserialize of the serialized state) instead of the raw new_state, so the first run yields the same result the replay path would produce. Preserve None handling to match the replay behavior. Add a test using a non-identity SerDes
that asserts the first-run result equals the replayed result.

Issue #406, opened 2026-06-01
[Bug]: Child operation should call ser & des before return the result


wangyb-A:
### Expected Behavior

I would like the the child operatoin to return the deseralized of [seralized result](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py#L179) on first run. So that the first run and replay will have same result.

- Note that the ser/des should be applied to all child calls, including normal, large payload, virtual context.

### Actual Behavior

The first run will return the raw result. The seralized result was checkpointed. So replay will return the deseralized result.

### Steps to Reproduce

```
from typing import Any

from aws_durable_execution_sdk_python.config import ChildConfig
from aws_durable_execution_sdk_python.context import (
    DurableContext,
    StepContext,
    durable_step,
    durable_with_child_context,
)
from aws_durable_execution_sdk_python.execution import durable_execution
from aws_durable_execution_sdk_python.serdes import SerDes, SerDesContext


class UppercaseSerDes(SerDes[str]):
    def serialize(self, value: str, _: SerDesContext) -> str:
        return value.upper()

    def deserialize(self, data: str, _: SerDesContext) -> str:
        return data


@durable_step
def return_input(_step_context: StepContext, value: str) -> str:
    return value


@durable_with_child_context
def serdes_child(ctx: DurableContext, value: str) -> str:
    return ctx.step(return_input(value))


@durable_execution
def handler(event: Any, context: DurableContext) -> str:
    result: str = context.run_in_child_context(
        serdes_child(event),
        name="serdes-child",
        config=ChildConfig(serdes=UppercaseSerDes()),
    )
    return result
```

### SDK Version

1.5.0

### Python Version

3.14

### Is this a regression?

No

### Additional Context

Please note this might be a breaking change.

PR #550, opened 2026-07-17
fix: round-trip serdes result on first run across ops

ayushiahjolia:
*Issue #, if available:* 
https://github.com/aws/aws-durable-execution-sdk-python/issues/406
https://github.com/aws/aws-durable-execution-sdk-python/issues/544

*Description of changes:*
Deserialize every operation result before the SUCCEED checkpoint so a succeeded operation is always reconstructable and first-run matches replay. A permanent serdes failure surfaces as SerDesError (catchable
independently of the operation type); a transient failure (RetryableSerDesError) escapes the invocation for backend retry.

#### Behavior changes to note
- For wait_for_condition, the wait strategy now sees the round-tripped value.
- A result that wasn't serializable before (for virtual or large-payload child contexts) will now raise SerDesError instead of silently passing through.
- Custom serdes authors should expect deserialize to be called right after serialize on the first run. For large payloads this means one extra round-trip.
- Serdes failures are now a catchable SerDesError (permanent, no retry) or RetryableSerDesError (transient, backend retry). Previously all serdes failures raised ExecutionError (which terminated the execution without retry).
- SerDesError is not a DurableOperationError. A broad except DurableOperationError will not catch it; use except SerDesError explicitly.
- In map/parallel branches, a RetryableSerDesError escapes the batch and fails the invocation for backend retry, rather than being recorded as a permanent failed item.
- For AT_MOST_ONCE steps, a transient serdes failure after the step body succeeds surfaces as StepInterruptedError on the next invocation (the body is not re-run).
- The wire ErrorType field now uses fully-qualified class names (e.g. `aws_durable_execution_sdk_python.exceptions.StepError`) instead of bare names. Builtins remain unqualified (e.g. `ValueError`). **This is a breaking change from v1.**
- The error_type attribute on caught operation errors reflects the qualified name (e.g. `exc.error_type == "aws_durable_execution_sdk_python.exceptions.StepError"`).


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
```diff
@@ -232,7 +232,16 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
                     self.operation_identifier.operation_id,
                     self.operation_identifier.name,
                 )
-                return new_state
+                # Round-trip so the first run matches replay. A None payload is
+                # returned as-is.
+                if serialized_state is None:
+                    return None  # type: ignore[return-value]
+                return deserialize(
```
zhongkechen: I think deserialize should happen before we checkpoint SUCCEEDED. If deserialize throws, we checkpointed SUCCEEDED and then checkpoint another FAILED for the same operation. Same issue in child operation.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
yaythomas: shouldn't this also be `str | None` like the others are now?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/child.py
```diff
@@ -133,6 +133,20 @@ def check_result_status(self) -> CheckResult[T]:
         # Ready to execute (checkpoint exists or was just created)
         return CheckResult.create_is_ready_to_execute(checkpointed_result)
 
+    def _deserialize_payload(self, serialized: str | None) -> T:
```
yaythomas, line 138: nice! keeping all three child return sites identical :-)

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
ayushiahjolia: Yes my proposal is to keep it consistent across all operations in python.
1. Deserialize before the SUCCEED checkpoint - a "SUCCEEDED" op should always be reconstructable.
2. No automatic retry - the failure is usually deterministic; auto-retry poison-pills or re-runs side effects.
3. Make it catchable as a distinct SerDesException - lets users recover / lets map/parallel tolerance absorb one bad item, while an uncaught one still fails loudly.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
ayushiahjolia: Sure, can update.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
yaythomas: Is the reordering (points 1–3) coming in this PR or a follow-up? If a follow-up, please could you file a tracking issue (linking this thread, please)?

ayushiahjolia: I'm updating current PR with the fix - For step, run_in_child_context, and wait_for_condition, we'll always deserialize the result before writing the SUCCEED checkpoint, so a "succeeded" operation is guaranteed to be readable and first-run matches replay. If deserialization fails permanently, it surfaces as a catchable SerDesError; if a serdes signals a transient failure via RetryableSerDesError, the invocation fails and the backend retries. No automatic step-retry either way.

packages/aws-durable-execution-sdk-python/tests/filesystem_serdes_test.py
```diff
@@ -632,25 +632,25 @@ def test_works_with_top_level_serialize_function(self, tmp_path):
...
-        """Serialize should raise ExecutionError on failure via top-level API."""
-        from aws_durable_execution_sdk_python.exceptions import ExecutionError
+        """Serialize should raise SerDesError on failure via top-level API."""
+        from aws_durable_execution_sdk_python.exceptions import SerDesError
         from aws_durable_execution_sdk_python.serdes import serialize
 
         fs_serdes = FileSystemSerDes("/nonexistent/readonly/path")
 
-        with pytest.raises(ExecutionError, match="Serialization failed"):
+        with pytest.raises(SerDesError, match="Serialization failed"):
```
yaythomas, line 641: should this be raising RetryableSerDesError or SerDesError in the new world?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -329,6 +329,16 @@ class WaitForConditionError(DurableOperationError):
     """Raised when a wait_for_condition operation fails."""
 
 
+class SerDesError(DurableOperationError):
```
yaythomas: subclassing DurableOperationError makes every broad except DurableOperationError:` handler silently swallow serdes bugs. Is this intentional? Or was the intent that a caller should specifically opt-in to catching this one? 

> Use a distinct SerDesException (not a subclass of the ordinary operation-failed error) so catching it is intentional, a broad business except won't silently swallow it, and an uncaught SerDesException still surfaces the (usually-a-bug) failure loudly.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -378,6 +389,22 @@ def __init__(self, message: str, step_id: str | None = None):
         self.step_id = step_id
 
 
+class RetryableSerDesError(InvocationError):
```
yaythomas, line 422: child_handler correctly re-raises this with no FAIL checkpoint, but the worker then records it as an ordinary branch failure, so inside map/parallel the invocation never fails and the backend never retries, contradicting the class's contract ("fails the invocation so the backend retries")? 

The parent then checkpoints a BatchResult with that item permanently failed off a transient blip, and tolerance can fail the whole batch on it. 

A serdes author reading the docstring will expect transients to retry everywhere; inside a batch they instead become a permanently checkpointed FAILED item that no retry can ever repair.

You could handle it and let the coordinator note any retryable invocation error it observes and re-raise it before returning the batch result, so the invocation fails after the drain rather than mid-flight.

For comparison: Java's MapOperation already does this. Its aggregation loop rethrows UnrecoverableDurableExecutionException via terminateExecution before recording failed items, so invocation-class errors escape the batch; only ordinary failures become items. JS terminates the invocation on any serdes failure, so a branch serdes failure is never a failed item there either.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
```diff
@@ -228,7 +244,7 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
                     self.operation_identifier.operation_id,
                     self.operation_identifier.name,
                 )
-                return new_state
+                return return_value
```
ayushiahjolia, line 249: Currently JS and Python return the round-tripped state object directly, assuming the wait strategy treats it as read-only. Java instead returns the value re-deserialized from the checkpoint, so a mutating strategy can't affect the returned result. We can adopt the same in JS/Python as an improvement.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
yaythomas:
```
except InvocationError as e:
    if e.is_retryable():
        # infrastructure failure: escape the batch, fail the invocation
        ...
    events.put(BranchEvent.failed(executable.index, e))
```

packages/aws-durable-execution-sdk-python/tests/filesystem_serdes_test.py
ayushiahjolia: SerDesError is ExecutionError and RetryableSerDesError is InvocationError, so this is correct. However, we can rework filesystem serdes now that we have new error.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
ayushiahjolia: It should be a part of DurableExecutionError similar to java - https://github.com/aws/aws-durable-execution-sdk-java/blob/main/sdk/src/main/java/software/amazon/lambda/durable/exception/SerDesException.java

ayushiahjolia: I'll revert this change and keep it as it was.

yaythomas:
re `SerDesError` name clash raised by the bot reviewer... 

A collision here means except StepError: misses a real step failure and except SerDesError: catches a non-serdes one

could be an idea to store the namepssced class name rather than the short/bare class name? i.e `aws_durable_execution_sdk_python.exceptions.SerDesError`?

collisions become difficult... To collide, a user would have to ship a module literally named aws_durable_execution_sdk_python.exceptions containing a class of the same name.

At which point they've shadowed the SDK's own import path and have gone to a fair degree of trouble.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -329,6 +329,39 @@ class WaitForConditionError(DurableOperationError):
...
+        error_type: Type name of the error that escaped serdes; defaults to the
+            class name when no escaping error is known.
+        data: Optional serialized error payload.
+        stack_trace: Optional stack trace lines captured from the origin.
+    """
+
+    # Namespaced wire token used as the checkpoint ErrorType discriminator.
+    # Matching on this rather than the bare class name prevents a user
+    # exception coincidentally named "SerDesError" from being misidentified.
+    WIRE_ERROR_TYPE: str = "aws_durable_execution_sdk_python.exceptions.SerDesError"
```
yaythomas: `f"{__module__}.{__qualname__}"`?

Also, so only this one type gets a namespaced token and eight others get bare class names? So the ErrorType field now carries two conventions.

You could argue this is the only discriminant the SDK branches on to choose a raised type, so this bare-names for the others are fine and this one is special?

If you want to go the other way for consistency, a relatively light way of enforcing this for the others is something like this in ErrorObject.from_exception?
 
```
cls_ = type(exception)
module: str = "" if cls_.__module__ == "builtins" else f"{cls_.__module__}."
wire_type: str = f"{module}{cls_.__qualname__}"

type=wire_type"
```

This would give you ValueError rather than `builtins.ValueError`, and everything else full.y qualified.

this would also mean to change
_DURABLE_OPERATION_ERROR_REGISTRY to derived from the classes rather than typed by hand.

And raise_as_operation_error's comparison to use the same derived expression for SerDesError.

ayushiahjolia: We can do this to make it consistent, but should be a separate change.
And this change means we are accepting that python will differ in convention with other SDKs.

yaythomas: Re "differ in convention", my worry is that this change is introducing two different conventions in the same PR in the same SDK. And when looking at convention for other SDKs, doesn't Java's ErrorType values use fully qualified (Class.forName(errorType))? JS does use bare names.

And the other part I was asking about is whether it's feasiable to derive the value for WIRE_ERROR_TYPE instead of hardcoding the literal?

Which parts of these are you proposing to do as a separate change?

ayushiahjolia: yes it's feasible to derive the value of WIRE_ERROR_TYPE from module name, I'll include that change. I am proposing to do the same change in other Error types in separate PR (still in v2).

ayushiahjolia: To align with Java, I'll make all operations use fully qualified name.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -224,20 +236,22 @@ def from_dict(cls, data: MutableMapping[str, Any]) -> ErrorObject:
...
-        # error records its own type; a typed wrapper records its operation kind.
-        # For a DurableOperationError also preserve its own data and stack_trace
-        # so the inner info survives across a single operation boundary.
-        if isinstance(exception, DurableOperationError):
+        # The wire ErrorType is the fully-qualified class name, with builtins
+        # left unqualified for brevity.
+        # For errors that carry wire fields (DurableOperationError, SerDesError)
+        # also preserve their data and stack_trace so the inner info survives
+        # across a single operation boundary.
+        wire_type: str = _qualified_error_type(exception)
```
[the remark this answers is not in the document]
ayushiahjolia: This change ships in v2 which is a known breaking release - there are no v1 checkpoints to be backward-compatible with. The qualified wire format is intentional and aligns with Java's getClass().getName().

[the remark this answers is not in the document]
ayushiahjolia: This ships in v2 which is a breaking release. There are no v1 checkpoints to maintain backward compatibility with. The qualified wire format is the v2 contract going forward.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/step.py
```diff
@@ -251,17 +269,28 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
...
-                operation_id=self.operation_identifier.operation_id,
-                durable_execution_arn=self.state.durable_execution_arn,
+            return return_value
+        except RetryableSerDesError:
+            # Transient serdes failure: fail the invocation for backend retry,
+            # bypassing the step retry strategy. This narrow catch relies on the
+            # serdes wrappers raising only RetryableSerDesError or SerDesError;
+            # any other retryable InvocationError would fall through to the step
+            # retry strategy below.
+            raise
```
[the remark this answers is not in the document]
ayushiahjolia: For AT_MOST_ONCE, the step body already ran successfully - only serialization failed transiently. We can't re-run the body (side effects) and can't persist the result (serialization failed). The START checkpoint remains, so the next invocation surfaces StepInterruptedError, which is the correct signal: "your step may have completed, check external state." This matches the existing interruption semantics. Added a comment documenting this interaction.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/operation/wait_for_condition.py
```diff
@@ -193,16 +208,16 @@ def execute(self, checkpointed_result: CheckpointedResult) -> T:
             )
             new_state = wrapped_user_func(current_state, check_context)
 
+            serialized_state = self._serialize(new_state)
+
+            # Round-trip before the wait strategy so it evaluates the
+            # reconstructable state, matching replay. None round-trips as None;
+            # a permanent failure fails before any checkpoint.
+            round_tripped_state: T = self._deserialize(serialized_state)
```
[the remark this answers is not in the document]
ayushiahjolia: The wait strategy doesn't produce or modify state - it only returns a continue/stop decision.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -360,17 +392,22 @@ class CallbackSubmitterError(CallbackError):
...
+    for cls in [
+        StepError,
+        InvokeError,
+        ChildContextError,
+        WaitForConditionError,
+        CallbackError,
+        CallbackExternalError,
+        CallbackTimeoutError,
+        CallbackSubmitterError,
+    ]
```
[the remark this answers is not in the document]
ayushiahjolia: Acknowledged as a v2 breaking change. Customers must complete or terminate in-flight v1 executions before upgrading. We write qualified names going forward; v1 short names are not supported on replay. This will be documented in the v1-to-v2 migration guide.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -333,6 +335,36 @@ class WaitForConditionError(DurableOperationError):
...
 
+class SerDesError(DurableExecutionsError):
+    """Raised when serializing or deserializing an operation result fails.
+
+    Signals a permanent failure; use :class:`RetryableSerDesError` for a
+    transient one that should retry.
+
+    Attributes:
+        message: Human-readable failure message.
+        error_type: Type name of the error that escaped serdes; defaults to the
```
yaythomas: it is not the escaping error's type (no call site passes one), and the default is the qualified name rather than the bare class name.

> error_type: Fully-qualified name of this class. A serdes failure records
            its own type rather than the escaping error's, so replay
            reconstructs it as :class:`SerDesError`.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
yaythomas: also superseded: 

> error_type: Type name of the error that escaped the operation (the
            original/inner error, e.g. ``"ValueError"``), not the operation
            class. Defaults to this class's fully-qualified name only when no
            escaping error is known.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -382,6 +419,26 @@ def __init__(self, message: str, step_id: str | None = None):
...
+    """Signal a transient SerDes failure that should retry the invocation.
+
+    Raised by a SerDes for a transient failure (e.g. an offloading serdes whose
+    network call timed out). It fails the invocation so the backend retries,
+    rather than surfacing to user code. Use :class:`SerDesError` for a permanent
+    failure.
+
+    In a map or parallel branch it escapes the batch rather than becoming a
+    failed item, so the invocation still fails and retries.
+    """
```
yaythomas, line 437: > A step configured ``AT_MOST_ONCE_PER_RETRY`` has already persisted its START
    checkpoint, so the retried invocation treats the step as interrupted and
    applies the step retry strategy. A strategy that declines or exhausts its
    retries fails the step and surfaces :class:`StepError` to the caller.

leaves your response here https://github.com/aws/aws-durable-execution-sdk-python/pull/550#discussion_r3768128152 in the code for posterity :-)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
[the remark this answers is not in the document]
yaythomas: Both named examples are impossible: StepError and CallbackExternalError don't exist in any released version, they were introduced in #540 an d aren't released yet.

CallbackError did exist in v1 (as an ExecutionError subclass), so a v1 checkpoint can carry the bare token "CallbackError". Under the qualified-only registry that falls back to base DurableOperationError, and wait_for_callback's isinstance(inner, CallbackError) then misses, surfacing ChildContextError instead of passing the callback error through.

That's the divergence described, just for a different token than the one cited.

The dominant v1 token can't be fixed by aliases anyway. v1 recorded "CallableRuntimeError" at wrapping contexts, and v2 deletes that class outright.

Is it worth registering the bare aliases for the v1-era names that still exist as classes (CallbackError, arguably SerDesError) to cover an execution that started pre-#540 and is still replaying after the function upgrades to v2?

Impact of not doing it: an execution that started pre-#540 and replays post-upgrade surfaces ChildContextError where a v2-native execution surfaces CallbackError. It isn't a v1 regression, since v1 surfaced CallableRuntimeError for that same checkpoint (child.py wrapped ExecutionError subclasses via to_callable_runtime_error), and that class is deleted in v2, so the original behaviour is unreproducible regardless. The window closes as pre-upgrade executions drain.

So I think you can make the argument that the bot review comment is over-zealous here.

yaythomas (changes requested):
just some docstrings got left behind

yaythomas:
is this wait_for_callback err related (this PR touches that execution path)?

 ❌ otel-invocation-10 (Otel10WaitForCallback)
   OpenTelemetry: Expected at least 10 span(s), found 9
   OpenTelemetry: span_assertions[5].expect.before matched no spans
   OpenTelemetry: span_assertions[6].select matched no spans

ayushiahjolia:
> is this wait_for_callback err related (this PR touches that execution path)?
> 
> ❌ otel-invocation-10 (Otel10WaitForCallback) OpenTelemetry: Expected at least 10 span(s), found 9 OpenTelemetry: span_assertions[5].expect.before matched no spans OpenTelemetry: span_assertions[6].select matched no spans

No, that was transient error in test, passed on rerun.

yaythomas (approved):
for follow-up:

initial_state round-trip. wait_for_condition now round-trips initial_state before the first check, where JS and Java both pass it raw. I think that's the better behaviour, since the check then sees one consistent shape on every poll, but it is a deliberate divergence and it newly requires initial_state to be serializable by the configured serdes. 

WaitForConditionConfig's docstring is currently one line with no attribute docs, so please could we state the requirement there, and mention it in the v2 release notes as a migration note? Worth also filing tracking issues on JS and Java to match, so the divergence is tracked intent rather than drift.

PR #555, opened 2026-07-20
fix: keep empty string payloads when serializing

### LINKED ISSUE #554: Empty serialized payloads are dropped, causing "" to replay as None
### Expected behavior

An empty serialized string is a valid payload and should remain distinct from
an absent payload:

```text
serialize(result) == ""  ->  replay deserializes ""  ->  original result
serialize(result) is None  ->  replay returns None
```

Operations using a custom SerDes should therefore return the same value on the
first run and replay when `SerDes.serialize()` returns `""`.

### Actual behavior

`OperationUpdate.to_dict()` only includes `Payload` when the value is truthy:

```python
if self.payload:
    result["Payload"] = self.payload
```

Consequently, `payload=""` is omitted from the
`CheckpointDurableExecution` request. The backend records a successful
operation without a result, and replay reconstructs that missing result as
`None`.

For example, the built-in `PassThroughSerDes` exposes the problem when an
operation returns an empty string:

```text
First run: ""
Replay:    None
```

The underlying wire behavior can be reproduced directly:

```python
update = OperationUpdate.create_step_succeed(
    OperationIdentifier(
        "step-1",
        OperationSubType.STEP,
        None,
        "empty-result",
    ),
    payload="",
)

assert update.payload == ""
assert "Payload" in update.to_dict()  # currently fails
```

### Affected operations

This is shared `OperationUpdate` serialization behavior and can affect any
operation that checkpoints a serialized result, including:

- `step`
- `run_in_child_context` (and map/parallel paths using child contexts)
- `wait_for_condition`

### Suggested fix

Preserve every payload except `None`:

```python
if self.payload is not None:
    result["Payload"] = self.payload
```

Add coverage that:

1. `OperationUpdate.to_dict()` includes `Payload: ""`.
2. An operation using `PassThroughSerDes` with an empty-string result returns
   `""` on both first run and replay.

This issue predates PR #550 but was identified while reviewing that PR's
first-run/replay SerDes consistency changes.

nvasiu:
*Issue #, if available:*

https://github.com/aws/aws-durable-execution-sdk-python/issues/554

*Description of changes:*

Fixed : `OperationUpdate.to_dict()` dropped empty string payloads, so on replay the payload would be `None`.

Also added 3 tests:
- Test `OperationUpdate.to_dict()` returns empty string payloads.
- Test an operation using `PassThroughSerdes` returns an empty string payload on first run.
  - \+ Test the same for replay.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
@@ -470,7 +470,7 @@ class OperationUpdate:
             result["Name"] = self.name
         if self.sub_type:
             result["SubType"] = self.sub_type.value
-        if self.payload:
+        if self.payload is not None:
             result["Payload"] = self.payload
         if self.error:
             result["Error"] = self.error.to_dict()
--- a/packages/aws-durable-execution-sdk-python/tests/lambda_service_test.py
+++ b/packages/aws-durable-execution-sdk-python/tests/lambda_service_test.py
@@ -743,6 +743,19 @@ def test_operation_update_create_step_succeed():
     assert update.sub_type is OperationSubType.STEP
 
 
+def test_operation_update_to_dict_includes_empty_payload():
+    """Test OperationUpdate.to_dict includes an empty-string payload."""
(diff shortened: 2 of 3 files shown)


yaythomas (approved):
:shipit: 

nice find! 🙇

Issue #560, opened 2026-07-21
[Bug]: Invocation plugin hooks report execution start time as per-invocation start time on replay


nvasiu:
### Expected Behavior

Following the JS SDK as baseline, `InvocationStartInfo` should expose the execution's start timestamp.

### Actual Behavior

`InvocationStartInfo.start_time` is currently set to local `now()` for initial invocations. 

For replay invocations, `InvocationStartInfo.start_time` is set to the execution's start time.

### Steps to Reproduce

1. Register a plugin that records invocation start/end times.
2. Write a durable handler that suspends and resumes (e.g. a wait).
3. Run the handler and compare the logged values for both invocations.
4. The first invocation's duration will be reasonable, but the replay invocation's duration will be much longer, because it's start time is when the execution began, not when the current invocation began.

### SDK Version

1.7.0

### Python Version

3.14

### Is this a regression?

No

PR #561, opened 2026-07-21
feat(plugin)!: add execution_start_time to invocation info

nvasiu:
*Issue #, if available:*

https://github.com/aws/aws-durable-execution-sdk-python/issues/560

*Description of changes:*

This PR includes breaking changes to the plugin interface (which is still experimental). 
These changes follow the implementation of the JS SDK plugin interface, which the Python SDK uses as a baseline.

`InvocationInfo.start_time` and `InvocationEndInfo.end_time` are both removed:
- `InvocationInfo.start_time` was set to the local clock time for new invocations. `InvocationEndInfo.end_time` was always set to the local clock time. Passing this value is not useful because plugins can simply get local clock time themselves.
- `InvocationInfo.start_time` was set to the execution start time for replay invocations, which is misleading because the variable name implied that it's value was the invocation's start time.

`InvocationInfo.execution_start_time` added:
- `execution_start_time` is present in both `InvocationStartInfo` and `InvocationEndInfo`, just like in the JS SDK plugin interface.

`plugin_test.py`
- Updated tests to use `execution_start_time`.

*This PR also updates the OTel plugin to work with the above breaking changes to the plugin interface.*

- Replaced `start_time` with `execution_start_time` when deriving trace ID.
- Removed usage of `end_time`. Invocation end span timestamps now default to using the current wall-clock time, same as in the JS OTel plugin.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
@@ -161,8 +161,8 @@ class UserFunctionEndInfo(OperationInfo):
 class InvocationInfo:
     request_id: str | None
     execution_arn: str | None
-    start_time: datetime.datetime | None
     is_first_invocation: bool
+    execution_start_time: datetime.datetime | None = None
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


zhongkechen (approved):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -172,9 +172,8 @@ class InvocationStartInfo(InvocationInfo):
 
 @dataclass(frozen=True)
 class InvocationEndInfo(InvocationInfo):
-    status: InvocationStatus
-    end_time: datetime.datetime | None
-    error: ErrorObject | None
+    status: InvocationStatus | None = None
```
zhongkechen, line 175: Can we make this a mandatory field?

nvasiu: https://github.com/aws/aws-durable-execution-sdk-python/pull/567

Issue #313, opened 2026-02-20
[Feature]: add specific err for callbacks


yaythomas:
### What would you like?

https://github.com/aws/aws-durable-execution-sdk-js/pull/441

Add CallbackTimeoutError and CallbackSubmitterError to provide clear, unambiguous error handling for callback operations. Replace errorClass with more flexible errorMapper in ChildConfig.

Changes:
- Add CallbackTimeoutError for callback timeout scenarios
- Add CallbackSubmitterError for submitter failures in waitForCallback
- Add errorMapper to ChildConfig for flexible error type mapping
- Remove errorClass from ChildConfig (replaced by errorMapper)
- Update callback.ts to throw CallbackTimeoutError for TIMED_OUT status
- Update waitForCallback to use errorMapper for consistent error types
- Update all promise combinators to use errorMapper
- Update fromErrorObject to reconstruct new error types
- Export CallbackTimeoutError and CallbackSubmitterError
- Update all tests to expect correct error types

Error types by operation:
- createCallback timeout -> CallbackTimeoutError
- createCallback failure -> CallbackError
- waitForCallback timeout -> CallbackTimeoutError
- waitForCallback failure -> CallbackError
- waitForCallback submitter failure -> CallbackSubmitterError

### Is this a breaking change?

No

### Does this require an RFC?

No

yaythomas:
Absorbed into the #120 v2 error hierarchy: `CallbackTimeoutError`/`CallbackSubmitterError` land as subclasses of `CallbackError` (graded, not flat siblings). The `errorMapper` portion is superseded — Python translates at the consumer call site (`wait_for_callback` try/except) since there is only one reuse site and no combinators. See #120 for the full design.

zhongkechen:
Explicitly absorbed by [#120](https://github.com/aws/aws-durable-execution-sdk-python/issues/120); close as superseded.

ayushiahjolia:
Reopened this to track CallbackError fix seperately.

PR #563, opened 2026-07-21
feat(errors): Add graded callback error types

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-python/issues/313
https://github.com/aws/aws-durable-execution-sdk-python/issues/564

*Description of changes:*
```
DurableOperationError
-> CallbackError            # base; internal-invariant case; keeps callback_id
---> CallbackExternalError  # callback FAILED (SendDurableExecutionCallbackFailure)
---> CallbackTimeoutError   # callback TIMED_OUT (timeout / heartbeat)
---> CallbackSubmitterError # submitter step failed
```

Before this change, callback failures collapsed into a generic error: a failed callback surfaced as `ChildContextError` and callers couldn't distinguish an external failure from a timeout or a submitter failure. Now each surfaces as a specific, catchable type.

### What changed

- `CallbackError` now subclasses `DurableOperationError` (plain subclass, no custom constructor - it inherits `message`, `error_type`, `data`, `stack_trace`). Dropped the unused `TerminationReason.CALLBACK_ERROR`.

- `Callback.result()` grades terminal failures: `TIMED_OUT` → `CallbackTimeoutError`; other terminal (external `FAILED`, cancelled, stopped) → `CallbackExternalError`; missing-operation invariant → base `CallbackError`. Preserves `message` / `data` / `stack_trace`.

- `wait_for_callback` translates at its own call site: the `CallbackError` family passes through unchanged; a submitter `StepError` becomes `CallbackSubmitterError`; anything else re-raises as the generic `ChildContextError`. Replay-safe: `run_in_child_context` re-raises from the FAILED checkpoint on replay, so the `except` fires identically on first run and replay.

- `callback_id` is not carried on callback errors. It isn't an error wire field, so it can't survive the `wait_for_callback` child-context boundary, and once that context is terminal (`replay_children=false`) its child operations are dropped from later invocations' `InitialExecutionState` - the CALLBACK op, and any id derived from it, is absent on replay. Rather than expose a field that is populated on first run and `None` after replay, we omit it; the caller already holds it from `create_callback()` / the submitter argument.

- No synthetic `__cause__` on terminal failures. The graded errors don't chain a reconstructed cause - it would only duplicate `message` / `data` / `stack_trace` already on the error, and the original external `ErrorType` remains on the CALLBACK operation's details for anyone who needs it. 

- Registered the four callback types in `_DURABLE_OPERATION_ERROR_REGISTRY` so they reconstruct on replay (including when `wait_for_callback` is nested in another child context); exported them from the package root.

### Breaking change

`CallbackError` is no longer an `ExecutionError`. Execution outcome is unchanged: the top-level `except Exception` still returns FAILED with no retry.

### Testing

- Unit tests
- Core integration tests
- Example integration tests

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
@@ -5,7 +5,15 @@ import logging
 from contextlib import contextmanager
 from dataclasses import dataclass
 from threading import Lock
-from typing import TYPE_CHECKING, Any, Concatenate, Generic, ParamSpec, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Concatenate,
+    Generic,
+    NoReturn,
+    ParamSpec,
+    TypeVar,
+)
 
 from aws_durable_execution_sdk_python.config import (
     BatchedInput,
@@ -21,6 +29,11 @@ from aws_durable_execution_sdk_python.config import (
 )
 from aws_durable_execution_sdk_python.exceptions import (
     CallbackError,
+    CallbackExternalError,
+    CallbackSubmitterError,
+    CallbackTimeoutError,
+    ChildContextError,
+    StepError,
     SuspendExecution,
     ValidationError,
 )
@@ -69,6 +82,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
 
     from aws_durable_execution_sdk_python.concurrency.models import BatchResult
+    from aws_durable_execution_sdk_python.lambda_service import ErrorObject
     from aws_durable_execution_sdk_python.state import CheckpointedResult
     from aws_durable_execution_sdk_python.types import LambdaContext
     from aws_durable_execution_sdk_python.waits import WaitForConditionConfig
@@ -252,8 +266,10 @@ class Callback(Generic[T], CallbackProtocol[T]):  # noqa: PYI059
         )
 
         if not checkpointed_result.is_existent():
+            # Should never happen (create_callback already checkpointed this op).
+            # Not external/timeout/submitter, so raise the base CallbackError.
             msg = "Callback operation must exist"
-            raise CallbackError(message=msg, callback_id=self.callback_id)
+            raise CallbackError(message=msg)
 
         if (
             checkpointed_result.is_failed()
@@ -261,12 +277,7 @@ class Callback(Generic[T], CallbackProtocol[T]):  # noqa: PYI059
             or checkpointed_result.is_timed_out()
             or checkpointed_result.is_stopped()
         ):
-            msg = (
-                checkpointed_result.error.message
-                if checkpointed_result.error and checkpointed_result.error.message
-                else "Callback failed"
-            )
-            raise CallbackError(message=msg, callback_id=self.callback_id)
+            self._raise_terminal_failure(checkpointed_result)
 
         if checkpointed_result.is_succeeded():
             if checkpointed_result.result is None:
@@ -283,6 +294,29 @@ class Callback(Generic[T], CallbackProtocol[T]):  # noqa: PYI059
         # therefore we should wait
         msg = "Callback result not received yet. Suspending execution while waiting for result."
         raise SuspendExecution(msg)
+
+    def _raise_terminal_failure(
+        self, checkpointed_result: CheckpointedResult
+    ) -> NoReturn:
+        """Raise the graded error for a terminal callback failure.
+
+        TIMED_OUT -> CallbackTimeoutError; any other non-succeeded terminal
+        (external FAILED, cancelled, stopped) -> CallbackExternalError.
+        Carries the message/data/stack_trace; the external error's own type
+        stays on the callback operation's details.
+        """
+        error: ErrorObject | None = checkpointed_result.error
+        is_timeout: bool = checkpointed_result.is_timed_out()
+        error_cls: type[CallbackError] = (
+            CallbackTimeoutError if is_timeout else CallbackExternalError
+        )
+        default_message: str = "Callback timed out" if is_timeout else "Callback failed"
+        message: str = error.message if error and error.message else default_message
+        raise error_cls(
+            message=message,
+            data=error.data if error else None,
+            stack_trace=error.stack_trace if error else None,
+        )
 
 
 class DurableContext(DurableContextProtocol):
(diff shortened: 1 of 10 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -338,6 +329,50 @@ class WaitForConditionError(DurableOperationError):
...
+        stack_trace: list[str] | None = None,
+        callback_id: str | None = None,
+    ) -> None:
+        super().__init__(
+            message=message,
+            error_type=error_type,
+            data=data,
+            stack_trace=stack_trace,
+        )
+        self.callback_id: str | None = callback_id
```
zhongkechen: callback_id is lost when replay. It's not a new issue in this PR. I'll create a new github issue: https://github.com/aws/aws-durable-execution-sdk-python/issues/564

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -284,6 +287,31 @@ def result(self) -> T | None:
         msg = "Callback result not received yet. Suspending execution while waiting for result."
         raise SuspendExecution(msg)
 
+    def _raise_terminal_failure(self, checkpointed_result: CheckpointedResult) -> None:
```
yaythomas: is intent here `-> NoReturn` ?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
zhongkechen: You could include a fix in this PR if you'd prefer.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -825,11 +853,28 @@ def wait_for_callback(
...
+        except ChildContextError as e:
+            inner = e.__cause__
+            # Callback failures (external/timeout/base) pass through.
+            if isinstance(inner, CallbackError):
+                raise inner
+            # A failed submitter step surfaces as CallbackSubmitterError, keeping
+            # the step's data/stack_trace (original StepError on __cause__).
+            if isinstance(inner, StepError):
+                raise CallbackSubmitterError(
+                    str(inner), data=inner.data, stack_trace=inner.stack_trace
```
yaythomas: `str(inner)` will become `"None"` on `None`.

for comparison, JS does CallbackSubmitterError constructor `(message || "Callback submitter failed")`

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
```diff
@@ -825,11 +853,28 @@ def wait_for_callback(
...
+        # the call site (not inside the child body) so it re-fires identically on
+        # replay, where run_in_child_context re-raises from the FAILED checkpoint.
+        try:
+            return self.run_in_child_context(
+                wait_in_child_context,
+                step_name,
+                ChildConfig(sub_type=OperationSubType.WAIT_FOR_CALLBACK),
+            )
+        except ChildContextError as e:
+            inner = e.__cause__
```
yaythomas: nit: inner: BaseException | None = e.__cause__

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
yaythomas: +1

seems it gets lost on the ErrorObject round-trip at the child boundary for wait for callback. the direct create_callback()/result() path retains callback_id.

The original external ErrorType is also only on the dropped __cause__. the CALLBACK op already checkpoints CallbackDetails.callback_id and the external error, so wait_for_callback could re-hydrate both from state deterministically.

that said, we could also remove `callback_id` from the err obj entirely? the caller knows that the id is (create_callback() returns it, the submitter receives it as an argument). 

re on __cause__, in JS the reconstructed cause is a synthetic duplicate of the wire fields (fromErrorObject builds Error(name=ErrorType, message=ErrorMessage)) with the same message, name set to the outer error's own recorded ErrorType, same stack, no additional information survives the boundary there either. https://github.com/aws/aws-durable-execution-sdk-js/blob/b2e3d985a32accb67fc533d8941da0f17491ba00/packages/aws-durable-execution-sdk-js/src/errors/durable-error/durable-error.ts#L32-L34

Python's message/data/stack_trace already round-trip as fields, and the full external error stays on the CALLBACK operation's details for anyone who needs the original ErrorType. So we could leave __cause__==None as-is, rather than make a synthetic duplicate?

Java SDK err obj does have the callback_id, so there is definitely a strong case for _not_ removing it too!

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
ayushiahjolia: yes, since this method raises failure in either case, NoReturn is a better choice.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
ayushiahjolia: JS solution is clean, let's follow the same in python.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/context.py
ayushiahjolia: I'll make this change along with other changes.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
ayushiahjolia: Thanks both.

1. Keeping `callback_id` and fixing it properly here rather than deferring.

`callback_id` is meaningful correlation data and we already expose it for direct create_callback()/result() calls - I'd like it consistent for wait_for_callback too. Since it isn't an error wire field, I re-derive it from the checkpointed callback operation at the wait_for_callback boundary, so it's populated identically on first run and replay.

This is a deliberate divergence from the JS SDK, which omits `callback_id` from its callback errors entirely and relies on the caller holding the id from createCallback()/the submitter. I'm following the Java behavior because the correlation/observability value is worth surfacing and it keeps the error self-contained.

2. Dropping `__cause__` for both direct and wrapped calls.

The cause we'd build just duplicates fields already on the error (message/data/stack_trace), and the one extra bit - the original external ErrorType - stays on the CALLBACK operation's details. JS's reconstructed cause carries nothing extra across the boundary either, so there's no value in manufacturing one.

Let me know if we want to change any other SDK's behavior for parity.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -420,6 +420,23 @@ def has_prior_operations(self) -> bool:
                 for op in self._operations.values()
             )
 
+    def get_child_callback_id(self, parent_operation_id: str) -> str | None:
```
yaythomas: After a child context is terminal(replay_children=false), its child operations are not included in subsequent invocations' InitialExecutionState.

i.e the callback_id will still disappear on replay. 

So we wouldn't have that callback id available, unless we make it cross the wire. 

The test for this mocks LambdaClient and hand-crafts the InitialExecutionState dict, directly seeding a CALLBACK op under a FAILED context. So the new `test_failed_callback_rehydrates_callback_id_on_replay` test encodes the desired behaviour rather than any runtime's behavior.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
ayushiahjolia: Discussed offline, we are removing callback_id.

yaythomas (approved):
:shipit:

PR #568, opened 2026-07-22
fix(otel): preserve retried operation start spans

zhongkechen:
## Summary
- correct `OperationStartInfo.is_replayed` for repeated START checkpoints by checking whether the logical operation existed before the checkpoint
- keep the plugin interface unchanged; no continuation-specific field is added
- let the OTel plugin emit retry and polling continuation spans with a fresh span ID linked to the deterministic initial operation span
- cover initial starts, repeated starts, batched terminal responses, and OTel span links

## Root cause
Retried STEP and wait-for-condition operations checkpoint START more than once for the same operation ID. The SDK hardcoded `OperationStartInfo.is_replayed=False` for every START, so the OTel plugin reused the deterministic operation span ID. X-Ray collapsed the earlier STARTED span into the later terminal span, causing cases 3 and 9 in aws/aws-durable-execution-conformance-tests#12 to fail.

## Solution
`ExecutionState` now supplies the operation snapshot from before the checkpoint to `PluginExecutor`. A START is reported with `is_replayed=True` when that logical operation already existed. OTel already treats replayed operation starts as existing spans, generating a random continuation span ID and linking it to the deterministic initial span.

## Verification
- `hatch run dev-core:python -m pytest packages/aws-durable-execution-sdk-python/tests/plugin_test.py packages/aws-durable-execution-sdk-python/tests/state_test.py -q` (196 passed)
- `hatch run dev-otel:python -m pytest packages/aws-durable-execution-sdk-python-otel/tests/test_plugin.py -q` (23 passed)
- `hatch run test:all` (2934 passed, 2 skipped)
- `hatch run dev-core:typecheck`
- `hatch run dev-otel:typecheck`
- `hatch fmt --check` for core and OTel

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
@@ -375,14 +375,18 @@ class PluginExecutor:
         )
 
     def on_operation_action(
-        self, update: OperationUpdate, operation: Operation | None = None
+        self,
+        update: OperationUpdate,
+        operation: Operation | None = None,
+        previous_operation: Operation | None = None,
     ):
         """Execute any registered plugins for a given operation when an update is checkpointed
 
         Args:
             update: the operation update that is checkpointed
+            operation: the operation after the checkpoint
+            previous_operation: the operation before the checkpoint
         """
-        # todo: this could be called more than once for step when it's retried
         if update.action is OperationAction.START:
             # we handle only START action here because on_operation_update may not be able to see a STARTED update
             # when START is checkpointed in batch with terminal status updates.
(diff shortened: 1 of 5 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/plugin.py
```diff
@@ -394,7 +398,7 @@ def on_operation_action(
                     name=update.name,
                     parent_id=update.parent_id,
                     start_time=operation.start_timestamp if operation else None,
-                    is_replayed=False,
+                    is_replayed=previous_operation is not None,
```
SilanHe, line 401: isn't this only triggered once anyway since on_operation_action is only called when after we persist the operationAction.START? not sure what this code really does

zhongkechen: For Step, START is checkpointed not only when it's initially started, but also when it's retried.

SilanHe (approved):

Issue #565, opened 2026-07-22
[Bug]: [Bug]: ParamValidationError: Invalid length for parameter CheckpointToken during context.wait()


Lmex89vexi:
### Expected Behavior

Description
When suspending an execution using context.wait(), the background state-syncing thread crashes with a botocore.exceptions.ParamValidationError.

The lambda_service.py client appears to be passing an empty string ("") or invalid zero-length value for the CheckpointToken parameter during the checkpoint_durable_execution API call. Boto3 validation requires this parameter to have a minimum length of 1 if it is included in the payload.

### Actual Behavior

Actual Behavior
The checkpoint fails immediately during Boto3 parameter validation before the network request is made, crashing the background thread.

Environment Details
SDK Version: 1.5.0

Python Version: 3.13

Execution Environment: AWS Lambda

OS: Amazon Linux 2023 (Lambda standard runtime)

### Steps to Reproduce

Steps to Reproduce
Initialize a durable execution inside an AWS Lambda function running the python3.13 runtime.

Invoke context.wait(Duration.from_seconds(35)) (or similar duration).

The SDK's background checkpointing process triggers lambda_service.checkpoint().

Observe the ParamValidationError in CloudWatch logs and the resulting BackgroundThreadError.

Expected Behavior
The SDK should successfully checkpoint the execution state. If a CheckpointToken is not yet available or is empty, the SDK should omit the parameter from the checkpoint_durable_execution kwargs rather than passing a zero-length string.

### SDK Version

1.5.0

### Python Version

3.13

### Is this a regression?

No

zhongkechen:
Thanks for reporting this. We traced the checkpoint-token lifecycle and this does appear to be a valid SDK bug, but the proposed fix should be adjusted.

`CheckpointToken` is a required input to `CheckpointDurableExecution` with a minimum length of 1. Omitting it from a subsequent request would therefore fail botocore validation with a missing-required-parameter error.

The backend can legitimately return an empty/null response `CheckpointToken` when that checkpoint completes the execution, because there is no next checkpoint token. The Python SDK currently mishandles that terminal response:

1. `CheckpointOutput.from_dict()` maps an absent response token to `""`.
2. The checkpoint batching loop unconditionally installs the response token as its current token.
3. If another queued or later checkpoint is processed, the SDK sends `CheckpointToken=""`, producing the reported `ParamValidationError` before a request reaches the service.

[...]

Issue #43, opened 2025-10-15
Standardize Handling of None/Null CheckpointToken


yaythomas:
## Context
**GitHub Reference**: [lambda_service.py#L833](https://github.com/aws/aws-durable-execution-sdk-python/blob/0307cf96e22a0bdbfdb7539f2780cde19f44a55d/src/aws_durable_execution_sdk_python/lambda_service.py#L833)

Standardize how None/null `CheckpointToken` values are handled throughout the system. Align with TypeScript.

## Technical Requirements
- Define consistent behavior for null checkpoint tokens
- Update all checkpoint token handling code
- Ensure proper validation and error messages
- Handle edge cases in token lifecycle

## Implementation Details
- **File**: `src/aws_durable_execution_sdk_python/lambda_service.py`
- **Line**: Around line 833
- **Validation**: Add null checks and appropriate defaults

```python
def handle_checkpoint_token(self, token: str | None) -> str:
    """Standardize checkpoint token handling."""
    if token is None:
        # Define standard behavior for null tokens
        logger.warning("Checkpoint token is None, using empty string")
        return ""

        # OR?
        # raise FatalError(f"Checkpoint token must be string, got {type(token)}")
    
    return token.strip()  # Normalize whitespace
```

## Acceptance Criteria
- [ ] Align with TypeScript on behavior
- [ ] Consistent null token handling across codebase
- [ ] Clear validation and error messages
- [ ] Edge cases properly handled
- [ ] Behavior is documented
- [ ] Tests cover null token scenarios

**Priority**: Low  
**Estimated Effort**: 0.5 day

PR #571, opened 2026-07-22
fix: handle missing checkpoint token from service

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-python/issues/43
https://github.com/aws/aws-durable-execution-sdk-python/issues/565

*Description of changes:*
Standardizes how the SDK handles a missing/empty `CheckpointToken` from the Durable Functions API, and makes the checkpoint loop treat an absent response token as a first-class terminal signal. Previously the token was defaulted to an empty string and could be forwarded to `checkpoint_durable_execution`, where boto3 rejects a zero-length value with a `ParamValidationError` — crashing the background checkpoint thread.

Background - 
The service omits `CheckpointToken` from a checkpoint response in exactly one case: the terminal checkpoint that ends the execution (an `EXECUTION` op with action `SUCCEED`/`FAIL`). Every non-terminal checkpoint — including empty/no-op ones — returns a fresh token, and the invocation event always carries a seed token. So an absent token legitimately means "the execution is over"; anywhere else it is an invalid response, and an empty token must never be sent back.

| Situation | Behavior |
| --- | --- |
| Non-terminal checkpoint, token present | rotate to it (unchanged) |
| Execution end, small result | SDK checkpoints no terminal op; backend finalizes from the invocation response — no null token seen (unchanged) |
| Execution end, large result / failure (SDK checkpoints the terminal op) | recognized as completion: loop stops, queued orphan work settled, no further request issued |
| Non-terminal checkpoint returns no token | invalid response → retryable `CheckpointError` → re-drive |
| Empty/`None` token about to be sent (e.g. malformed seed) | retryable `CheckpointError` / `GetExecutionStateError`; no boto crash |
| `parallel`/`map` early completion (`min_successful`, failure tolerance) with branches still running | racing branch checkpoints settled as orphaned; later ones rejected fast; nothing sent post-completion |
| Nested concurrency + timed waits, orphaned after completion | resubmitter stops cleanly, no timer-thread crash |

Every token anomaly resolves to a retryable error that re-drives with a fresh token; the one legitimate absent token (execution completion) is handled silently, and an empty token is never transmitted to the service.

Testing -  Added unit tests

NOTE: Intentionally diverges from the JS/Java SDKs; they should be updated to match (guard empty-token sends, fail on non-terminal null).

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
@@ -16,6 +16,7 @@ from typing import TYPE_CHECKING, Callable, NoReturn
 
 from aws_durable_execution_sdk_python.exceptions import (
     BackgroundThreadError,
+    CheckpointError,
     DurableExecutionsError,
     DurableOperationError,
     GetExecutionStateError,
@@ -37,7 +38,7 @@ from aws_durable_execution_sdk_python.lambda_service import (
 from aws_durable_execution_sdk_python.plugin import (
     PluginExecutor,
 )
-from aws_durable_execution_sdk_python.threading import CompletionEvent, OrderedLock
+from aws_durable_execution_sdk_python.threading import CompletionEvent
 
 
 if TYPE_CHECKING:
@@ -282,7 +283,6 @@ class ExecutionState:
         self._operations: dict[str, Operation] = dict(operations)
         self._service_client: DurableServiceClient = service_client
         self._plugin_executor: PluginExecutor = plugin_executor
-        self._ordered_checkpoint_lock: OrderedLock = OrderedLock()
         self._operations_lock: Lock = Lock()
 
         # Checkpoint batching configuration
@@ -295,6 +295,12 @@ class ExecutionState:
         self._overflow_queue: queue.Queue[QueuedOperation] = queue.Queue()
         self._checkpointing_stopped: threading.Event = threading.Event()
         self._checkpointing_failed: CompletionEvent = CompletionEvent()
+        # Set once the service confirms the execution has completed (a checkpoint
+        # response with no token). No further checkpoint can succeed afterward.
+        self._execution_completed: threading.Event = threading.Event()
+        # Serializes the execution-completed check with enqueueing so a checkpoint
+        # is never enqueued after _settle_after_execution_completed drains the queue.
+        self._completion_lock: Lock = Lock()
 
         # Concurrency management for parallel operations: parent_id -> {child_operation_ids}
         self._parent_to_children: dict[str, set[str]] = {}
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
zhongkechen: When we call `_checkpoint_queue.put` here, the condition `self._execution_completed.is_set()` might have been changed by another thread since previously checked.

ayushiahjolia: yes we need a lock here, improving this.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
```diff
@@ -766,16 +811,43 @@ def checkpoint_batches_forever(self) -> None:
...
+                        update.operation_type is OperationType.EXECUTION
+                        and (
+                            update.action is OperationAction.SUCCEED
+                            or update.action is OperationAction.FAIL
+                        )
+                        for update in updates
+                    ):
+                        execution_completed = True
+                    else:
+                        raise CheckpointError(
```
yaythomas, line 826: If you hit a quota limit (e.g for operations size quotas), gets a retryable CheckpointError claiming the service response was invalid, when the execution is already FAILED with the real limit error. 

The retryable classification and the message would both mislead during debugging.

maybe the keep the invalid-response error only for empty-only batches? Empty checkpoints always return a fresh token, and the limit fail can't occur on them, so that case genuinely is malformed. For a batch with updates, treat an absent token as completion and route it through _settle_after_execution_completed just like the terminal case (so queued waiters settle instead of failing).


This would implement #565 exactly.

packages/aws-durable-execution-sdk-python/tests/state_test.py
```diff
@@ -1884,6 +1885,409 @@ def run_batching():
...
+    )
+
+    completion_event = CompletionEvent()
+    state._checkpoint_queue.put(QueuedOperation(terminal_update, completion_event))
+
+    thread = threading.Thread(daemon=True, target=state.checkpoint_batches_forever)
+    thread.start()
+    try:
+        # Returns without raising once the terminal checkpoint is processed.
+        completion_event.wait()
```
yaythomas: `assert completion_event.wait(timeout=2.0)` would bound the wait, prevent infinite hangs if test goes wrong.

a regression in the completion path hangs the suite at the wait, before the finally block's bounded join is ever reached.

ayushiahjolia: Good suggestion, will update this to be bound wait.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
ayushiahjolia: Good catch - agreed, and fixed. Verified behavior for quota limit as well.

yaythomas (approved):

PR #576, opened 2026-07-23
feat(otel): add ExecutionOtelPlugin with Workflow-rooted trace hierarchy

SilanHe:
*Issue #, if available:*

*Description of changes:* Port aws-durable-execution-sdk-js#729 (ExecutionOtelPlugin) to Python.

Adds a self-contained execution-view plugin that produces the deterministic span hierarchy Workflow -> Invocation -> Operation -> Attempt, stitching one trace across every Lambda invocation of a durable execution. The Workflow span is the root (created in an empty context, deterministic span ID derived from the execution ARN) and is exported only on a terminal invocation status; non-terminal statuses (PENDING/RETRY) drop it. Operations are parented under the Workflow span and linked to the current invocation span.

Shared infrastructure is extracted so both plugins reuse it:
- execution_plugin_config.py: canonical ExecutionOtelPluginConfig + ExporterConfig
- provider.py: create_tracer_provider factory (explicit -> global -> auto OTLP)
- instrumentations.py: register_standalone_instrumentations (botocore + urllib3, graceful ImportError degradation)
- deterministic_id_generator.py: derive_workflow_span_id()

The existing OtelPlugin is aliased as InvocationOtelPlugin to match the JS rename while preserving the published public API. pyproject gains an optional "instrumentation" extra for the auto-configured provider path.

Tests: 22 new (test_execution_plugin.py, test_provider.py); full otel suite 76 passed.


SilanHe:
passing tests: https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/30037065713/job/89307566986?pr=576

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -0,0 +1,544 @@
...
+            # invocation. Create + immediately end a linked span.
+            parent = self._resolve_parent(info.parent_id)
+            span = self._start_span(
+                operation_id=info.operation_id,
+                name=info.name or info.operation_id,
+                info=info,
+                parent=parent,
+                start_time=info.start_time,
+                existed=True,
+            )
```
zhongkechen, line 367: > I'm actually not sure what to do here since I'm not very familiar with how Python is resolving the parent operation span when we create an attempt span. 

Actually in Python we don't `resolve` a parent. Python just ensures parent span is always created before the child span.

SilanHe:
yeah I checked the other PR, I think the current implementation should be good then. I opted to not do the cloud integration tests since we're going to implementing them in the conformance test suite anyway.

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -0,0 +1,526 @@
...
+        """Start a span for an operation/attempt and register it."""
+        key = span_key if span_key is not None else operation_id
+        with self._lock:
+            links = self._build_invocation_links()
+            if deterministic:
+                # Operation spans always use the deterministic logical-operation
+                # span ID so a suspended-then-completed operation exports a
+                # single span (on completion) with a stable ID across invocations.
+                self._id_generator.set_next_span_id(
+                    operation_id_to_span_id(self._execution_arn, operation_id)
```
zhongkechen, line 406: Something to note: this always reuses the deterministic operation ID, including replayed terminal operations. Two invocations produced two exported spans sharing one ID but containing different invocation links. OpenTelemetry treats (trace_id, span_id) as a span’s unique identity and provides no merge semantics for multiple completed records.

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -0,0 +1,526 @@
...
+        logger.debug("Durable invocation ended: %s", info)
+        # Operation spans still open here belong to operations that suspended
+        # (e.g. PENDING/RETRYING) rather than completed this invocation. They are
+        # ended only by on_operation_end; drop the references without ending them
+        # so they are not exported as if completed. _reset_state
+        # clears the span map below.
+
+        # End the invocation span regardless of terminal status.
+        if self._invocation_span is not None:
+            self._invocation_span.end()
```
zhongkechen, line 301: no metadata (status and etc) for the span?

zhongkechen (approved):

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
SilanHe: ah good point, I didn't address all the drift between the SDK features in this PR and missed the changes from your PR. I'll follow up in another PR to unblock other work

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
SilanHe: they can start 2 spans with the same ID but the first operation span in the first invocation will not be exported since the operation has not ended

Issue #532, opened 2026-07-14
Remove unused config types: ItemBatcher, BatchedInput, ItemsPerBatchUnit, TerminationMode, StepFuture


yaythomas:
### Summary

`config.py` exposes several types that are not wired into any operation. They accept user input and silently do nothing. This proposes removing them in v2, since removal changes the public config surface and is a breaking change.

### Unused types

**1. `ItemBatcher`, `MapConfig.item_batcher`, `ItemsPerBatchUnit`, `BatchedInput`**

`MapConfig.item_batcher` is never read. `MapExecutor.from_items` (`src/aws_durable_execution_sdk_python/operation/map.py`) consumes every other `MapConfig` field but not `item_batcher`. Items are never batched.

- `ItemsPerBatchUnit` has no consumers. It also contains a latent bug: `COUNT = ("COUNT",)` defines a tuple value due to a trailing comma, and `tests/config_test.py` asserts the tuple.
- `BatchedInput` is never constructed. It appears only in the `map` callable type hint (`U | BatchedInput[Any, U]` in `context.py` and `types.py`), advertising batching that does not exist.
- #304 confirmed the docs side: "ItemBatcher is not implemented yet and reserved for future use." The docs section was removed but the code remained.
- The JS SDK has no batching concept. `MapConfig` in `aws-durable-execution-sdk-js/src/types/batch.ts` has no `itemBatcher`, and `MapFunc` is `(context, item, index, array)`.

Proposed change: remove `ItemBatcher`, `ItemsPerBatchUnit`, `BatchedInput`, and `MapConfig.item_batcher`. Simplify the `map` callable signature to `Callable[[DurableContext, U, int, Sequence[U]], T]`, matching JS `MapFunc`.

**2. `TerminationMode`**

Defined in `config.py` with values TERMINATE, CANCEL, WAIT, ABANDON. Zero references in `src/` outside its own definition. The only usage is a value-assertion test in `tests/config_test.py`. Neither the JS SDK nor the Java SDK has an equivalent.

Proposed change: remove the enum and its test.

**3. `StepFuture`**

Defined in `config.py`. Zero references in `src/` outside its own definition. It mirrors the JS `DurablePromise`, but the Python SDK's synchronous execution model never adopted it. The only usage is a shape test in `tests/config_test.py`.

Proposed change: remove the class and its tests.

**4. `ChildConfig.item_serdes`**

[...]

PR #579, opened 2026-07-23
refactor(config): remove unused config types

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-python/issues/532

*Description of changes:*
Removes dead public surface from the SDK: config types that accept input but are never read, and error-type enums that were never wired into any path. All are non-functional today, so removal only changes the importable
public surface (none are exported from `__init__.py`). Targets v2.

#### Unused config types 
- Remove `ItemBatcher`, `ItemsPerBatchUnit`, `BatchedInput` and `MapConfig.item_batcher`; simplify the map callable to `Callable[[DurableContext, U, int, Sequence[U]], T]`, matching JS `MapFunc`.
- Remove `TerminationMode` (no SDK has an equivalent).
- Remove `StepFuture` (mirrors JS `DurablePromise`; never adopted by the synchronous Python model).
- Remove `ChildConfig.item_serdes` and its stale backward-compat docstring. `MapConfig.item_serdes` / `ParallelConfig.item_serdes` are unaffected (still read by the concurrency executor).
- Remove `WaitStrategyConfig.timeout` and its `timeout_seconds` property. `create_wait_strategy` never reads it; the strategy is client-side with no backend enforcement, and neither the Java nor JS SDK has a working equivalent.

#### Dead error-type enums
- Remove `ChainedInvokeFailedToStartType`, `ChainedInvokeTimeoutType` and `ChainedInvokeStopType` from `lambda_service.py`. The invoke executor already classifies terminal failures by checkpoint status and carries the error type through from the backend `ErrorObject`.

#### Test cleanup
- Consolidate the `suspend_with_optional_resume_delay` helper tests: drop the four duplicates from `invoke_test.py` and cover none/negative/zero/positive in one parametrized test in `suspend_test.py`.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
@@ -14,11 +14,9 @@ from aws_durable_execution_sdk_python.exceptions import ValidationError
 P = TypeVar("P")  # Payload type
 R = TypeVar("R")  # Result type
 T = TypeVar("T")
-U = TypeVar("U")
 
 if TYPE_CHECKING:
     from collections.abc import Callable
-    from concurrent.futures import Future
 
     from aws_durable_execution_sdk_python.lambda_service import OperationSubType
     from aws_durable_execution_sdk_python.retries import RetryDecision
@@ -63,19 +61,6 @@ class Duration:
     def from_days(cls, value: float) -> Duration:
         """Create a Duration from days."""
         return cls(seconds=int(value * 86400))
-
-
-@dataclass(frozen=True)
-class BatchedInput(Generic[T, U]):
-    batch_input: T
-    items: list[U]
-
-
-class TerminationMode(Enum):
-    TERMINATE = "TERMINATE"
-    CANCEL = "CANCEL"
-    WAIT = "WAIT"
-    ABANDON = "ABANDON"
 
 
 class NestingType(Enum):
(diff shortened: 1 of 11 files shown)


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
```diff
@@ -405,10 +337,6 @@ class MapConfig(Generic[T]):
             If None, no limit is imposed and all items are processed concurrently.
             Use this to control resource usage when processing large collections.
```
yaythomas, line 338: nit: a few lines up (line 330) the class docstring still says "including concurrency, batching, completion criteria" — pls drop "batching" now that the batching surface is gone.

yaythomas (commented):
nit: `config.py:423` (InvokeConfig) — while we're removing dead config surface, the commented-out `# retry_strategy: ...` line above the fields could go too (pre-existing, optional). (Posting here because the line isn't in the diff.)

yaythomas:
Verified every removed symbol has zero residual references across all five packages, the examples, and the docs repo, and that MapConfig.item_serdes/ParallelConfig.item_serdes remain live in the concurrency executor. 1421 tests, typecheck, and fmt all clean locally. The ChainedInvoke*Type removal beyond the #532 list is also correct — nothing consumes them and the invoke executor classifies terminal failures from the checkpoint status, with the error type carried through from the backend error object.

Nice catch tightening the zero-delay suspend assertion to TimedSuspendExecution — the deleted duplicate in invoke_test.py asserted the broader SuspendExecution, which only passed via subclassing. The parametrized version now pins the actual behavior (from_delay(msg, 0)).

Thank you very much :-) 

Pretty much ready to approve, just those 2 nits if possible :-)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
ayushiahjolia: Done, updated stale comments.

yaythomas (approved):

PR #580, opened 2026-07-23
test: add plugin conformance suite handlers (10-1..10-7)

wangyb-A:
## What

7 handlers for the `plugin` conformance suite (instrumentation-plugin lifecycle hooks) in `packages/aws-durable-execution-sdk-python-conformance-tests/handlers/plugin/`, plus `template_plugin.yaml`.

## Validation
- Plugin suite is passing, child suite failure not related (fixed in #550 )


wangyb-A:
re AI review: We use print + json intentionally, since the test runner is filtering the top level durable execution arn, and the plugin does not have access to the `context.logger`, the test case print the raw json with durable execution arn so that the test runner can discover and validate the log.

SilanHe (approved):

PR #581, opened 2026-07-24
fix(otel): classify uncaught invocation errors

zhongkechen:
## Summary

- report raw uncaught handler errors to instrumentation as `RETRY` while preserving the terminal `FAILED` durable execution response
- mark `RETRY` invocation spans with OpenTelemetry `ERROR` status
- add regression coverage that keeps durable operation failures classified as `FAILED`

## Root cause

The execution wrapper converts an uncaught handler exception into a `FAILED` durable output before `PluginExecutor` emits the invocation-end event. Instrumentation therefore received `FAILED`, and the OTel plugin also treated an explicit `RETRY` event as `UNSET` rather than `ERROR`.

This caused [OTEL-19](https://github.com/aws/aws-durable-execution-conformance-tests/actions/runs/30063461310/job/89389613751?pr=24) to export an invocation span with `durable.invocation.status: FAILED` instead of the required `RETRY` value.

## Testing

- `hatch run test:all` (`2980 passed, 2 skipped`)
- `hatch run types:check`
- `hatch fmt --check` in the core package
- `hatch fmt --check` in the OTel package

The hosted conformance fixture installs the SDK from `main`, so OTEL-19 will consume this fix once the fixture points at this branch or the change lands.

## CI note

The hosted `child`, `map`, and `parallel` conformance jobs fail on the initial run and a targeted rerun. They fail on custom-serdes output and cloud history assertions unrelated to invocation instrumentation. The immediately preceding unrelated PR run [30064400076](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/30064400076) has the same three failing jobs. All other PR checks pass.

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/plugin.py
@@ -350,7 +350,7 @@ class OtelPlugin(DurableInstrumentationPlugin):
             invocation_span.set_attribute(
                 "durable.invocation.status", info.status.value
             )
-            if info.status is InvocationStatus.FAILED:
+            if info.status in (InvocationStatus.FAILED, InvocationStatus.RETRY):
                 invocation_span.set_status(
                     StatusCode.ERROR, info.error.message if info.error else ""
                 )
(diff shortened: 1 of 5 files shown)


packages/aws-durable-execution-sdk-python-otel/tests/test_invocation_plugin.py
```diff
@@ -175,7 +175,7 @@ def test_invocation_span_records_subsequent_invocation():
     ("invocation_status", "expected_span_status"),
     [
         (InvocationStatus.PENDING, StatusCode.UNSET),
-        (InvocationStatus.RETRY, StatusCode.UNSET),
+        (InvocationStatus.RETRY, StatusCode.ERROR),
```
zhongkechen, line 178: I think we decided to map RETRY to UNSET.

PR #582, opened 2026-07-24
test: add error preservation regression tests

ayushiahjolia:
*Issue #, if available:*
https://github.com/aws/aws-durable-execution-sdk-python/issues/297

*Description of changes:*
The underlying bug was already fixed in [#540](https://github.com/aws/aws-durable-execution-sdk-python/pull/540) - that PR introduced the typed per-operation error hierarchy and stopped `ConcurrentExecutor._create_result` from double-wrapping a branch failure via `ErrorObject.from_exception` (which dropped the caller's original `error_type`).

This PR adds regression tests only to verify that fix holds and to guard against it silently regressing.

Tests - 
- A unit test that `data`/`stack_trace` survive nested `run_in_child_context`  boundaries.
- An end-to-end test that `ctx.parallel()` and `ctx.run_in_child_context()` surface the same `error_type` for the same custom exception.

### DIFF
--- a/packages/aws-durable-execution-sdk-python/tests/e2e/checkpoint_response_int_test.py
+++ b/packages/aws-durable-execution-sdk-python/tests/e2e/checkpoint_response_int_test.py
@@ -11,9 +11,13 @@ from unittest.mock import Mock, patch
 
 import pytest
 
+from aws_durable_execution_sdk_python.concurrency.models import BatchResult
 from aws_durable_execution_sdk_python.config import ChildConfig, Duration
 from aws_durable_execution_sdk_python.context import DurableContext, durable_step
-from aws_durable_execution_sdk_python.exceptions import InvocationError
+from aws_durable_execution_sdk_python.exceptions import (
+    ChildContextError,
+    InvocationError,
+)
 from aws_durable_execution_sdk_python.execution import (
     InvocationStatus,
     durable_execution,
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-python/tests/e2e/checkpoint_response_int_test.py
```diff
@@ -762,3 +765,75 @@ def my_handler(event, context: DurableContext) -> str:
...
+    """ctx.parallel() and ctx.run_in_child_context() surface the same
+    error_type for an identical custom exception: the caller's original type,
+    not the ChildContextError wrapper class name.
+    """
+
+    class PermanentFailure(Exception):
+        pass
+
+    def branch(child_context: DurableContext) -> str:
+        msg = "Invalid input data"
```
yaythomas: nit: a few locals are untyped (`msg` here, `result` below, and `inner_error` in `child_test.py`) — please could you add annotations (e.g. `msg: str = ...`)? Reason: house style types every variable so the type-checker catches typos — see [CONTRIBUTING](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md).

yaythomas:
minor nits, but :shipit: otherwise :-)  nice regression guard to assert exactly the behavior #540 fixed, and the nested-boundary unit test also pins the FAIL checkpoint payloads so the replay path is covered by proxy.

yaythomas (approved):

packages/aws-durable-execution-sdk-python/tests/e2e/checkpoint_response_int_test.py
zhongkechen: Thanks for the review and for pointing to the current `CONTRIBUTING.md` guidance. I understand that this request is consistent with the repository's existing house rule, so this is not a criticism of the review itself.

I do think the underlying rule should be reconsidered, however. Requiring an explicit annotation for every local variable, including variables whose types are directly and precisely inferred, is contrary to normal Python typing practice and makes the code less Pythonic without necessarily improving type safety.

Mypy explicitly documents local type inference:

> "For most variables, if you do not explicitly specify its type, mypy will infer the correct type based on what is initially assigned to the variable."

Source: [Mypy: Type inference and type annotations](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#type-inference)

For example, in a checked function:

```python
msg = "Invalid input data"
```

Mypy infers `msg` as `str`. It will report errors if `msg` is subsequently assigned an incompatible value or used in an operation that does not accept `str`. Writing:

```python
msg: str = "Invalid input data"
```

normally gives the checker no additional information. An explicit annotation can serve as an intentional constraint on future changes to the initializer, but that is useful selectively; it does not justify requiring an annotation on every local.

The Google Python Style Guide, which this repository's contributing guide says it generally follows, gives more selective guidance:

> "If an internal variable has a type that is hard or impossible to infer, specify its type with an annotated assignment."

Source: [Google Python Style Guide: Typing Variables](https://google.github.io/styleguide/pyguide.html#typing-variables)

That guidance treats local annotations as useful when inference is insufficient, not as mandatory duplication when the inferred type is already precise.

The three variables identified here illustrate the distinction:

```python
msg = "Invalid input data"
result = context.parallel([branch])
inner_error = StepError(...)
```

`msg` is inferred directly as `str`. `result` should be inferred from the annotated return type of `context.parallel()`. `inner_error` is inferred directly from the `StepError` constructor.

There is a separate issue with `inner_error`: its enclosing test function does not have a return annotation. By default, mypy does not fully check the bodies of untyped functions. If the goal is comprehensive checking, the more effective solution is to annotate the function:

```python
def test_child_handler_preserves_data_across_nested_boundaries() -> None:
```

Alternatively, the project can enable mypy's `check_untyped_defs` setting. Annotating one local inside an otherwise unchecked function gives much narrower coverage than making the function itself checked.

More generally, blanket local annotations have several costs:

- They duplicate information already produced by the type checker.
- They add visual noise and make meaningful annotations harder to identify.
- They increase maintenance burden during routine refactoring.
- They can unnecessarily constrain a variable to a concrete implementation type when inference or an intentionally selected abstraction would be more appropriate.
- They encourage measuring typing quality by annotation density rather than by whether interfaces and checked code paths are correctly typed.

Explicit local annotations remain valuable for cases such as empty collections, initial `None` values, values originating from `Any` or untyped libraries, intentionally widened interface types, unions across branches, and expressions whose inferred type is unclear or too narrow.

This is also consistent with Python's gradual-typing philosophy. PEP 484 states:

> "Python will remain a dynamically typed language, and the authors have no desire to ever make type hints mandatory, even by convention."

Source: [PEP 484: Non-goals](https://peps.python.org/pep-0484/#non-goals)

That statement does not prevent a repository from adopting a stricter house style, but it does show that annotating every variable is not a general Python typing principle.

I suggest revisiting the corresponding `CONTRIBUTING.md` rule and replacing it with something closer to:

> Annotate function and method parameters and return types so their bodies are checked. For local variables, rely on type inference when it is precise and obvious. Add explicit annotations when inference is unavailable, results in `Any`, is too narrow, or when the annotation expresses an intentional contract that is not evident from the initializer.

The review comment accurately reflects the repository's current written rule. My concern is with that rule itself, not with the reviewer or contributor. Would the maintainers consider revisiting the guidance separately so it better aligns with mypy's documented behavior, the Google Python Style Guide, and common Python typing practice?

zhongkechen: https://github.com/aws/aws-durable-execution-sdk-python/issues/583

PR #585, opened 2026-07-24
ci: add AI pull request review workflow

zhongkechen:
## Summary

- add the Claude and Codex pull request review workflow used by the conformance-tests repository
- isolate review execution from untrusted pull request code and credentials
- minimize superseded review summaries and Claude inline comments

## Reference

- [Original workflow in aws-durable-execution-conformance-tests](https://github.com/aws/aws-durable-execution-conformance-tests/blob/main/.github/workflows/ai-pr-review.yml)

## Testing

- verified workflow and summary helper parity with conformance-tests `origin/main`
- parsed the workflow YAML and review output JSON schema
- ran Bash syntax checks for all support scripts

### DIFF
--- a/.github/prompts/ai-pr-review-schema.json
+++ b/.github/prompts/ai-pr-review-schema.json
@@ -1,0 +1,15 @@
+{
+  "type": "object",
+  "properties": {
+    "summary": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 4000,
+      "description": "Concise Markdown summary of the PR review and any residual test risk"
+    }
+  },
+  "required": [
+    "summary"
+  ],
+  "additionalProperties": false
+}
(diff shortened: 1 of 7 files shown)


yaythomas (approved):
Read,Grep,Glob still lets the model read anything the review user can read, and summary is posted verbatim to a public PR.

The inline-comment MCP server is configured with env: { GITHUB_TOKEN: … } (src/mcp/install-mcp-server.ts:143-151 at the pinned SHA).

I don't know if that config reaches the CLI as an argv blob or a file, nor whether either is reachable by the claude-review user.

Independent of the answer, a cheap belt-and-braces move: post_ai_review_summary.sh already validates the summary's shape, so having it also refuse bodies matching gh[pousr]_[A-Za-z0-9]{36,} or ASIA[0-9A-Z]{16} costs one line? But maybe this is irrelevant.

For large PRs hard-fail the Fail-closed is the right call; the message "GitHub returned an incomplete PR diff" just reads like a transient API fault, maybe mentioning the 300-file cap would stop someone chasing a phantom bug.

zhongkechen:
The workflow serializes the token-bearing MCP config into an inline --mcp-config argument, and forwards that to the Claude CLI.

Good catch with the diff error. I think we can do local diff instead of depending on Github api. I'll follow up later.

PR #591, opened 2026-07-27
fix(plugin): correct replay callbacks and operation span timing

zhongkechen:
## Summary

- stop the plugin executor from invoking `on_operation_start` or `on_operation_end` for terminal operations encountered during replay
- continue invoking `on_operation_start` for replayed operations that are still non-terminal
- remove terminal-replay filtering from both OTel plugins so the behavior is guaranteed by the SDK plugin interface
- start invocation-view operation spans with the current time so continuation spans remain within their invocation

## Root cause

`PluginExecutor.on_operation_replay` unconditionally emitted `on_operation_start` and then also emitted `on_operation_end` for terminal checkpointed operations. That made every plugin responsible for filtering replay history and caused OTel to emit an already-successful submitter again.

The executor now returns before dispatching either callback when the replayed operation has reached `SUCCEEDED`, `FAILED`, `TIMED_OUT`, `CANCELLED`, or `STOPPED`. This aligns the Python plugin lifecycle with the TypeScript behavior.

Invocation-view operation spans also reused durable timestamps from prior invocations. They now use the current local time for both start and end so continuation spans do not overlap their invocation parent.

Conformance failures:

- https://github.com/aws/aws-durable-execution-conformance-tests/actions/runs/30304154081/job/90104146029
- https://github.com/aws/aws-durable-execution-conformance-tests/actions/runs/30310125440/job/90123537458

## Testing

- `hatch run test:all` (`3062 passed, 2 skipped`, 5 subtests passed)
- plugin executor replay tests (`2 passed`, 5 subtests passed)
- replay-flow state tests (`7 passed`)
- `hatch run dev-otel:test` (`92 passed`)
- `hatch run types:check`
- `hatch fmt --check` for core and OTel packages

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
@@ -9,7 +9,6 @@ from typing import Any
 
 from aws_durable_execution_sdk_python.lambda_service import (
     InvocationStatus,
-    OperationStatus,
     OperationType,
 )
 from aws_durable_execution_sdk_python.plugin import (
@@ -50,16 +49,6 @@ logger = logging.getLogger(__name__)
 
 _SpanAttributes = dict[str, str | bool | int]
 
-_TERMINAL_OPERATION_STATUSES = frozenset(
-    {
-        OperationStatus.SUCCEEDED,
-        OperationStatus.FAILED,
-        OperationStatus.TIMED_OUT,
-        OperationStatus.CANCELLED,
-        OperationStatus.STOPPED,
-    }
-)
-
 
 def _to_otel_timestamp(dt: datetime.datetime | None) -> int | None:
     """Convert a datetime to OTel timestamp (nanoseconds since epoch), or None."""
@@ -80,8 +69,7 @@ class InvocationOtelPlugin(DurableInstrumentationPlugin):
     span for an operation uses that deterministic ID; later continuation spans
     use newly generated span IDs and link back to the deterministic span ID so
     trace viewers can relate retries and cross-invocation completions to the
-    original logical operation. Terminal operations encountered only as replay
-    history do not emit another span.
+    original logical operation.
 
     Args:
         trace_provider: OpenTelemetry tracer provider used to create spans.
@@ -385,8 +373,6 @@ class InvocationOtelPlugin(DurableInstrumentationPlugin):
     def on_operation_start(self, info: OperationStartInfo) -> None:
         """Called when an operation begins. Creates a span for the operation."""
         logger.debug("Durable operation started: %s", info)
-        if info.is_replayed and info.status in _TERMINAL_OPERATION_STATUSES:
-            return
         if info.operation_type is OperationType.CONTEXT:
             # Context operations are tracked using on_user_function_start.
             return
@@ -410,14 +396,8 @@ class InvocationOtelPlugin(DurableInstrumentationPlugin):
         invocation is completing an operation that began earlier, so a short
         continuation span is created and linked to the deterministic logical
         operation span before being ended.
-
-        Terminal replay callbacks describe history that completed before this
-        invocation and do not produce another span. Operations completed by the
-        backend while suspended are delivered with ``is_replayed=False``.
         """
         logger.debug("Durable operation ended: %s", info)
-        if info.is_replayed:
-            return
         if info.operation_type is OperationType.CONTEXT:
             # Context operations are tracked using on_user_function_end.
             return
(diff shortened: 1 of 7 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
```diff
@@ -373,6 +385,8 @@ def on_invocation_end(self, info: InvocationEndInfo) -> None:
     def on_operation_start(self, info: OperationStartInfo) -> None:
```
SilanHe, line 378: more meta comment about the plugin interface in Python: does this mean that given an operation that completes with a terminal status in the first invocation, and there are 5 invocations in total. Will the on_operation_start be triggered with OperationStartInfo 5 times for that operation?

zhongkechen: Yeah, that is the behavior without this fix

SilanHe: hmm in TS, the plugin interface onOperationStart behaves in the following manner:

1. SDK calls onOperationStart with isFirst=true when the operation has started for the first time. 
2. SDK calls onOperationStart is called with isFirst=false for subsequent times the operation is processed but whose status is not terminal.
3. SDK does NOT call onOperationStart when the operation status is terminal

SilanHe: As a result, the Otel Plugin doesn't need this "fix" at all. Is there no way to close this gap within the SDK's plugin interface?

zhongkechen: Python plugin interface behaves the same except no 3:

> SDK does NOT call onOperationStart when the operation status is terminal

onOperationStart is still called when its context is replayed

SilanHe: hmm this is different in Java as well but rather 2.

> SDK calls onOperationStart is called with isFirst=false for subsequent times the operation is processed but whose status is not terminal.


| Rule | TS | Java | Match |
|---|---|---|---|
| 1. First → `isReplay=false` | all types | all types | ✅ |
| 2. Non-terminal replay → `isReplay=true` | all types | STEP/CONTEXT only | ⚠️ partial |
| — WAIT / INVOKE / CALLBACK (rule 2) | fires | does not fire | ❌ |
| 3. Terminal → no call | no call | no call | ✅ |

zhongkechen: We can align the behavior to what TS currently has.

zhongkechen: Updating the PR with the terminal status check on plugin executor side.

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
```diff
@@ -383,7 +397,7 @@ def on_operation_start(self, info: OperationStartInfo) -> None:
             operation_id=info.operation_id,
             name=info.name or info.operation_id,
             attributes=attributes,
-            start_time=info.start_time,
+            start_time=datetime.datetime.now(datetime.UTC),
```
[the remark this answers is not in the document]
zhongkechen: Addressed in 9835af3. `on_operation_end` now calls `_end_span(info.operation_id)` without passing the backend `info.end_time`; `Span.end(end_time=None)` timestamps the end from OpenTelemetry’s local clock. The regression tests assert `invocation.start <= operation.start <= operation.end <= invocation.end`, and continuation spans are bounded by the local callback times.

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
```diff
@@ -426,7 +447,7 @@ def on_operation_end(self, info: OperationEndInfo) -> None:
         else:
             span.set_status(StatusCode.OK)
 
-        end_timestamp = info.end_time
+        end_timestamp = None if is_continuation else info.end_time
```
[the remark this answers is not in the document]
zhongkechen: Addressed in 9835af3. The operation end path no longer supplies `info.end_time`; `_end_span` calls `Span.end(end_time=None)`, so the end timestamp comes from OpenTelemetry’s local clock, matching the local-clock start. Regression coverage verifies non-negative local-clock ordering.

SilanHe (approved):

PR #595, opened 2026-07-28
fix(testing): honor history reverse order

### LINKED ISSUE #594: [Bug]: Testing WebServer ignores ReverseOrder in GetDurableExecutionHistory
## Expected Behavior

Calling Boto3's `get_durable_execution_history` with `ReverseOrder=True` against the testing SDK's local `WebServer` should return events in reverse chronological order.

For a completed execution with event IDs `[1, 2, 3, 4, 5]`, the expected IDs are:

```text
ReverseOrder=False: [1, 2, 3, 4, 5]
ReverseOrder=True:  [5, 4, 3, 2, 1]
```

## Actual Behavior

The local `WebServer` returns the same chronological order for both values:

```text
ReverseOrder=False: [1, 2, 3, 4, 5]
ReverseOrder=True:  [1, 2, 3, 4, 5]
```

The server access log confirms that Boto3 sends the parameter:

```text
GET /2025-12-01/durable-executions/<execution-id>/history?ReverseOrder=true&MaxItems=100
```

## Steps to Reproduce

1. Start an `aws_durable_execution_sdk_python_testing.web.server.WebServer` for a `DurableFunctionTestRunner`.
2. Run any durable workflow that produces multiple history events and wait for it to reach `SUCCEEDED`.
3. Point a Boto3 Lambda client directly at the local testing SDK WebServer.
4. Fetch the history in each order:

```python
forward = client.get_durable_execution_history(
    DurableExecutionArn=execution_arn,
    ReverseOrder=False,
    MaxItems=100,
)["Events"]

reverse = client.get_durable_execution_history(
    DurableExecutionArn=execution_arn,
    ReverseOrder=True,
    MaxItems=100,
)["Events"]

forward_ids = [event["EventId"] for event in forward]
reverse_ids = [event["EventId"] for event in reverse]

print(forward_ids)  # [1, 2, 3, 4, 5]
print(reverse_ids)  # [1, 2, 3, 4, 5], expected [5, 4, 3, 2, 1]
```

I reproduced this by calling the testing SDK WebServer directly, without an HTTP proxy between Boto3 and the server.

## SDK Version

- `aws-durable-execution-sdk-python-testing==1.2.1`
- `aws-durable-execution-sdk-python==1.5.0`
- `boto3==1.43.28`
- `botocore==1.43.28`

## Python Version

3.12

## Is this a regression?

No — I am not aware of a previously working version.

## Additional Context

The executor already supports reverse ordering, but `GetDurableExecutionHistoryHandler` does not parse the `ReverseOrder` query parameter and currently calls:

```python
self.executor.get_execution_history(
    execution_arn,
    include_execution_data=include_execution_data,
    reverse_order=False,
    marker=marker,
    max_items=int(max_items) if max_items else None,
)
```

Relevant prior work:

- aws/aws-durable-execution-sdk-python-testing#103 added executor support for `reverse_order`.
- aws/aws-durable-execution-sdk-python-testing#125 wired `IncludeExecutionData`, `MaxItems`, and `Marker` into the HTTP handler, but left `reverse_order=False`.

This affects clients that depend on newest-first pagination to incrementally process unseen history events.

mahayash315:
## Summary

- parse the `ReverseOrder` query parameter in the testing WebServer
- forward the parsed value to `get_execution_history`
- cover omitted, `false`, and `true` values with a regression test

## Root cause

The WebServer handler hard-coded `reverse_order=False`, even though the executor already supports reverse-order history and pagination.

## Testing

- `hatch run dev-testing:test` (1,365 passed)
- `hatch run dev-testing:typecheck`
- `hatch fmt --check`
- focused history-handler tests (8 passed)

Fixes #594

### DIFF
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/web/handlers.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/web/handlers.py
@@ -502,12 +502,18 @@ class GetDurableExecutionHistoryHandler(EndpointHandler):
                 if include_execution_data_str
                 else False
             )
+            reverse_order_str: str | None = self._parse_query_param(
+                request, "ReverseOrder"
+            )
+            reverse_order: bool = (
+                reverse_order_str == "true" if reverse_order_str else False
+            )
 
             history_response: GetDurableExecutionHistoryResponse = (
                 self.executor.get_execution_history(
                     execution_arn,
                     include_execution_data=include_execution_data,
-                    reverse_order=False,
+                    reverse_order=reverse_order,
                     marker=marker,
                     max_items=int(max_items) if max_items else None,
                 )
--- a/packages/aws-durable-execution-sdk-python-testing/tests/web/handlers_test.py
+++ b/packages/aws-durable-execution-sdk-python-testing/tests/web/handlers_test.py
@@ -1377,6 +1377,53 @@ def test_get_durable_execution_history_handler_with_include_execution_data_false
     )
 
 
+@pytest.mark.parametrize(
+    ("reverse_order_param", "expected_reverse_order"),
+    [
+        (None, False),
+        ("false", False),
+        ("true", True),
+    ],
+)
+def test_get_durable_execution_history_handler_with_reverse_order(
+    reverse_order_param: str | None, expected_reverse_order: bool
+):
+    """Test GetDurableExecutionHistoryHandler with ReverseOrder."""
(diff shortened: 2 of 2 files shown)


yaythomas:
thank you so much @mahayash315, great contribution, welcome to dex!

Verified the wire contract against the public Lambda API model: ReverseOrder querystring name, lowercase true/false serialization from boto3, and absent-defaults-to-false all match the service exactly.

:shipit:

just waiting for CI to clear.

yaythomas (approved):

PR #599, opened 2026-07-29
fix(otel): leave pending invocation spans unset

zhongkechen:
## Summary
- leave invocation span status unset when the durable invocation is pending
- preserve `durable.invocation.status=PENDING` for lifecycle filtering
- align both OTel plugin implementations and their status-mapping tests

## Testing
- `hatch run dev-otel:test` (92 passed)
- `hatch run types:check`
- `hatch fmt --check` in `packages/aws-durable-execution-sdk-python-otel`

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
@@ -285,19 +285,15 @@ class ExecutionOtelPlugin(DurableInstrumentationPlugin):
 
         # End the invocation span regardless of terminal status. Record the
         # invocation status and map it to a span status:
-        #   SUCCEEDED/PENDING -> OK  (this invocation did its work, whether it
-        #                             completed the execution or cleanly suspended)
-        #   FAILED            -> ERROR
-        #   RETRY             -> UNSET
-        # RETRY is left UNSET because the plugin interface cannot tell whether the
-        # execution/workflow was STOPPED or TIMED_OUT: a RETRY invocation is not a
-        # definitive failure of the execution, so we avoid marking the span ERROR.
+        #   SUCCEEDED     -> OK
+        #   FAILED        -> ERROR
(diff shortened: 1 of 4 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -285,19 +285,15 @@ def on_invocation_end(self, info: InvocationEndInfo) -> None:
...
-        #   SUCCEEDED/PENDING -> OK  (this invocation did its work, whether it
-        #                             completed the execution or cleanly suspended)
-        #   FAILED            -> ERROR
-        #   RETRY             -> UNSET
-        # RETRY is left UNSET because the plugin interface cannot tell whether the
-        # execution/workflow was STOPPED or TIMED_OUT: a RETRY invocation is not a
-        # definitive failure of the execution, so we avoid marking the span ERROR.
+        #   SUCCEEDED     -> OK
+        #   FAILED        -> ERROR
+        #   PENDING/RETRY -> UNSET
```
SilanHe, line 290: this isn't what we agreed IIRC:

https://github.com/aws/aws-durable-execution-conformance-tests/issues/23#issuecomment-5097376122

>In execution view, execution/workflow span status can use this mapping from PluginInvocationStatus to span Status:

>FAILED -> ERROR
SUCCEEDED -> OK
RETRY/PENDING -> UNSET

>which reflects the execution status

>For invocation span status, we want to use the following mapping from PluginInvocationStatus to span Status:

>SUCCEEDED/PENDING -> OK
RETRY -> UNSET
FAILED -> ERROR

>We also want to document that the plugin interface doesn't know if an invocation.workflow was STOPPED or TIMED_OUT, which is also why RETRY maps to UNSET. The workflow span should be INTERNAL as well (span.kind=INTERNAL).

zhongkechen: Oh, can't remember we changed pending to be ok. I can update the test requirements.

PR #602, opened 2026-07-29
fix(otel): reuse deterministic ID generator

zhongkechen:
## Summary
- install the deterministic ID generator on a tracer provider only once, then reuse it across plugin instances
- apply the shared-generator behavior to both `InvocationOtelPlugin` and `ExecutionOtelPlugin`
- add chained-handler regressions that invoke only the second plugin instance

## Problem
Chained-invoke modules can define both `handler` and `target_handler` in the same module. Each decorator creates its own OTel plugin during module import, so both plugins resolve the same tracer provider and instrumentation scope.

OpenTelemetry caches tracers by instrumentation scope, and each tracer captures the provider ID generator when that tracer is created. Previously the initialization sequence was:

1. The first plugin installed its deterministic generator and created the cached tracer.
2. The second plugin installed a different deterministic generator, but received the already-cached tracer.
3. The target handler configured the second generator while its `Workflow` span was created by a tracer that still referenced the first generator.

Only the target plugin receives lifecycle callbacks in the target Lambda. Its `Invocation` span can still inherit the propagated X-Ray context correctly, while its parentless `Workflow` span falls back to a different trace ID. Modules with one plugin instance do not encounter the mismatch.

## Fix
`DeterministicIdGenerator.install_on_provider` now reuses an existing deterministic generator or installs one when the provider still has its original generator. Both OTel plugins use this path before requesting their tracer.

As a result, every plugin instance sharing a provider also shares the generator already captured by the cached tracer. The original provider generator remains the fallback when the deterministic generator is installed for the first time.

## Regression coverage
The invocation-view and execution-view integration suites each create two plugins with the same provider and instrumentation scope, verify that OpenTelemetry returns the cached tracer, invoke only the second plugin, and assert that its `Workflow` span uses the propagated X-Ray trace ID. This mirrors the source/target handler import order in the chained-invoke conformance example.

## Testing
- `hatch run dev-otel:test` (101 passed)
- `hatch run types:check`
- `hatch fmt --check` (OTel package)

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
@@ -116,10 +116,9 @@ class ExecutionOtelPlugin(DurableInstrumentationPlugin):
         from opentelemetry.sdk.trace import TracerProvider as SdkTracerProvider
 
         if isinstance(self._provider, SdkTracerProvider):
-            self._id_generator = DeterministicIdGenerator(
-                fallback_id_generator=getattr(self._provider, "id_generator", None)
+            self._id_generator = DeterministicIdGenerator.install_on_provider(
+                self._provider
             )
-            self._provider.id_generator = self._id_generator
         else:
             logger.warning(
                 "ExecutionOtelPlugin expected an SDK TracerProvider but got %s; "
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/deterministic_id_generator.py
@@ -7,8 +7,13 @@ import hashlib
 import os
 import re
 from datetime import UTC, datetime
+from typing import TYPE_CHECKING
 
 from opentelemetry.sdk.trace import IdGenerator, RandomIdGenerator
+
+
+if TYPE_CHECKING:
+    from opentelemetry.sdk.trace import TracerProvider
 
 
 HASHED_ID_PATTERN = re.compile(r"^[0-9a-f]{16}$")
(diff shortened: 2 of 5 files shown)


ayushiahjolia (approved):

SilanHe:
I don't really get this fix. Is this strictly for the case where a customer might be using both the Invocation plugin and the Execution plugin?

zhongkechen:
> I don't really get this fix. Is this strictly for the case where a customer might be using both the Invocation plugin and the Execution plugin?

If users create any of our Otel plugins more than once, the second deterministic id generator we created for the second plugin will actually not be used by the tracer provider.

I think it's because the tracer provider creates a cached tracer when get_tracer is called for the first time.

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -116,10 +116,9 @@ def __init__(self, config: OtelPluginConfig | None = None) -> None:
         from opentelemetry.sdk.trace import TracerProvider as SdkTracerProvider
 
         if isinstance(self._provider, SdkTracerProvider):
-            self._id_generator = DeterministicIdGenerator(
-                fallback_id_generator=getattr(self._provider, "id_generator", None)
```
SilanHe, line 120: hmm isn't 

>self._id_generator = DeterministicIdGenerator(
                fallback_id_generator=getattr(self._provider, "id_generator", None)

identical to

>generator = cls(fallback_id_generator=current_generator)

SilanHe: I really don't see what this PR is doing

zhongkechen: The fix is the check before it actually installs the id generator

```
        current_generator = provider.id_generator
        if isinstance(current_generator, cls):
            return current_generator
```

Issue #519, opened 2026-07-08
[Feature]: Custom completion predicate (shouldComplete) for map and parallel


ParidelPooya:
### What would you like?

Similar to this: https://github.com/aws/aws-durable-execution-sdk-js/issues/701

### Is this a breaking change?

No

### Does this require an RFC?

No

PR #605, opened 2026-07-31
feat: add should_complete to CompletionConfig

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/519

*Description of changes:*
Adds a `should_complete` predicate to CompletionConfig for map and parallel operations, giving full control over when a batch completes early.

```
from aws_durable_execution_sdk_python import complete_batch, continue_batch

config = CompletionConfig(
    should_complete=lambda status: (
        complete_batch() if status.success_count >= 2
        else continue_batch()
    )
)

```
The predicate receives a CompletionStatus snapshot (counts + per-item statuses) and returns a CompletionDecision - either continue_batch() or complete_batch(outcome). The outcome determines whether `CUSTOM_COMPLETION_SUCCEEDED` or `CUSTOM_COMPLETION_FAILED` is reported.

Key design decisions - 
- Cannot be combined with threshold fields (raises ValidationError)
- Predicate must be deterministic and side-effect-free
- Replay uses the checkpointed decision, never re-invokes the predicate
- On resumed invocations, deferred until prior checkpoints replay

Testing - 
- Unit tests, integration tests and examples


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
```diff
@@ -131,11 +216,17 @@ class CompletionConfig:
...
+        # Custom predicate: stop once 2 successes are observed
+        config = CompletionConfig(
+            should_complete=lambda status: status.success_count >= 2
+        )
     """
 
     min_successful: int | None = None
     tolerated_failure_count: int | None = None
     tolerated_failure_percentage: int | float | None = None
+    should_complete: Callable[[CompletionStatus], bool] | None = None
```
yaythomas: JS uses `shouldComplete: (status) => CompletionDecision` with `continueBatch()` / `completeBatch(outcome)` factories and `CompletionOutcome.SUCCEEDED | FAILED`. The FAILED outcome marks the whole batch failed (`throwIfError()` throws) even with zero item failures for the quorum-can't-be-met case. A bool can't express that.

Is there a reason not to mirror the JS shape with a frozen `CompletionDecision` dataclass + factories?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
```diff
@@ -30,16 +37,13 @@
...
-    SUCCEEDED = "SUCCEEDED"
-    FAILED = "FAILED"
-    STARTED = "STARTED"
 
 
 class CompletionReason(Enum):
     ALL_COMPLETED = "ALL_COMPLETED"
     MIN_SUCCESSFUL_REACHED = "MIN_SUCCESSFUL_REACHED"
     FAILURE_TOLERANCE_EXCEEDED = "FAILURE_TOLERANCE_EXCEEDED"
+    CUSTOM_COMPLETION = "CUSTOM_COMPLETION"
```
yaythomas: Single `CUSTOM_COMPLETION` vs JS has `CUSTOM_COMPLETION_SUCCEEDED` / `CUSTOM_COMPLETION_FAILED`. This string is serialized into checkpoints, so it is used. Follows on from the `CompletionDecision` comment on `config.py`.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -190,10 +237,42 @@ def submit(branch: Branch[CallableType, ResultType]) -> None:
...
+            # failing predicates could give a different answer on re-call).
+            custom_predicate_fired: bool = False
+            # Gate custom predicate evaluation until at least one branch
+            # reaches a terminal state. Without this, a predicate like
+            # `failure_count == 0` would fire immediately before any work runs.
+            # Note: on resumed invocations, previously-completed branches
+            # replay concurrently and report terminal events quickly from
+            # their checkpoints. The predicate is only evaluated after these
+            # events arrive, so it sees accurate terminal states.
+            has_terminal_event: bool = False
```
yaythomas: The `has_terminal_event` gate means an always-true predicate runs exactly 1 item, but in JS the same predicate stops the batch before any item runs (explicit hang guard, empty result). So the same customer predicate results in different behavior across SDKs. Is there a reason to diverge here? The CI Codex bot is independently also flagging a related resumed-replay partial-snapshot concern.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
```diff
@@ -98,21 +129,53 @@ def is_tolerance_exceeded(self, failed: int) -> bool:
...
+            # has reached a terminal state. This prevents predicates like
+            # `failure_count == 0` from firing before any work runs.
+            if not has_terminal_event:
+                return False
+            status: CompletionStatus = self._build_status(succeeded, failed, items)
+            return self.should_complete(status)
         return self.min_successful is not None and succeeded >= self.min_successful
 
-    def reason(self, succeeded: int, failed: int) -> CompletionReason:
+    def reason(
```
yaythomas, line 215: Python checks all-completed before the predicate → `ALL_COMPLETED`; JS checks the predicate first → `CUSTOM_COMPLETION_*` when both are true at once. Is there a reason not to match the JS ordering, since customers may branch on this field?

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
```diff
@@ -99,6 +99,77 @@ class NestingType(Enum):
...
+class BatchItemStatus(Enum):
+    """The status of a batch item in map/parallel operations."""
+
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+    STARTED = "STARTED"
+
+
+@dataclass(frozen=True)
+class CompletionItemStatus:
```
yaythomas, line 160: JS is `{index, name?, status?}`; this PR drops `name` (parallel branches can be named) and adds `result`/`error_message`, which JS deliberately doesn't expose. Is there a reason for the richer shape? Also the two snapshot builders disagree on error format (`str(branch.error)` vs `item.error.message`).

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
```diff
@@ -166,7 +260,12 @@ def _validate_for_total(self, total: int) -> None:
         before the operation's child context starts, so the error surfaces as
         a bare ValidationError (matching wait and wait_for_condition
         validation) instead of a checkpointed operation failure.
+
+        When should_complete is set the predicate takes full precedence and
+        threshold fields are ignored, so threshold validation is skipped.
         """
+        if self.should_complete is not None:
```
yaythomas: JS makes the combo a compile-time error; Python's analog is a `ValidationError` in `__post_init__`, not silently ignoring thresholds. Is there a reason not to raise? (This would also make the `_validate_for_total` skip unnecessary.)

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
```diff
@@ -346,7 +409,28 @@ def from_items(
...
+                index=item.index,
+                status=item.status,
+                result=item.result
+                if item.status is BatchItemStatus.SUCCEEDED
+                else None,
+                error_message=(
+                    item.error.message
+                    if item.status is BatchItemStatus.FAILED
+                    and item.error is not None
+                    and hasattr(item.error, "message")
```
yaythomas: `hasattr(item.error, "message")` can never be False — `error` is typed `ErrorObject | None` and `ErrorObject` is a dataclass where the `message` attribute always exists. The `item.error is not None` check is the only guard needed. Please see [CONTRIBUTING → Data Structures & Typing](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#data-structures--typing): rely on exact type declarations rather than duck typing.

packages/aws-durable-execution-sdk-python/tests/concurrency_test.py
```diff
@@ -3731,3 +3850,643 @@ def test_operation_id_namespace_derivation_is_stable():
...
+
+    result: BatchResult = executor.execute(execution_state, executor_context)
+
+    assert result.completion_reason is CompletionReason.CUSTOM_COMPLETION
+
+    # After a terminal event the snapshot was rebuilt from live branch state.
+    # Verify predicate observed correct statuses:
+    assert len(observed_items) == 4
+    statuses: set = {status for _, status in observed_items}
+    from aws_durable_execution_sdk_python.config import BatchItemStatus
```
yaythomas: Inline `from aws_durable_execution_sdk_python.config import BatchItemStatus` inside the test body — please move to the top of the module with the other imports, per [CONTRIBUTING → General style](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#general-style) → [Google Python Style Guide §3.13](https://google.github.io/styleguide/pyguide.html#313-imports-formatting) ("imports are always put at the top of the file").

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -370,7 +460,13 @@ def _create_result(
                     raise InvalidStateError(msg)
 
         if completion_reason is None:
-            completion_reason = self.policy.reason(succeeded, failed)
+            # Fallback: supply items snapshot so quorum predicates work here too.
```
yaythomas, line 471: For posterity: replay normally uses the checkpointed completion reason and never re-invokes the predicate, but this fallback and `BatchResult.from_items` (for checkpoints that carry no recorded reason) re-infer the reason and therefore call the predicate again, against reconstructed items. JS does the same in its legacy-summary fallback, so no action on the re-invocation itself — but note these need updating if the decision-API change is made and we align with JS.

yaythomas (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -190,10 +237,42 @@ def submit(branch: Branch[CallableType, ResultType]) -> None:
...
+                    items_snapshot = (
+                        self._build_items_snapshot()
+                        if self.policy.should_complete is not None
+                        else ()
+                    )
+                    needs_snapshot_rebuild = False
                 if self.policy.is_complete(
-                    succeeded, failed
-                ) or not self.policy.should_continue(failed):
+                    succeeded, failed, items_snapshot, has_terminal_event
```
yaythomas: > P1] Restore checkpointed branch state before evaluating the predicate — packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py:264

The Codex review raised a legit issue here? When a batch that suspended midway resumes, every branch restarts as PENDING and the coordinator only rediscovers checkpointed outcomes as replayed worker events arrive, in arbitrary thread order. The predicate gate opens on the first terminal event, so `should_complete` can evaluate a snapshot where an already checkpointed success still shows as STARTED. That means replay arrival order can change the decision, the completion reason, and which results the batch reports.

The fix Codex suggests is to initialize branch state from checkpoints, but this is troublesome. Marking branches COMPLETED up front on the coordinator thread means materializing results outside the worker and child context path, which is more invasive and touches the threading assumptions of the coordinator.

So maybe instead: count the terminal checkpoints before the loop, and defer predicate evaluation until that many terminal events have arrived again. Something like replacing the `has_terminal_event` bool with `terminal_events_seen >= restored_target`, where `restored_target` is computed up front via `get_checkpoint_result` per executable (cheap, checkpoints are already in memory). The replayed branches still flow through the normal worker path so results materialize correctly, and the first predicate evaluation is guaranteed to see everything the previous invocation knew. On a fresh run `restored_target` is 0, which also happens to restore the JS zero progress semantics from my other comment, so one rule closes both.

No deadlock risk. The gate only suppresses early completion, scheduling continues, and replayed branches complete near instantly, so the gate always opens and `ALL_COMPLETED` stays reachable.

Plus the regression test Codex asks for: suspend and resume with two terminal checkpointed branches forced to report again in reverse order, asserting the reason and result set are order independent.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
ayushiahjolia: result/error_message was for conditions like 'stop when we get this specific error'. but that will complicate this and diverge from JS. Let me update to match JS.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
ayushiahjolia: Agree, Fail-fast is better than silent ignore.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
ayushiahjolia: Makes sense, Done - implemented suggested approach with restored_terminal_target counting prior checkpoints and deferring evaluation until that many events re-arrive, added the reversed-order regression test as well.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
ayushiahjolia: Updated, both paths now use the CompletionDecision.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
ayushiahjolia: we removed error_message from CompletionItemStatus entirely to match the JS.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
ayushiahjolia: Reordered, Predicate is now checked before all-completed in both is_complete and reason, matching JS.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/models.py
ayushiahjolia: split into CUSTOM_COMPLETION_SUCCEEDED and CUSTOM_COMPLETION_FAILED, matching JS.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
ayushiahjolia: Agreed and done - added frozen CompletionDecision dataclass with CompletionOutcome enum and continue_batch()/complete_batch(outcome) factories.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
ayushiahjolia: We can't fully match JS here because JS processes branch results sequentially so replay order is always deterministic. Our threading model means replayed events arrive in nondeterministic order. Fixed it to fire immediately on fresh runs (target=0, matching JS), but on resumed invocations we defer until all prior checkpoints have reported back to avoid order-dependent decisions.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -195,10 +242,68 @@ def submit(branch: Branch[CallableType, ResultType]) -> None:
...
+                        else ()
+                    )
+                    needs_snapshot_rebuild = False
                 if self.policy.is_complete(
-                    succeeded, failed
-                ) or not self.policy.should_continue(failed):
+                    succeeded,
+                    failed,
+                    items_snapshot,
+                    restored_events_seen >= len(restored_terminal_indexes),
```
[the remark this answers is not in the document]
ayushiahjolia: Not an issue. The gate defers the completion decision, not scheduling. New branches starting during restoration is correct behavior - they would have started on the prior invocation too (same index order, same max_concurrency). Blocking them would change the observable execution order, which is a worse determinism violation. Branch side effects are checkpointed regardless of scheduling order - that's the SDK's core guarantee.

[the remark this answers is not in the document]
ayushiahjolia: The completion decision can't be persisted synchronously before the parent checkpoint - that crash window is inherent to the checkpoint model, same as the existing threshold configs. We document the same contract the JS SDK uses: the predicate depends only on the provided CompletionStatus, and the batch may complete with slightly more items than it first observed when several finish at once. Added a regression test (test_executor_should_complete_stable_when_extra_items_complete_on_resume) covering the resume window where the gate opens at a higher count.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -195,10 +243,26 @@ def submit(branch: Branch[CallableType, ResultType]) -> None:
...
-                if self.policy.is_complete(
-                    succeeded, failed
-                ) or not self.policy.should_continue(failed):
+                if needs_snapshot_rebuild:
+                    items_snapshot = (
+                        self._build_items_snapshot()
+                        if self.policy.should_complete is not None
+                        else ()
+                    )
+                    decision = self.policy.evaluate(succeeded, failed, items_snapshot)
```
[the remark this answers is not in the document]
ayushiahjolia: Predicates are required to be monotonic by contract. If a batch is incomplete, its terminal set didn't satisfy the predicate, so no subset of it can either - reordered replay on resume can't make a monotonic predicate fire early.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
```diff
@@ -212,6 +276,7 @@ def submit(branch: Branch[CallableType, ResultType]) -> None:
                     submit(pending.popleft())
                     in_flight += 1
                     running += 1
+                    needs_snapshot_rebuild = True
```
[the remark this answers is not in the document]
ayushiahjolia: Admitting only flips an item to STARTED; the counts don't change. So this only matters to a predicate keyed on STARTED, which carries no outcome info. Real predicates are already evaluated at the initial call and on every terminal/suspend event.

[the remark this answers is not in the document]
ayushiahjolia: Admitting only flips an item to STARTED; the counts don't change. So this only matters to a predicate keyed on STARTED, which carries no outcome info. Real predicates are already evaluated at the initial call and on every terminal/suspend event.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/concurrency/executor.py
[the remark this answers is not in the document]
ayushiahjolia: Predicates are required to be monotonic by contract. If a batch is incomplete, its terminal set didn't satisfy the predicate, so no subset of it can either - reordered replay on resume can't make a monotonic predicate fire early.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
```diff
@@ -119,8 +222,40 @@ class CompletionConfig:
...
+            snapshot and never assume any item has completed. When a resumed
+            invocation finds the batch already completed, the recorded
+            completion outcome is replayed and the predicate is not called
+            again; otherwise the batch runs live and the predicate is
+            evaluated as described.
+
+            The predicate must be deterministic and side-effect-free, and
+            depend only on the CompletionStatus provided, never on external
+            state. For a completion that must survive a mid-run resume,
+            prefer a monotonic decision (for example success_count >= n) so
```
yaythomas: nit: "prefer" is a "must"? You argued on the codex review comments that resume determinism relies on monotonicity as a hard contract. A predicate that is false on the suspension time terminal set is false on every subset of it, but only if monotonicity is required.

packages/aws-durable-execution-sdk-python/tests/e2e/should_complete_suspend_resume_int_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_should_complete_across_suspend_resume() -> None:
+    """Predicate fires correctly after a branch suspends and resumes."""
+    with DurableFunctionTestRunner(handler=handler) as runner:
+        result: DurableFunctionTestResult = runner.run(input="{}", timeout=15)
+
+    assert result.status is InvocationStatus.SUCCEEDED
+    assert result.result is not None
+
+    import json
```
yaythomas: inline `import. Please move it to the top of the module. [Google style §3.13](https://google.github.io/styleguide/pyguide.html#313-imports-formatting) via [CONTRIBUTING → General style](https://github.com/aws/aws-durable-execution-sdk-python/blob/main/CONTRIBUTING.md#general-style).

yaythomas (changes requested):
just two minor noin-func nits :)

packages/aws-durable-execution-sdk-python/tests/e2e/should_complete_suspend_resume_int_test.py
ayushiahjolia: Sorry agent keeps making this mistake.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/config.py
ayushiahjolia: I'll reword this.

yaythomas (approved):
:shipit:

PR #615, opened 2026-08-06
Add hook-info field-shape conformance handlers

wangyb-A:
## Summary
Adds 5 conformance handlers (`handlers/plugin/plugin_{invocation_info,operation_info,attempt_info,operation_change,context_info}_shape.py`) + `template_plugin.yaml` entries implementing plugin hook-info FIELD-SHAPE requirements 10-19..10-23 from aws/aws-durable-execution-conformance-tests#72 (land that PR first).

Each handler logs ONE single-line JSON record per hook event: a CANONICAL camelCase DUMP of that hook's own info parameter (None/unexposed fields OMITTED — a missing key fails its assertion). GA note: experimental payload fields (execution input/result, operation result) are still DUMPED when exposed but are not asserted by the requirements.

## Live validation (us-west-2): 22/23 — ONE DELIBERATE FAILURE (SDK parity gap, tracked in #617)
- ❌ **10-19**: Python's invocation infos expose no `operations` map and no externally-updated-operations collection (JS exposes both on every invocation hook). The dump omits those keys, failing the map-count assertions.
- ✅ 10-20..10-23 pass, plus the entire 10-1..10-18 regression.

### DIFF
--- a/packages/aws-durable-execution-sdk-python-conformance-tests/handlers/plugin/plugin_attempt_info_shape.py
+++ b/packages/aws-durable-execution-sdk-python-conformance-tests/handlers/plugin/plugin_attempt_info_shape.py
@@ -1,0 +1,118 @@
+"""10-21: Attempt hook info field shape (interface-shape probe).
+
+A single step named "flaky" throws on its first attempt and succeeds on the
+second, using the SDK's real retry strategy (``RetryStrategyConfig`` +
+``create_retry_strategy``, max_attempts >= 2, ~1s delay); it returns "ok". The
+plugin emits from the SDK's real per-attempt hooks (``on_user_function_start`` /
+``on_user_function_end``), filtering to step-type operations. Every logged field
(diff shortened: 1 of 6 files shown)


wangyb-A:
Rebased onto `plugin-parity-invocation-info` (#623) and retargeted; force-pushed `7537017` -> `81e6e3b`.

Rationale: the 10-19 handler dumps `operationsCount` / `updatedOperationsCount`, which read `info.operations` and `info.updated_operations` — fields that #623 adds. So the SDK parity fix has to land first, and this PR is now stacked on it. Two handler commits moved here from the (now closed) #622, where they had been incorrectly bundled with the SDK change.

The branch was previously based on a `main` that was 3 commits behind, so this also picks up #620/#621/#612. Handler file contents are byte-identical to before the rebase; the diff against the new base is the 6 handler files only, +646/-0. GitHub should auto-retarget this to `main` once #623 merges.

PR #622, opened 2026-08-10
feat(plugin): operation maps and payloads on invocation hooks

### LINKED ISSUE #617: Plugin invocation hook infos missing operations and updated-operations maps
## Summary
The plugin invocation-level hook info objects lack the operations-state surfaces the JS SDK exposes on every invocation hook.

**GA scope update:** execution input/result payload surfaces (`execution_input`, `execution_result`) have been moved OUT of GA conformance scope — they are experimental and no longer asserted (see aws/aws-durable-execution-conformance-tests#72). The remaining IN-SCOPE gap for this issue is the operations-state visibility below. The payload fields remain listed at the end as post-GA/experimental parity items.

## In-scope missing fields (verified against `src/aws_durable_execution_sdk_python/plugin.py`)
- **`operations`** — the map of checkpointed operations (JS: `operations` on `InvocationBaseInfo`, available at invocation-start AND invocation-end; convert via the existing `OperationInfo.from_operation`)
- **`updated_operations`** — operations externally updated since the previous invocation (JS: `InvocationInfo.updatedOperations`, from the invocation input's updated-operation ids; start hook)

## Evidence
Conformance requirement **plugin/10-19** (aws/aws-durable-execution-conformance-tests#72): Python fails exactly the three operations-map assertions (`operationsCount` at first start and on both end records, `updatedOperationsCount` at replay start) — live run 22/23, sole red is 10-19. Handler: `plugin_invocation_info_shape.py` (#615), a canonical dump — adding the fields flips it green with no test changes.

## Post-GA / experimental parity items (not currently asserted)
- `execution_input` (JS `executionInput`), `execution_result` (JS `InvocationEndInfo.executionResult`, raw value semantics)

## Suggested fix
Add the two map fields to the invocation info dataclasses, populated from the durable invocation input (operations + `UpdatedOperationIds`); consider lazy construction for operation-heavy executions. Conformance 10-19 flips green with no test changes.

wangyb-A:
Closes #617. Also implements #616.

Stacked on #615 (base is `plugin-hook-field-shape`) because the 10-19 handler
commit modifies `plugin_invocation_info_shape.py`, which only exists there.
Please merge #615 first; this should auto-retarget to `main`.

## What

Invocation-level plugin info objects carried no view of the execution's
operation state, and no view of the execution payloads. Python plugins could
not see what the JS SDK exposes on every invocation hook.

**#617 (GA scope)**
- `InvocationInfo.operations` — the checkpointed operation map converted to
  `OperationInfo`, on invocation-start **and** invocation-end. Mirrors JS,
  where it sits on `InvocationBaseInfo`.
- `InvocationStartInfo.updated_operations` — the subset named by the durable
  invocation input's `UpdatedOperationIds`, i.e. operations completed
  externally while the execution was suspended. Start hook only, matching JS's
  `InvocationInfo.updatedOperations`.

**#616 (experimental, out of GA conformance scope)**
- `InvocationInfo.execution_input` — the deserialized input event, the same
  object the durable handler receives.
- `InvocationEndInfo.execution_result` — the serialized result from the
  invocation output; `None` when the invocation suspended or failed.

All four fields are kw-only with empty/`None` defaults, so existing plugins,
hook constructors, and positional callers are unaffected.

## Notable decisions

- **The end hook re-reads the operation map** rather than reusing the
  start-of-invocation snapshot. This matches JS, which overrides
  `operations: toOperationInfoMap(...)` at every `onInvocationEnd` site, and it
  is strictly more informative — the suspending invocation reports 1 operation
  at start and 2 at end.
- **Snapshot eagerly, convert lazily.** #617 suggested lazy construction for
  operation-heavy executions. Fully lazy was wrong: a plugin that stashes the
  info and reads it later would observe a *later* state than its own hook. So
  the raw map is snapshotted when the hook fires (a cheap dict copy) and only
  the per-operation `OperationInfo` conversion is deferred and cached. A unit
  test pins this.
- `is_replayed` is left `False` on map entries, matching JS's
  `toOperationInfoMap` (`isReplay: false`) — these describe stored state, not
  replay events.
- The payload fields are marked `EXPERIMENTAL:` in their docstrings. Reaching
  any of this already requires the `plugins=` parameter, which emits a
  `FutureWarning`.
- The 10-19 handler deliberately does **not** dump `executionInput` /
  `executionResult`: the requirement's canonical schema omits them and asserts
  nothing about them. Its docstring was corrected accordingly.

## Testing

- Live plugin conformance suite, us-west-2: **23/23, exit 0**.
  **10-19 flipped PASSED with no changes to the requirement.** 10-1..10-18 and
  10-20..10-23 unaffected.
- Unit tests: 1492 -> 1498 passing; `plugin_test.py` 68 -> 81.
- `hatch fmt --check` clean.

### Known pre-existing flake (not from this change)

One earlier run of the suite showed 10-22 (`PluginOperationChangeShape`) red;
an immediate re-run on byte-identical code was 23/23 clean. Evidence it is not
a regression: the only edit on that path is a pure refactor (an inline dict
comprehension replaced by `_to_operation_info_map`, the same conversion), and
60/60 local in-process repeats of the 10-22 handler satisfy its primary
matcher. Two candidate causes remain (terminal-batch coalescing in the
checkpoint loop, or the runner's ~20s CloudWatch ingestion window); JS's
handler for that requirement is structurally identical, so it is likely
cross-SDK. Tracking separately.

### DIFF
--- a/packages/aws-durable-execution-sdk-python-conformance-tests/handlers/plugin/plugin_invocation_info_shape.py
+++ b/packages/aws-durable-execution-sdk-python-conformance-tests/handlers/plugin/plugin_invocation_info_shape.py
@@ -10,11 +10,11 @@ honest parity signal). One derived scalar is added on the end record:
 ``terminal`` := status in (SUCCEEDED, FAILED). No cross-hook reconstruction —
 ``isFirstInvocation`` on the end record comes from the invocation-end info.
 
-Python surface note: ``InvocationInfo`` carries only ``request_id``,
-``execution_arn``, ``is_first_invocation`` and ``execution_start_time``; the end
-info adds ``status`` + ``error``. It has NO execution-input, operations-map,
-externally-updated-operations, or execution-result field, so the canonical
-``executionInput`` / ``operationsCount`` / ``updatedOperationsCount`` /
+Python surface note: ``InvocationInfo`` carries ``request_id``,
+``execution_arn``, ``is_first_invocation``, ``execution_start_time`` and the
+``operations`` map; ``InvocationStartInfo`` adds ``updated_operations`` and the
+end info adds ``status`` + ``error``. It has NO execution-input or
+execution-result field, so the canonical ``executionInput`` /
 ``executionResult`` keys are omitted — the honest red for those probes.
 """
 
@@ -42,13 +42,15 @@ def _emit(record: dict[str, Any], execution_arn: str | None) -> None:
 
 class InvocationInfoShapePlugin(DurableInstrumentationPlugin):
     def on_invocation_start(self, info: InvocationStartInfo) -> None:
-        # Canonical dump of InvocationStartInfo. executionInput / operationsCount
-        # / updatedOperationsCount are absent from the Python type and therefore
-        # omitted — that omission is the parity signal under test.
+        # Canonical dump of InvocationStartInfo. executionInput is absent from
+        # the Python type and therefore omitted — that omission is the parity
+        # signal under test.
         record: dict[str, Any] = {
             "plugin": "CONFPLUGIN",
             "hook": "invocation-start",
             "isFirstInvocation": info.is_first_invocation,
+            "operationsCount": len(info.operations),
+            "updatedOperationsCount": len(info.updated_operations),
         }
         if info.request_id is not None:
             record["requestId"] = info.request_id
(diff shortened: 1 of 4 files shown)


wangyb-A:
Superseded by #623, which splits this correctly: #623 is the SDK-only parity fix against `main`, and the conformance handler changes move to the handlers PR (#615), stacked on #623.

PR #632, opened 2026-08-12
feat: add Workflow Insight instrumentation plugin

wangyb-A:
## Summary

Adds a **Workflow Insight instrumentation plugin** as a new package,
`packages/aws-durable-execution-sdk-python-insight/` — a port of the JS SDK's
`workflowInsight()` plugin (`aws-durable-execution-sdk-js-insight`), treated as the
reference implementation throughout. Experimental, matching the JS plugin's status.

It listens to the SDK's instrumentation hooks and emits one curated `WorkflowInsight`
record (`schemaVersion: "1.0"`) per execution. The wire record keeps the JS camelCase
field names so records read identically across SDKs and land in the same stores/queries.

### Behavior (mirrors the JS plugin)

- **Exporters:** `LambdaLogExporter` default (one JSON line to the function's log group,
  carrying the name-keyed `operationsByName` summary) and `S3Exporter` (the lossless
  per-occurrence `operations` array; upsert-by-execution-name; `none`/`date`/
  `function-name` partitioning). `boto3` is an extra (`[s3]`) since Lambda provides it.
- **Emit model:** `on-complete` / `on-failure` / `on-change`. Exporter calls run
  synchronously on the SDK checkpoint path today — there is no async scheduling or
  export coalescing yet (deferred/tracked in #687). Exports never propagate errors
  into the execution.
- **Sampling:** deterministic per-execution ARN hash; all-or-nothing per execution.
- **Content config:** input/output omission or transform (redaction), `include_errors`
  gating operation-level error detail only, per-operation result opt-in with optional
  transform.
- **Truncation:** phase 1 drops opted-in results oldest-first, phase 2 drops whole
  operations oldest-first, input/output last; per-exporter `max_record_size_bytes`
  measured against the exact shape each exporter emits.
- **Operation detail:** `top-level` (default; children with `parentId` suppressed) vs
  `full-tree`; unnamed operations are dropped (JS parity).

### Depends on #616 (merged)

The plugin reads `InvocationInfo.execution_input` / `InvocationEndInfo.execution_result`
introduced by #616 — the dependency floor is set to `>=1.8.0` accordingly (first release
that will carry those hooks). Capability note kept in the module docstring: the operations
map is reconstructed by accumulating per-operation hooks into per-execution state (keyed
by execution ARN to isolate warm-container reuse), since Python hooks carry no
end-of-invocation operations snapshot.

### Conformance validation (live, us-west-2)

Validated against the cross-SDK `insight` conformance suite
(aws/aws-durable-execution-conformance-tests#73, 18 requirements): **18/18 on the s3
sink and 18/18 on the cloudwatch sink**. Two known cross-SDK divergences are documented
in that suite rather than patched over here: operation ids pass through the SDK's native
`blake2b[:64]` format (JS uses `MD5[:16]`; the suite asserts ids as opaque), and the
per-operation `error.name` surfaces the customer error class while the record-level error
carries the SDK wrapper name (the suite asserts non-empty).

The suite's Python example handlers land in the conformance repo as a follow-up to #73
once this package is available.

## Testing

- 17 unit tests (`hatch run test:all packages/aws-durable-execution-sdk-python-insight/tests/`)
  covering record shaping, operations indexing, truncation phases, sampling, emit modes,
  and exporter rendering
- `hatch fmt` clean; package registered in the root `known-first-party`
- Live conformance runs as above (JS-parity behavior confirmed record-for-record)

### DIFF
--- a/.github/scripts/type-checks.sh
+++ b/.github/scripts/type-checks.sh
@@ -13,6 +13,10 @@ mypy --install-types --non-interactive \
   packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel \
   packages/aws-durable-execution-sdk-python-otel/tests
 
+mypy --install-types --non-interactive \
+  packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight \
+  packages/aws-durable-execution-sdk-python-insight/tests
+
 # comment out this for now as there are many type check errors in this package
 #mypy --install-types --non-interactive \
 #  packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing \
--- a/packages/aws-durable-execution-sdk-python-insight/pyproject.toml
+++ b/packages/aws-durable-execution-sdk-python-insight/pyproject.toml
@@ -1,0 +1,79 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "aws-durable-execution-sdk-python-insight"
+dynamic = ["version"]
+description = 'Workflow Insight instrumentation plugin for the AWS Durable Execution SDK for Python'
+readme = "README.md"
+requires-python = ">=3.11"
+license = "Apache-2.0"
+keywords = ["observability", "workflow-insight", "durable-execution"]
+authors = [{ name = "AWS durable-execution-dev", email = "durable-execution-dev@amazon.com" }]
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: Implementation :: CPython",
+]
+dependencies = [
+  # >=1.8.0: first release carrying the plugin invocation-hook fields
+  # (InvocationInfo.execution_input / InvocationEndInfo.execution_result).
+  "aws-durable-execution-sdk-python>=1.8.0",
+]
+
+[project.optional-dependencies]
+# boto3 is provided by the Lambda runtime; declared as an extra for local dev
+# (e.g. the S3Exporter) without vendoring it into deployments.
+s3 = ["boto3>=1.26.0"]
+
+[project.urls]
+Documentation = "https://github.com/aws/aws-durable-execution-sdk-python#readme"
+Issues = "https://github.com/aws/aws-durable-execution-sdk-python/issues"
+Source = "https://github.com/aws/aws-durable-execution-sdk-python"
+
+[tool.hatch.build.targets.sdist.force-include]
+"../../LICENSE" = "LICENSE"
+"../../NOTICE" = "NOTICE"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/aws_durable_execution_sdk_python_insight"]
+
+[tool.hatch.build.targets.wheel.force-include]
+"../../LICENSE" = "aws_durable_execution_sdk_python_insight/LICENSE"
+"../../NOTICE" = "aws_durable_execution_sdk_python_insight/NOTICE"
+
+[tool.hatch.version]
+path = "src/aws_durable_execution_sdk_python_insight/__about__.py"
+
+[tool.hatch.publish.index]
+disable = true
+
+[tool.coverage.run]
+source_pkgs = ["aws_durable_execution_sdk_python_insight"]
+branch = true
+parallel = true
+omit = ["src/aws_durable_execution_sdk_python_insight/__about__.py"]
+
+[tool.coverage.report]
+exclude_lines = ["no cov", "if __name__ == .__main__.:", "if TYPE_CHECKING:"]
+
+[tool.ruff]
+line-length = 88
+target-version = "py311"
+
+[tool.ruff.lint]
+preview = true
+select = ["E4", "E7", "E9", "F", "TID252"]
+
+[tool.ruff.lint.isort]
+known-first-party = ["aws_durable_execution_sdk_python_insight"]
+force-single-line = false
+lines-after-imports = 2
+
+[tool.ruff.lint.per-file-ignores]
+"tests/**" = ["ARG001", "ARG002", "ARG005", "S101", "PLR2004", "PLR6301", "SIM117", "TRY301"]
--- a/packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/exporters/s3_exporter.py
+++ b/packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/exporters/s3_exporter.py
@@ -1,56 +1,16 @@
 # SPDX-FileCopyrightText: 2026-present Amazon.com, Inc. or its affiliates.
 #
 # SPDX-License-Identifier: Apache-2.0
-"""First-party Workflow Insight exporters.
-
-Ports of the JS ``S3Exporter`` and ``LambdaLogExporter``. Both serialize the
-curated record with JS-compatible compact JSON (no whitespace) so the wire bytes
-match across SDKs. Records are written verbatim — no synthetic emission.
-"""
+"""S3 Workflow Insight exporter."""
 
 from __future__ import annotations
 
-import json
-import re
 from typing import Any
 
-from aws_durable_execution_sdk_python_insight.operations_index import (
-    with_operations_by_name,
+from aws_durable_execution_sdk_python_insight.exporters._common import (
+    compact_dumps,
+    sanitize,
 )
-
-
-def _dumps(value: Any) -> str:
-    return json.dumps(value, separators=(",", ":"), ensure_ascii=False)
-
-
-def _sanitize(value: str) -> str:
-    return re.sub(r"[^a-zA-Z0-9._-]", "_", value)
-
-
-class LambdaLogExporter:
-    """Writes ``operationsByName`` records to the function's own log group via ``print``.
-
-    Port of the JS ``LambdaLogExporter``: ``console.log(JSON.stringify(
-    withOperationsByName(record)))``. Requires no extra IAM. Emits the name-keyed
-    summary map (``OPERATIONS_BY_NAME``).
-    """
-
-    def __init__(self, max_record_size_bytes: int | None = None) -> None:
-        self.max_record_size_bytes = (
-            256_000 if max_record_size_bytes is None else max_record_size_bytes
-        )
-
-    def render(self, record: dict[str, Any]) -> dict[str, Any]:
-        return with_operations_by_name(record)
-
-    def export(self, record: dict[str, Any]) -> None:
-        # Raw JSON line to stdout -> the function's CloudWatch log group. The
-        # conformance CloudWatch sink json.loads each line (and unwraps the
-        # Lambda structured-log envelope when present).
-        print(_dumps(self.render(record)), flush=True)  # noqa: T201
-
-    def flush(self) -> None:
-        return None
 
 
 class S3Exporter:
@@ -93,7 +53,7 @@ class S3Exporter:
         self._client.put_object(
             Bucket=self.bucket,
             Key=key,
-            Body=_dumps(record).encode("utf-8"),
+            Body=compact_dumps(record).encode("utf-8"),
             ContentType="application/json",
         )
 
@@ -102,7 +62,7 @@ class S3Exporter:
 
     def _build_key(self, record: dict[str, Any]) -> str:
         file_name = (
-            _sanitize(
+            sanitize(
                 record.get("executionName") or record.get("executionArn") or "record"
             )
             + ".json"
@@ -111,7 +71,7 @@ class S3Exporter:
 
     def _partition(self, record: dict[str, Any]) -> str:
         if self.partitioning == "function-name":
-            return f"function={_sanitize(record.get('functionName', ''))}/"
+            return f"function={sanitize(record.get('functionName', ''))}/"
         if self.partitioning == "date":
             start = str(record.get("startTime", ""))
             # YYYY-MM-DD... -> year=YYYY/month=MM/day=DD/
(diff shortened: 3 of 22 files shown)


wangyb-A:
Note to reviewer: I am tracking the AI reviewer's comments in an issue: https://github.com/aws/aws-durable-execution-sdk-python/issues/687
And will have following PRs to fix them.
Will not include the full change in this PR, for easier review

packages/aws-durable-execution-sdk-python-insight/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+s3 = ["boto3>=1.26.0"]
+
+[project.urls]
+Documentation = "https://github.com/aws/aws-durable-execution-sdk-python#readme"
+Issues = "https://github.com/aws/aws-durable-execution-sdk-python/issues"
+Source = "https://github.com/aws/aws-durable-execution-sdk-python"
+
+[tool.hatch.build.targets.sdist.force-include]
+"../../LICENSE" = "LICENSE"
+"../../NOTICE" = "NOTICE"
```
zhongkechen, line 41: add a plugin entry point here to allow this plugin to be auto loaded. See otel plugin for reference

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+        )
+        if input_value is not None:
+            record["input"] = input_value
+        if output_value is not None:
+            record["output"] = output_value
+        if error is not None:
+            record["error"] = {"name": error.type, "message": error.message}
+        record["operations"] = self._build_operations(operations)
+
+        for exporter in self._exporters:
```
ParidelPooya, line 430: **`flush()` is never called.**

`InsightExporter` declares `flush()`, both shipped exporters implement it, and the JS plugin calls `flushAll(exporters)` in `wrapInvocation`. Nothing in this package ever invokes it — `grep -rn flush src/` returns only definitions.

Both first-party exporters are unbuffered, so this is latent today, but the protocol advertises a lifecycle contract the plugin doesn't honor: any customer exporter that batches will silently drop records. The Python SDK has no `wrap_invocation` hook, so the natural place is right after the terminal export here (or at the end of `on_invocation_end`).

Alternatively, if flushing is deliberately out of scope, make `flush()` optional in the protocol rather than required.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+            self._emit(
+                arn,
+                state,
+                status="RUNNING",
+                end_time=None,
+                output_raw=None,
+                error=None,
+            )
+
+    def on_operation_change(self, info: OperationChangeInfo) -> None:
```
ParidelPooya, line 261: **No export scheduler — and the PR description claims one.**

The description says "emit model … with export coalescing — a newer record supersedes a pending one." There is no `ExportScheduler` equivalent here; `_emit` calls `exporter.export()` inline. In `on-change` mode that has two consequences:

1. `OperationChangeInfo` is dispatched by the SDK with `sync=True` (`PluginExecutor.execute_plugins`, `plugin.py:823`), i.e. inline on the checkpoint-processing thread. A blocking `put_object` here stalls the SDK's checkpoint pump for every S3 round trip.
2. Every operation status change becomes one `PutObject` to the *same* key, with no coalescing. A 200-operation execution issues ~200 writes where JS would issue far fewer.

I checked for an out-of-order/overwrite hazard and there isn't one: `on_invocation_start` runs before the checkpoint thread starts, changes are serialized on the single checkpoint thread, and `on_invocation_end` fires from `handle_durable_output` after the `ThreadPoolExecutor(max_workers=2)` block has joined both workers. So the terminal record is always written last. The issues are latency and write amplification, not correctness.

Either implement coalescing or correct the description — right now the two disagree.

Separately: this hook calls `_ensure_state` and `_adopt_operations` unconditionally, whereas JS returns early unless `emitMode === "on-change"`. In the default `on-complete` mode that work is pure waste, since `on_invocation_end` re-adopts a fresh snapshot anyway. Cheap (shallow copy, small next to the SDK's own eager `_to_operation_info_map`), but an early return is strictly better.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _should_sample(execution_arn: str, rate: float) -> bool:
+    if rate >= 1:
+        return True
+    if rate <= 0:
+        return False
+    return _fnv1a32(execution_arn) / 0xFFFFFFFF < rate
+
+
+def _resolve_sampling_rate(rate: float | None) -> float:
```
ParidelPooya, line 105: **`sampling_rate=NaN` silently disables all instrumentation.**

NaN falls through every guard here (`nan < 0` and `nan > 1` are both `False`), and then `_should_sample` returns `False` for every ARN. Verified:

```
rate_in=nan       resolved=nan   sampled_in=False
rate_in='0.5'     resolved=1.0   sampled_in=True
```

JS handles this explicitly — `Number.isNaN(rate)` → `1.0` plus a `console.warn`. Fail-open on misconfiguration is the safer default, and the current fail-closed behavior is completely invisible: no records, no warning.

Related: the string case silently coerces to `1.0` with no warning either. That sits oddly next to `EmitMode` / `OperationDetail` / `S3Partitioning`, which all raise `ValueError` on a bad dynamic value. Worth picking one philosophy for the config surface.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+    return None
+
+
+def _duration_ms(start: Any, end: Any) -> int | None:
+    if isinstance(start, datetime.datetime) and isinstance(end, datetime.datetime):
+        return int((end - start).total_seconds() * 1000)
+    return None
+
+
+def _apply_data_content(value: Any, setting: Any) -> Any:
```
ParidelPooya, line 127: **A `null` execution output is dropped.**

The early return on `value is None` collapses JS's `undefined`-vs-`null` distinction. A handler returning `None` yields `execution_result == "null"` → `json.loads` → `None` → `output` omitted from the record, where JS emits `output: null`.

Cosmetic, but it is a wire divergence in exactly the schema the cross-SDK conformance suite compares.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+        # plugin is never a silent no-op. A non-empty list is used verbatim.
+        self._exporters: list[InsightExporter] = (
+            list(config.exporters) if config.exporters else [LambdaLogExporter()]
+        )
+        self._state: dict[str, _ExecutionState] = {}
+        self._lock = threading.Lock()
+
+    # -- sampling / state -----------------------------------------------------
+
+    def _sampled_in(self, execution_arn: str) -> bool:
```
ParidelPooya, line 205: Sampling is recomputed on every hook — this re-hashes the ~120-char ARN each time. JS computes it once and caches it in `ExecutionState`. Harmless, but the cache is free and the state object already exists.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/plugin.py
```diff
@@ -0,0 +1,449 @@
...
+            record["error"] = {"name": error.type, "message": error.message}
+        record["operations"] = self._build_operations(operations)
+
+        for exporter in self._exporters:
+            try:
+                shaped = truncate_record(
+                    record, exporter.max_record_size_bytes, exporter.render
+                )
+                exporter.export(shaped)
+            except Exception as exc:  # noqa: BLE001 - one exporter must not break others / the execution
```
ParidelPooya, line 436: A library should use `logging` rather than `print` to stderr; the SDK already has a module logger and customers can't filter or route this.

Also worth knowing: the SDK's `_dispatch_plugin` already catches and logs exceptions escaping any hook, so this broad `except` is belt-and-braces rather than the only line of defense.

packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight/exporters/s3_exporter.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _partition(self, record: dict[str, Any]) -> str:
+        if self.partitioning == S3Partitioning.FUNCTION_NAME:
+            return f"function={sanitize(record.get('functionName', ''))}/"
+        if self.partitioning == S3Partitioning.DATE:
+            start = str(record.get("startTime", ""))
+            # YYYY-MM-DD... -> year=YYYY/month=MM/day=DD/
+            if len(start) >= 10 and start[4] == "-" and start[7] == "-":
+                return f"year={start[0:4]}/month={start[5:7]}/day={start[8:10]}/"
+            return ""
```
ParidelPooya, line 104: When `startTime` is absent or malformed, date partitioning silently degrades to no partition and the object lands at the prefix root, mixing unpartitioned objects into an otherwise partitioned layout — awkward for Athena partition projection.

This is *better* than JS, which produces `year=NaN/month=NaN/day=NaN/`, but a `year=unknown/` sentinel would keep the layout uniform.

packages/aws-durable-execution-sdk-python-insight/tests/test_plugin.py
```diff
@@ -0,0 +1,570 @@
...
+def test_on_failure_success_emits_nothing():
+    exporter = CaptureExporter()
+    plugin = workflow_insight(
+        WorkflowInsightConfig(exporters=[exporter], emit_mode="on-failure")
+    )
+    _run(plugin, ops=[_step("greet")], status=InvocationStatus.SUCCEEDED)
+    assert exporter.records == []
+
+
+def test_sampling_zero_emits_nothing():
```
ParidelPooya, line 190: **Sampling's deterministic hash is never executed.** Coverage shows `plugin.py:90` (`_fnv1a32` body) unhit — this is the only sampling test and rate 0 short-circuits before the hash. That leaves the one piece that must agree with JS record-for-record completely untested.

Worth adding:
- a vector test (`_fnv1a32(known_arn) == <value computed by the JS implementation>`) to lock in cross-SDK parity;
- a fractional-rate test asserting the same ARN yields the same decision across repeated calls / a fresh plugin instance;
- the clamp and non-numeric paths in `_resolve_sampling_rate` (`plugin.py:109`, `111`).

Three other gaps in this file, all cheap:
- `OperationOverride.exclude` is untested (`plugin.py:339` unhit) — a documented config knob with zero coverage.
- Per-operation `error` inclusion is untested (`plugin.py:360` unhit); only the `include_errors=False` path is covered.
- Exporter-failure isolation is untested (`plugin.py:436` unhit). "One exporter must not break others / the execution" is a core safety claim and a throwing-exporter test is a few lines.

packages/aws-durable-execution-sdk-python-insight/tests/test_shaping.py
```diff
@@ -0,0 +1,132 @@
...
+    record["operations"][2].pop("result", None)
+    out = truncate_record(record, 480, render=lambda r: r)
+    assert out["truncated"] is True
+    assert out.get("droppedOperations", 0) >= 1
+    names = {op["name"] for op in out["operations"]}
+    assert "bulk-1" not in names  # oldest dropped
+    assert "bulk-3" in names  # newest retained
+
+
+def test_truncation_noop_when_within_limit():
```
ParidelPooya, line 129: **Truncation phase 3 is untested.** `truncation.py:100–105` and the `pop` lines 68/71 are unhit, so `droppedInput` / `droppedOutput` — documented in both the README and the PR description — are never exercised. Phases 1 and 2 are covered; a case where dropping every operation still leaves the record over the limit would close this out.

.github/scripts/type-checks.sh
```diff
@@ -13,6 +13,10 @@ mypy --install-types --non-interactive \
   packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel \
   packages/aws-durable-execution-sdk-python-otel/tests
 
+mypy --install-types --non-interactive \
+  packages/aws-durable-execution-sdk-python-insight/src/aws_durable_execution_sdk_python_insight \
```
ParidelPooya, line 17: Two CI-wiring omissions while you're in here:

1. `ci.yml`'s "Verify legal files in published distributions" step takes an explicit package list (core / otel / testing) and doesn't include insight, even though this package declares `force-include` for LICENSE and NOTICE in both sdist and wheel targets. I ran `check_dist_legal_files.py` against the built package manually and it passes — just add it to the list so the contract stays enforced.
2. The repo convention of a per-package `[tool.hatch.envs.dev-*]` env plus an entry in `.github/scripts/ci-checks.sh` (present for core, otel, testing, examples) wasn't followed, so the local dev script skips insight entirely.

`ci.yml`'s fmt and build steps loop over `packages/*/`, so those are already covered.

ParidelPooya (commented):
Reviewed locally at `953e66f`. Verified: 60 tests pass (~1s), `hatch run types:check` clean, `hatch fmt --check` clean for this package, wheel/sdist build with LICENSE+NOTICE correctly placed.

Solid, careful port. `truncation.py` and `operations_index.py` are near line-for-line faithful to the JS reference, the ARN parsing and FNV-1a math match, and sourcing operations from the SDK's authoritative `operations` snapshots (rather than accumulating `on_operation_end` events) is the right call — the e2e suspend/resume test proves it survives a cold resume. The comments are unusually good at explaining *why*.

Concerns are concentrated in one behavioral gap plus a few correctness edges and test holes. Three I'd want addressed before merge:

1. `flush()` is declared on the exporter protocol but never called anywhere.
2. There is no export scheduler / coalescing, contrary to the PR description, and exports run inline on the checkpoint thread in `on-change` mode.
3. `sampling_rate=NaN` silently disables all instrumentation (JS fails open to 1.0 with a warning).

Details inline.

ParidelPooya (approved):

PR #633, opened 2026-08-13
ci: split Lambda layer publish credentials by region

zhongkechen:
## Summary
- resolve Lambda layer publish targets into credential-scoped matrix jobs
- use one role ARN secret per opt-in commercial Region, one China secret for both China Regions, and one GovCloud secret for both GovCloud Regions
- configure the China OIDC audience, validate manual Region overrides, and document the required secret names
- retain the universal layer artifact introduced on current `main`

## Testing
- `python -m pytest .github/scripts/tests/test_resolve_layer_publish_targets.py .github/scripts/tests/test_lambda_layer_publish_workflow.py .github/scripts/tests/test_build_lambda_layer.py .github/scripts/tests/test_parse_sdk_branch.py .github/scripts/tests/test_resolve_layer_sdk_version.py` (18 passed)
- Ruff lint and format checks
- `actionlint .github/workflows/lambda-layer-publish.yml .github/workflows/test-parser.yml`


zhongkechen:
moved to one account per partition strategy

Issue #456, opened 2026-06-10
[Testing] Enhance get_assertable_operations method


dhegberg:
Migration from https://github.com/aws/aws-durable-execution-sdk-python-testing/issues/48

---

## Context
Add logic to handle optional EXECUTION operation at end of execution

## Technical Requirements
- Add logic to handle optional EXECUTION operation at end of execution
- Implement proper detection of final EXECUTION operations with large payloads
- Add context-aware inclusion/exclusion of final EXECUTION operations
- Ensure immutability of original execution state

## Implementation Details
- **File**: `src/aws_durable_execution_sdk_python_testing/execution.py:111`

## Acceptance Criteria
- [ ] Logic added to handle optional EXECUTION operation at end of execution
- [ ] Proper detection implemented for final EXECUTION operations with large payloads
- [ ] Context-aware inclusion/exclusion added for final EXECUTION operations
- [ ] Immutability of original execution state ensured

PR #637, opened 2026-08-13
fix(testing): filter EXECUTION ops by type, not index

SravyaPeri:
*Issue #, if available:* #456

*Description of changes:*

While looking into this I found get_assertable_operations() had a TODO on it questioning its own logic, it excludes the EXECUTION operation by just slicing off `operations[1:]`, assuming that operation is always first. That's true today, but it leaves a couple of things unhandled:

- If a second EXECUTION-type entry ever shows up at the end of the list (which is literally what the TODO was asking about, something like a future large-payload checkpoint), it would slip through and end up in the "assertable" results by mistake.
- This method is also the only place in the class that reads `self.operations` without grabbing `_state_lock` first. Every other method that touches operations (`get_navigable_operations`, `start`, `_end_execution`, the callback completions, etc.) does take the lock, so this one stood out. Since checkpoints can land on a worker thread while this is being read, it felt worth fixing rather than leaving as a latent race.

What I changed:
- Filter operations by `operation_type != OperationType.EXECUTION` instead of slicing by index, so it excludes any EXECUTION-type entry no matter where it shows up.
- Added the `_state_lock` around the read, matching how `get_navigable_operations()` already does it.
- Added a couple of tests: one for a trailing EXECUTION operation getting excluded correctly, and one checking the returned list doesn't alias self.operations (so mutating the execution afterward doesn't change a result you already got back).

Ran the full suite plus lint/type checks locally, all green:
hatch run dev-testing:test        # 1379 passed
hatch fmt --check                 # clean
hatch run dev-testing:typecheck   # clean

### DIFF
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/execution.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/execution.py
@@ -344,9 +344,21 @@ class Execution:
             return list(self.operations)
 
     def get_assertable_operations(self) -> list[Operation]:
-        """Get list of operations, but exclude the EXECUTION operations"""
-        # TODO: this excludes EXECUTION at start, but can there be an EXECUTION at the end if there was a checkpoint with large payload?
-        return self.operations[1:]
+        """Return a snapshot of operations, excluding EXECUTION operations.
+
+        Filters by ``operation_type`` rather than assuming the EXECUTION
+        operation is always the first entry, so a second EXECUTION-type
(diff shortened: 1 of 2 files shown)


yaythomas:
the extra `_state_lock` is actually not necessary, and if anything we could/should _probably_ remove it from get_navigable and friends, reason being the recent lane-based re-architecture isolated an execution into its own lane that serializes Checkpoint writes that run as CheckpointTask on that same lane, so the read and write cannot overlap.

however, that's wider refactor beyond the scope fo this PR, and since the lock is harmless and for the sake of symmetry might as well let it stand as introduced here and then revisit clean-up opportunities for redundant locks throughout later.

yaythomas (approved):

PR #655, opened 2026-08-18
Docs/migration 1x to 2x

ayushiahjolia:
*Issue #, if available:* N/A

*Description of changes:*
Migration guide for python v2


docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,152 @@
...
+`throw_if_error()` as a `ChildContextError`, so there is still no separate
+batch-completion error type to catch. Notes:
+
+- It cannot be combined with `min_successful` or the `tolerated_failure_*`
+  fields; doing so raises `ValidationError` at construction.
+- The predicate must be deterministic and side-effect-free. Replay uses the
+  checkpointed decision and never re-invokes it.
+- New exports: `complete_batch`, `continue_batch`, `CompletionStatus`,
+  `CompletionDecision`, `CompletionOutcome`, `CompletionItemStatus`,
+  `BatchItemStatus`.
```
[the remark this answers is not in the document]
ayushiahjolia: Will merge migration PR after https://github.com/aws/aws-durable-execution-sdk-python/pull/605

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,152 @@
...
+
+`wait_for_condition` also round-trips `initial_state` through the serdes before
+the first check, so `initial_state` must now be serializable by the configured
+serdes.
+
+## New in 2.x: Custom Completion Predicate (Optional)
+
+`2.x` adds a `should_complete` predicate to `CompletionConfig`, giving `map` and
+`parallel` full control over when a batch completes early. This is a new feature,
+not a breaking change - no action is required unless you adopt it.
```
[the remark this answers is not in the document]
ayushiahjolia: https://github.com/aws/aws-durable-execution-sdk-python/pull/605

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,156 @@
...
+
+```python
+from aws_durable_execution_sdk_python import complete_batch, continue_batch
+
```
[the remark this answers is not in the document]
ayushiahjolia: will merge this pr after https://github.com/aws/aws-durable-execution-sdk-python/pull/605

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,156 @@
...
+`throw_if_error()` as a `ChildContextError`, so there is still no separate
+batch-completion error type to catch. Notes:
+
+- It cannot be combined with `min_successful` or the `tolerated_failure_*`
+  fields; doing so raises `ValidationError` at construction.
+- The predicate must be deterministic and side-effect-free. Replay uses the
+  checkpointed decision and never re-invokes it.
+- New exports: `complete_batch`, `continue_batch`, `CompletionStatus`,
+  `CompletionDecision`, `CompletionOutcome`, `CompletionItemStatus`,
+  `BatchItemStatus`.
```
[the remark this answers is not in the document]
ayushiahjolia: https://github.com/aws/aws-durable-execution-sdk-python/pull/605

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,156 @@
...
+| `CallbackError` moved out of the termination tree; graded subtypes added | Remove any `termination_reason == TerminationReason.CALLBACK_ERROR` check (the enum member is gone). Optionally catch `CallbackTimeoutError` / `CallbackExternalError` / `CallbackSubmitterError`. |
+| `BatchResult.throw_if_error()` now raises a typed error | Catch `ChildContextError` instead of `CallableRuntimeError`. |
+| First-run serialize/deserialize round trip for `step`, child contexts, `map`/`parallel`, and `wait_for_condition` | Make custom `SerDes` round-trip safe: `deserialize(serialize(x)) == x`. Ensure `wait_for_condition` `initial_state` is serializable by the configured serdes. For a transient serdes failure, raise the new `RetryableSerDesError` (retries) instead of `SerDesError` (permanent). |
+| `InvokeConfig.timeout` and `InvokeConfig.timeout_seconds` removed | Remove them. Enforce any timeout inside the invoked function or as a separate timer. |
+| Removed `ItemBatcher`, `ItemsPerBatchUnit`, `BatchedInput`, `TerminationMode`, `StepFuture`, `MapConfig.item_batcher`, `ChildConfig.item_serdes` | Remove all uses. Replace `ChildConfig.item_serdes` with `ChildConfig.serdes`. |
```
yaythomas: Missing removed names. `ChainedInvokeFailedToStartType`, `ChainedInvokeTimeoutType`, and `ChainedInvokeStopType` were also removed from `lambda_service`.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,156 @@
...
+| `MapConfig` / `ParallelConfig` / `CompletionConfig` now validate at construction | Wrap construction in `try/except ValidationError` if you build configs from external input. |
```
yaythomas: This row covers two items from fdc952b but misses four.

1. `BatchResult.all` now omits never-started branches. `total_count` and index iteration change for early-completed batches.
2. Custom `summary_generator` output is now stored under a `"summary"` key inside an SDK-owned envelope. It no longer replaces the payload. Code parsing summary payloads from execution history sees a different shape.
3. `max_concurrency=0` and `min_successful=0` now raise `ValidationError`.
4. `min_successful > total` is validated at the `.map()` or `.parallel()` call, not at `CompletionConfig()` construction. The row text says "at construction", which is only partly true.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,156 @@
...
+```python
+result = context.map(items, process_item)
+try:
+    result.throw_if_error()           # raises ChildContextError for the first failure
+except ChildContextError:
+    for err in result.get_errors():   # every failed item's ErrorObject
+        context.logger.error("%s: %s", err.type, err.message)
+```
+
+## Serialize/Deserialize Round Trip
```
yaythomas: Missing a catch-path break. In 1.x, serdes failures raised `ExecutionError`. In 2.x they raise `SerDesError`, which inherits from `DurableExecutionsError` and not from `ExecutionError`. Existing `except ExecutionError` blocks silently stop catching serialization failures. Needs a table row.

docs/migration-1.x-to-2.x.md
yaythomas: Missing?

1. `StepContext.attempt` and `WaitForConditionCheckContext.attempt` (6aa6a02).
2. Plugin auto-discovery and `PluginLoadError`, now exported from the package root (808e9a8).

yaythomas (changes requested):
Is the idea to link this guide from the README the _tagged_ guide from the 2.0 GitHub release notes?

The plugin interface also changed incompatibly between 1.x and 2.x. Hook signatures changed, enums moved, and InvocationEndInfo.status became required. The interface is experimental, so it may be out of scope for this guide? If so, a single sentence saying the experimental plugin API changed would still save plugin authors a surprise.

docs/migration-1.x-to-2.x.md
ayushiahjolia: 1. Adding
2. Intentionally didn't add any plugin related changes.

ayushiahjolia:
> Is the idea to link this guide from the README the _tagged_ guide from the 2.0 GitHub release notes?
> 
> The plugin interface also changed incompatibly between 1.x and 2.x. Hook signatures changed, enums moved, and InvocationEndInfo.status became required. The interface is experimental, so it may be out of scope for this guide? If so, a single sentence saying the experimental plugin API changed would still save plugin authors a surprise.

yes, I'll link this guide in README. 
I'll add a line for plugin changes.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,217 @@
...
+| `BatchResult.all` now omits never-started branches, so `total_count` and positional iteration differ for early-completed batches | If you index `.all` by original position or expect `total_count` to include unstarted branches, update that logic. |
+| For `map`/`parallel`, a custom `summary_generator` output is now stored under a `"summary"` key in an SDK-owned envelope (it no longer replaces the checkpoint payload). `ChildConfig.summary_generator` is unchanged: its output is still checkpointed verbatim | If you parse `map`/`parallel` summary payloads from execution history, read the `"summary"` key from the envelope. Child-context summary consumers need no change. |
+| `WaitDecision` removed; `WaitStrategyConfig.timeout` / `timeout_seconds` removed | Use `WaitForConditionDecision` (`stop_polling()` / `continue_waiting(delay)`). |
+| `wait_for_condition` raises `WaitForConditionError` when it exhausts `max_attempts` | Catch `WaitForConditionError` instead of inspecting the returned state. |
+
```
yaythomas: empy str payloads?  1.x dropped them when serializing; 2.x preserves them (b448f3e). A workflow that observed None in 1.x can observe "" in 2.x.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,217 @@
...
+value your callback deserializer returns must itself be serializable by the
+default serdes; otherwise the child raises `SerDesError`.
+
+`wait_for_condition` also round-trips `initial_state` through the serdes before
+the first check, so `initial_state` must now be serializable by the configured
+serdes. Custom serdes should serialize polling state to a non-empty string; an
+empty string checkpoint payload is currently treated as no stored polling state
+on resume, so the operation may restart from `initial_state`.
+
+## New in 2.x: Custom Completion Predicate (Optional)
```
zhongkechen: New features belong elsewhere, don't they?

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,217 @@
...
+are gone with no alias. If you are not ready to migrate, stay on `1.x`.
+
+> Instrumentation plugins are out of scope here. The experimental `plugins=` hook
+> already existed in `1.x`; `2.x` adds opt-in auto-discovery and `PluginLoadError`.
+> Because plugins are opt-in and still evolving, they are documented with the
+> plugin/OpenTelemetry feature rather than in this guide.
+
+## What Changed and What to Do
+
+| Change | What you must do |
```
yaythomas: To help w seo and instead of table (formatting on table for longer text can be limiting)... taking example from https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md or https://docs.python.org/3/whatsnew/3.12.html#porting-to-python-3-12, how about a 

```
## Porting to 2.0
### CallableruntimeErrror removed
<<before>>.
<<after>>
Catch `StepError`, `InvokeError`, `ChildContextError`, or `WaitForConditionError` (or the base `DurableOperationError`) instead of `CallableRuntimeError`. 

### Remove x
<<before>>
<<after>>
```

this also makes it more feasible to have meaningful little code snippets in triple backticks etc.

docs/migration-1.x-to-2.x.md
ayushiahjolia: Hmm I am going to use this doc for release notes for v2, that's why I added all v2 changes/features here.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,364 @@
...
+> Instrumentation plugins are out of scope here. The experimental `plugins=` hook
+> already existed in `1.x`; `2.x` adds opt-in auto-discovery and `PluginLoadError`.
+> The plugin interface itself also changed incompatibly (hook signatures changed,
+> enums moved, and `InvocationEndInfo.status` is now required), so plugin authors
+> should expect to update. Because plugins are opt-in and still evolving, they are
+> documented with the plugin/OpenTelemetry feature rather than in this guide.
+
+## Porting to 2.0
+
+Each change below lists what changed and what you must do. The ones most likely
```
yaythomas: "what you must do" reads demanding. Suggest "what to do", matching the tone of the section headings.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,364 @@
...
+`""` explicitly.
+
+### `InvokeConfig.timeout` / `timeout_seconds` removed
+
+Both fields are gone.
+
+Remove them. Enforce any timeout inside the invoked function or as a separate
+timer.
+
+### `map` / child batching names removed
```
yaythomas: The `ChainedInvoke*` enums (added per https://github.com/aws/aws-durable-execution-sdk-python/pull/655#discussion_r3834502302) are not batching names, so this heading mislabels them. "Removed 1.x-only names" would cover everything in the section.

yaythomas (dismissed):

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,367 @@
+# Migrating from 1.x to 2.x
+
+`2.x` is a breaking major release. Every change is a bug fix or brings Python to
+parity with the JavaScript and Java SDKs. The changes most likely to touch your
+code are the typed, per-operation **error hierarchy**, the
+**serialize/deserialize round trip on the first run**, and the new fail-fast
+**default `map` / `parallel` completion** (which completes after the first
+observed failure and stops scheduling pending items; already-started items are
+not cancelled, so with unlimited concurrency all items may already be running -
+set `max_concurrency` if you need to bound that). Opt back into process-all with
```
zhongkechen: `max_concurrency` behavior changed. Seems we didn't mention it

yaythomas:
picking up this PR to incorporate https://github.com/aws/aws-durable-execution-sdk-python/pull/698

docs/migration-1.x-to-2.x.md
yaythomas: Addressed in 38a9158. The intro line is removed in the restructure; the remaining prose says "what to do".

docs/migration-1.x-to-2.x.md
yaythomas: Addressed in 38a9158. Heading renamed to "Removed 1.x-only names".

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,367 @@
...
+
+`step`, child contexts, `map`/`parallel`, and `wait_for_condition` now round-trip
+their result through the serdes on the first run, returning
+`deserialize(serialize(x))` - the same canonical value replay returns.
+
+If you relied on the raw pre-serialization object, use the deserialized shape
+instead (or make your `SerDes` round-trip identity). Ensure `wait_for_condition`
+`initial_state` is serializable by the configured serdes. For a transient serdes
+failure, raise the new `RetryableSerDesError` (retries) instead of `SerDesError`
+(permanent). See [Serialize/Deserialize round trip](#serializedeserialize-round-trip).
```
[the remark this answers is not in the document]
yaythomas: Addressed in 38a9158. The round-trip section now explains that RetryableSerDesError replays the whole invocation, that an AT_LEAST_ONCE_PER_RETRY step re-runs its body and repeats side effects, and that an at-most-once step follows interrupted-step handling.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,367 @@
...
+`WaitDecision` is gone, along with `WaitStrategyConfig.timeout` and
+`WaitStrategyConfig.timeout_seconds`.
+
+Use `WaitForConditionDecision` (`stop_polling()` / `continue_waiting(delay)`).
+
+### `wait_for_condition` raises on exhaustion
+
+It raises `WaitForConditionError` when it exhausts `max_attempts`.
+
+Catch `WaitForConditionError` instead of inspecting the returned state.
```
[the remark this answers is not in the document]
yaythomas: Addressed in 38a9158. Section renamed to "create_wait_strategy raises on exhaustion" and now states that a custom strategy must enforce its own limit.

docs/migration-1.x-to-2.x.md
```diff
@@ -0,0 +1,367 @@
...
+`StepError`, `InvokeError`, `ChildContextError`, `WaitForConditionError`,
+`CallbackError` (+ `CallbackExternalError`, `CallbackTimeoutError`,
+`CallbackSubmitterError`), plus `SerDesError` (now exported) and
+`RetryableSerDesError`. `SerDesError` stays a direct child of
+`DurableExecutionsError`; `RetryableSerDesError` is a retryable `InvocationError`.
+
+### Callbacks
+
+`context.wait_for_callback(...)` returns the payload directly and raises the
+callback error from the call itself (there is no `callback.result()`):
```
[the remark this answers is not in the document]
yaythomas: Addressed in 38a9158. The callbacks section now scopes the statement to the wait_for_callback return value and keeps create_callback().result() documented.

docs/migration-1.x-to-2.x.md
yaythomas: Addressed in 38a9158. New section "max_concurrency bounds in-flight branches": 1.x capped worker threads, 2.0 caps in-flight branches, a suspended branch holds its slot, and the parent suspends when all slots are held.

ParidelPooya:
The guide is unusually accurate — I checked ~25 factual claims and confirmed the large majority, including several that are easy to get wrong (__cause__ holding a reconstructed stand-in, empty-string payload preservation, BatchResult.all dropping PENDING branches, wait_for_callback's child context ignoring WaitForCallbackConfig.serdes, the
   empty-payload-means-no-stored-state quirk in wait_for_condition). Ruff passes on the waits.py docstring change.

   But there is one blocking error, and it's in the section the guide itself promotes as one of the top three changes.

   Blocking: the all_completed() section gets 1.x backwards

   The guide says:

   │ the default map and parallel completion config runs fail-fast … Pass all_completed() explicitly to keep the 1.x process-all behavior.

   Both halves are wrong about 1.x:

   The fail-fast default is not new. In 1.x, MapConfig.completion_config already defaulted to a bare CompletionConfig() and ParallelConfig to all_successful() — identical defaults to 2.0. And 1.x explicitly fail-fast on an empty config (concurrency/models.py at sdk-v1.7.0):

   if not has_any_completion_criteria:
       # Empty completion config - fail fast on any failure
       if failure_count > 0:
           return CompletionReason.FAILURE_TOLERANCE_EXCEEDED

   with the scheduler agreeing via should_continue() → return self.failure_count == 0. 2.0's is_tolerance_exceeded reduces to the same failed > 0. Nothing changed here, yet it is listed as a breaking change in both the opening paragraph and the "Porting to 2.0" summary.

   There was no 1.x process-all behavior to keep. 1.x's all_completed() returned all-None, which hit the fail-fast branch above. 2.0's returns tolerated_failure_percentage=100. So the change is a bug fix in the opposite direction: all_completed() used to fail fast and now actually tolerates every failure. Anyone who wrote all_completed() in 1.x and got fail-fast
   will silently start processing every item after upgrading — the real migration risk, and the guide's advice would lead them straight into it.

   Suggested reframing: retitle to something like "all_completed() now actually tolerates all failures", drop the default-completion claim from the intro and the porting summary, and tell readers who relied on 1.x all_completed() to switch to all_successful() or a bare CompletionConfig().

   Should fix

   should_complete invocation timing is wrong. The guide says the predicate "runs before the SDK schedules any branch (completed_count == 0) and on each suspension state change." In concurrency/executor.py, needs_snapshot_rebuild = True is set on branch submission and on COMPLETED, FAILED, SUSPENDED, and SUSPENDED_UNTIL events. Terminal completion and failure
   are the primary triggers and the whole point of a quorum predicate, but the guide's wording implies they don't fire it. CompletionStatus's own docstring says "on each branch terminal event". Should read "on every branch state change (scheduling, completion, failure, and suspension)".

   Wrong link for the plugin section. It points at /durable-execution/sdk-reference/observability/logging/, which documents the replay-aware logger and says nothing about the plugin interface. There is a dedicated page at /durable-execution/sdk-reference/observability/plugins/ that covers exactly the hooks and InvocationEndInfo changes @yaythomas asked about.

   The SerDesError section's stated reason is wrong, though its advice is right. The guide says SerDesError "descends directly from DurableExecutionsError, not from ExecutionError" — but that was already its base class in 1.x; the hierarchy did not change. What changed is the wrapping in serdes.py: 1.x's serialize/deserialize helpers caught everything and raised
   ExecutionError, while 2.0 raises SerDesError. The except ExecutionError → except SerDesError migration is correct; only the explanation misleads. (This misconception originated in the review thread, not the author's draft.)

   Worth doing while you're here

   Two SDK docstrings now contradict the guide, and the guide is right in both cases:

   - MapConfig.completion_config still says "Default allows any number of failures." A bare CompletionConfig() fail-fast. This is the docstring that would make a reader think the guide's fail-fast claim is wrong.
   - DurableOperationError's class docstring says "the escaping error is preserved as __cause__ on the first run and reconstructed on replay." ErrorObject.raise_as_operation_error sets __cause__ to a DurableOperationError rebuilt from the wire fields on both paths — deliberately, for replay determinism. The guide's "__cause__ does not hold the original
   exception" is the accurate statement.

   Since the guide's credibility depends on those docstrings agreeing, fixing them here is cheap.

yaythomas:
Thank you for the thorough check, all four confirmed against the sdk-v1.7.0 tag and fixed in 25de7cc, along with both docstrings:

- all_completed(): section rewritten per your reframing. The factory change is the breaking change, the default is unchanged, and 1.x all_completed() users are told to switch to all_successful() or a bare CompletionConfig() to keep fail-fast. Claim dropped from the intro and porting summary; replay identity validation takes the headline slot.
- should_complete timing: now "on every branch state change: scheduling, completion, failure, and suspension".
- Plugin link now points at /observability/plugins/.
- SerDesError: explanation rewritten around the raise-site change (1.x wrapped in ExecutionError); you are right the hierarchy never changed, and right about where the misconception started.
- MapConfig.completion_config and DurableOperationError docstrings fixed in the same commit.

docs/migration-1.x-to-2.0.md
```diff
@@ -0,0 +1,451 @@
...
+
+Catch `WaitForConditionError` instead of inspecting the returned state.
+
+### Replay validates operation identity
+
+On replay, the SDK validates each operation's checkpoint against the current
+code by `type`, `sub_type`, `name`, and `parent_id`. Any mismatch raises
+`NonDeterministicExecutionError` and fails the execution. `1.x` let a renamed or
+reordered operation consume a neighboring checkpoint silently. `2.0` fails fast.
+A `map` or `parallel` batch also validates FLAT against NESTED nesting drift.
```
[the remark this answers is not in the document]
yaythomas: Addressed in 25de7cc. The section now states validation catches drift in the four compared fields, cannot catch swaps of identical identities, and narrows rather than closes the silent-mismatch window; the deployment guidance prohibits reordering under in-flight executions regardless.

docs/migration-1.x-to-2.0.md
```diff
@@ -0,0 +1,451 @@
...
+currently treats an empty-string checkpoint payload as no stored polling state
+on resume, so the operation can restart from `initial_state`.
+
+`RetryableSerDesError` does more than retry serialization. An executor re-raises
+it before it writes a success checkpoint, so the backend re-invokes the whole
+execution. An `AT_LEAST_ONCE_PER_RETRY` step re-runs its body and repeats its
+side effects on that re-invocation, so raise `RetryableSerDesError` only from an
+idempotent step or accept the duplicate work. An `AT_MOST_ONCE_PER_RETRY` step
+already wrote its START checkpoint, so the retried invocation treats the step as
+interrupted and applies the step retry strategy instead of re-running the body.
```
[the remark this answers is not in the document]
yaythomas: Addressed in 25de7cc. The paragraph now distinguishes phases: pre-SUCCEED first-run failures re-run an at-least-once step body, at-most-once follows interrupted-step handling, and a retryable error while reading an already-succeeded checkpoint retries only the deserialization.

ParidelPooya (approved):
Re-reviewed at 25de7cc against 1.x (`sdk-v1.7.0`, 075b65a). All four points are fixed correctly, and the two claims you added beyond my feedback also check out.

**Verified fixed**

- `all_completed()`: the section now says what the code says. 1.x's all-`None` config hit `if not has_any_completion_criteria: if failure_count > 0` in `concurrency/models.py`, and `should_continue()` returned `failure_count == 0`, so 1.x `all_completed()` did fail fast. Telling those callers to switch to `all_successful()` is the right advice.
- Default completion dropped from the intro and porting summary. Both defaults are unchanged between versions (`MapConfig` → bare `CompletionConfig()`, `ParallelConfig` → `all_successful()`), so replacing it with replay identity validation in the headline is the right call.
- `SerDesError`: correctly reframed as a raise-site change. 1.x `serdes.py` wrapped into `ExecutionError`; the hierarchy was always `DurableExecutionsError`.
- `should_complete` timing: matches `needs_snapshot_rebuild`, which is set on submit and on `COMPLETED`, `FAILED`, `SUSPENDED`, and `SUSPENDED_UNTIL`.
- Plugin link now points at the page that documents the hooks.
- Both SDK docstrings now agree with the guide.

**New claims, both correct**

- Identity validation can't catch same-identity swaps. Operation ids are positional (`create_id_for_step`), so two unnamed steps of the same type under the same parent produce identical `type`/`sub_type`/`name`/`parent_id` at the same ids and `validate_checkpoint` sees no mismatch. Good caveat to add.
- `RetryableSerDesError` phase split. In `step.py`, the SUCCEEDED branch of `check_result_status` deserializes with no `try`, so the error escapes `process()`, no checkpoint is written, and the next invocation re-reads the same SUCCEEDED checkpoint and retries only the deserialize. The body is not re-run.

Anchors all resolve; ruff clean on the three touched source files. LGTM.

ParidelPooya (approved):

Issue #656, opened 2026-08-19
[Bug]: FAILED invocation without ErrorObject is completed as SUCCEEDED by local emulator


zhongkechen:
### Expected Behavior

An invocation response with explicit `Status=FAILED` must result in a failed durable execution.

If an error object is required by the invocation contract, a `FAILED` response without one should be rejected or converted to a generic failure. It must never be converted into a successful execution.

### Actual Behavior

In `aws_durable_execution_sdk_python_testing.executor.Executor._validate_invocation_response_and_store()`, the `FAILED` branch passes the nullable response error to `_complete_workflow()`:

```python
case InvocationStatus.FAILED:
    self._complete_workflow(
        execution_arn, result=None, error=response.error
    )
```

`_complete_workflow()` then chooses the terminal status based on whether the error is present:

```python
if error is not None:
    self.fail_execution(execution_arn, error)
else:
    self.complete_execution(execution_arn, result)
```

Consequently, this invocation response:

```json
{
  "Status": "FAILED"
}
```

is recorded as `ExecutionSucceeded`.

This was observed through SAM Local with the Java SDK when `waitForCondition` exhausted its attempts inside a synchronous child context. The operation history was:

```text
StepFailed (WaitForCondition)
ContextFailed (RunInChildContext)
InvocationCompleted
ExecutionSucceeded
```

The producing Java SDK issue is aws/aws-durable-execution-sdk-java#634.

### Steps to Reproduce

The emulator can be tested directly with a handler/invoker that returns:

```python
{
    "Status": "FAILED"
}
```

Alternatively:

1. Run a Java durable handler through SAM Local.
2. Inside synchronous `runInChildContext`, execute `waitForCondition` with a built-in wait strategy and a condition that always continues polling.
3. Allow the operation to exhaust its maximum attempts.
4. Inspect the durable execution history.

The step and child context are recorded as failed, but the execution is recorded as succeeded.

### SDK Version

`aws-durable-execution-sdk-python-testing` current `main` as of 2026-08-19. The affected code is used by the durable execution emulator image run by SAM Local.

### Python Version

3.13

### Is this a regression?

No known working version.

### Last Working Version

N/A

### Proposed Fix

Make the explicit invocation status authoritative:

```python
case InvocationStatus.FAILED:
    error = response.error or create_generic_failure_error()
    self.fail_execution(execution_arn, error)
```

[...]

PR #682, opened 2026-08-25
fix(testing): handle workflow failure states correctly

hln33:
*Issue #, if available:*
#656 

*Description of changes:*
* Modify handling of failed invocations to authoratively call `self._fail_workflow()` instead of `self._complete_workflow()` so that executions fail despite not having an error payload
  * Confirmed against the real backend service that failed invocations without error payloads should result in a failed execution 
  * Prior to this change, failed invocations with no error payloads would result in a successful execution
* Add tests covering the following behavior (requested in #656):
  * A failed invocation results in a failed execution 
  * Caught Child context failures do not cause execution failure

*Testing:*
* Ran all tests in package via `hatch run test:all`
* Invoked SAM CLI with local emulator changes and a reproducing Java durable function. This confirms failed invocations with empty error payloads result in failed execution:
```
sam local invoke WaitForConditionChildFailureJavaFunction

...

Execution Summary:
=========================
ARN:      31d25a2e-2e37-4bea-b67e-02441a5fa80c/55c89648-9aca-4b8e-8147-f01658dabf0a                                                                           Name:     N/A                                                                  Duration: 6.58s
Status:   FAILED ❌
Input:    {}

Commands you can use next
=========================
[*] Get execution details: sam local execution get 31d25a2e-2e37-4bea-b67e-02441a5fa80c/55c89648-9aca-4b8e-8147-f01658dabf0a
[*] View execution history: sam local execution history 31d25a2e-2e37-4bea-b67e-02441a5fa80c/55c89648-9aca-4b8e-8147-f01658dabf0a
```

### DIFF
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
@@ -1285,9 +1285,7 @@ class Executor(ExecutionObserver):
                     )
                     raise InvalidParameterValueException(msg_failed_result)
                 logger.info("[%s] Execution failed", execution_arn)
-                self._complete_workflow(
-                    execution_arn, result=None, error=response.error
-                )
+                self._fail_workflow(execution_arn, response.error)
 
             case InvocationStatus.SUCCEEDED:
                 if response.error is not None:
@@ -1591,7 +1589,7 @@ class Executor(ExecutionObserver):
         else:
             self.complete_execution(execution_arn, result)
 
-    def _fail_workflow(self, execution_arn: str, error: ErrorObject):
+    def _fail_workflow(self, execution_arn: str, error: ErrorObject | None):
         """Fail workflow with terminal state validation."""
         execution = self._store.load(execution_arn)
 
@@ -1671,8 +1669,8 @@ class Executor(ExecutionObserver):
             raise IllegalStateException(msg)
         self._complete_events(execution_arn=execution_arn)
 
-    def fail_execution(self, execution_arn: str, error: ErrorObject) -> None:
-        """Fail execution with error (FAIL_WORKFLOW_EXECUTION decision)."""
+    def fail_execution(self, execution_arn: str, error: ErrorObject | None) -> None:
+        """Fail execution with optional error (FAIL_WORKFLOW_EXECUTION decision)."""
         logger.error("[%s] Completing execution with error: %s", execution_arn, error)
         execution: Execution = self._store.load(execution_arn=execution_arn)
         execution.complete_fail(error=error, now=self._clock.now())
@@ -1688,7 +1686,7 @@ class Executor(ExecutionObserver):
         """Complete execution successfully. Observer method triggered by notifier."""
         self.complete_execution(execution_arn, result)
 
-    def on_failed(self, execution_arn: str, error: ErrorObject) -> None:
+    def on_failed(self, execution_arn: str, error: ErrorObject | None) -> None:
         """Fail execution. Observer method triggered by notifier."""
         self.fail_execution(execution_arn, error)
 
--- a/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
+++ b/packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
@@ -45,6 +45,7 @@ from aws_durable_execution_sdk_python_testing.exceptions import (
     ResourceNotFoundException,
 )
 from aws_durable_execution_sdk_python_testing.executor import Executor
+from aws_durable_execution_sdk_python_testing.execution import ExecutionStatus
 from aws_durable_execution_sdk_python_testing.invoker import (
     InProcessInvoker,
     LambdaInvoker,
@@ -498,6 +499,7 @@ class DurableFunctionTestResult:
     operations: list[Operation]
     result: OperationPayload | None = None
     error: ErrorObject | None = None
+    execution_status: ExecutionStatus | None = None
 
     @classmethod
     def create(cls, execution: Execution) -> DurableFunctionTestResult:
@@ -513,12 +515,16 @@ class DurableFunctionTestResult:
         if execution.result is None:
             msg: str = "Execution result must exist to create test result."
             raise DurableFunctionsTestError(msg)
+        if execution.close_status is None:
+            msg_status: str = "Execution close status must exist to create test result."
+            raise DurableFunctionsTestError(msg_status)
 
         return cls(
             status=execution.result.status,
             operations=operations,
             result=execution.result.result,
             error=execution.result.error,
+            execution_status=execution.close_status,
         )
 
     @classmethod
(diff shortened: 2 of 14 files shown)


packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
```diff
@@ -1285,9 +1285,7 @@ def _validate_invocation_response_and_store(
                     )
                     raise InvalidParameterValueException(msg_failed_result)
                 logger.info("[%s] Execution failed", execution_arn)
-                self._complete_workflow(
-                    execution_arn, result=None, error=response.error
-                )
+                self._fail_workflow(execution_arn, response.error)
```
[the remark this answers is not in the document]
hln33: There is one small obstacle to doing that. There currently isn't anything in the public API of `DurableFunctionTestRunner` or `DurableFunctionTestResult` that exposes an execution's status in the emulator. The public runner result only exposes the returned durable **invocation** status/result, not **execution**.

We could use private internals, like `DurableFunctionTestRunner._executor`, to get execution information, but that feels hacky to me for an e2e style test.

If we want to add an e2e test for this case, then I think we should first add a small public runner accessor for execution details/status, then use that in the test.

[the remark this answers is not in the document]
bchampp: I feel like we should expose a method for the overall execution status on the test runner here. Do we have this in the TS testing library? Thinking something like `result.execution_status`?

bchampp (dismissed):

packages/aws-durable-execution-sdk-python-testing/tests/executor_test.py
```diff
@@ -310,6 +310,42 @@ def test_should_complete_workflow_with_error_when_invocation_fails(
...
+    )
+
+    executor._validate_invocation_response_and_store(  # noqa: SLF001
+        execution.durable_execution_arn, response, execution
+    )
+
+    stored = store.load(execution.durable_execution_arn)
+    assert stored.is_complete is True
+    assert stored.close_status is not None
+    assert stored.close_status.value == "FAILED"
```
yaythomas: you can compare enums directly

```
assert stored.close_status is ExecutionStatus.FAILED
```

packages/aws-durable-execution-sdk-python-testing/tests/executor_test.py
```diff
@@ -310,6 +310,42 @@ def test_should_complete_workflow_with_error_when_invocation_fails(
...
+    assert stored.result is not None
+    assert stored.result.status == InvocationStatus.FAILED
```
yaythomas: assert stored.result.status is InvocationStatus.FAILED

packages/aws-durable-execution-sdk-python-testing/tests/e2e/child_context_failure_test.py
```diff
@@ -0,0 +1,63 @@
...
+        return context.step(recovery_step("handled"))
+
+    with DurableFunctionTestRunner(handler=handler, execution_timeout=10) as runner:
+        result: DurableFunctionTestResult = runner.run(input="input str")
+
+    assert result.status is InvocationStatus.SUCCEEDED
+    assert result.result == json.dumps("handled")
+
+    child_op: ContextOperation = result.get_context("failing-child")
+    assert child_op.status == OperationStatus.FAILED
```
yaythomas: `child_op.status is OperationStatus.FAILED`

packages/aws-durable-execution-sdk-python-testing/tests/execution_test.py
```diff
@@ -507,6 +507,26 @@ def test_complete_fail():
...
+        execution_timeout_seconds=300,
+        execution_retention_period_days=7,
+        invocation_id="test-invocation-id",
+    )
+    execution = Execution("test-arn", start_input, [Mock()])
+
+    execution.complete_fail(None)
+
+    assert execution.is_complete is True
+    assert execution.result.status == InvocationStatus.FAILED
```
yaythomas: execution.result.status is InvocationStatus.FAILED

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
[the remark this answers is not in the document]
yaythomas: nice idea @bchampp :-) 

`result.execution_status` sourced from `close_status`

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
```diff
@@ -1688,7 +1686,7 @@ def on_completed(self, execution_arn: str, result: str | None = None) -> None:
         """Complete execution successfully. Observer method triggered by notifier."""
         self.complete_execution(execution_arn, result)
 
-    def on_failed(self, execution_arn: str, error: ErrorObject) -> None:
+    def on_failed(self, execution_arn: str, error: ErrorObject | None) -> None:
```
yaythomas, line 1689: This type widening is right, and it exposes a pre-existing emulator divergence... Common typing W.

The checkpoint EXECUTION FAIL processor (checkpoint/processors/execution.py) synthesizes a generic ErrorObject when the update has no error, but it should instead preserve null there.

Proposal: remove the synthetic fallback, and widen ExecutionNotifier.notify_failed + ExecutionObserver.on_failed to ErrorObject | None to match the widened null signature.

yaythomas (changes requested):
Very nice, thank you! just some nits.

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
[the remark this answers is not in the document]
zhongkechen: Java runners have this https://github.com/aws/aws-durable-execution-sdk-java/blob/ba4bd48abcb7cf92e5c8181afb234e92618fc2fb/sdk-testing/src/main/java/software/amazon/lambda/durable/testing/TestResult.java#L57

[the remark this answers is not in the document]
hln33: Addressed in [e33de8b](https://github.com/aws/aws-durable-execution-sdk-python/pull/682/commits/e33de8b5c027d9bd8da3efdb4892862bb23819f3)

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/executor.py
hln33: Addressed in [bbdc1fe](https://github.com/aws/aws-durable-execution-sdk-python/commit/bbdc1fe96a59ef2f6ee45f7549bafbd8e0d73572)

packages/aws-durable-execution-sdk-python-testing/src/aws_durable_execution_sdk_python_testing/runner.py
```diff
@@ -1185,7 +1202,7 @@ def _wait_for_completion(
             if execution.status == "FAILED":
                 logger.warning("Execution failed")
                 return execution
-            if execution.status in ["TIMED_OUT", "ABORTED"]:
```
hln33, line 1188: I do not think "ABORTED" is a valid execution status. Could not find any references to it in other SDKs. The [API documentation](https://docs.aws.amazon.com/lambda/latest/api/API_GetDurableExecution.html#API_GetDurableExecution_ResponseSyntax) also does not list "ABORTED" as being a valid execution status. 

I think it should instead be "STOPPED", as that specific status is not covered in this function.

I noticed this when implementing a new "execution_status" field on the test result class and my AI coding agent flagged this to me.

yaythomas (approved):
:shipit:

friendly reminder to squash at merge with a conventional-commits message (lint-commits is currently red only because of the squash: markers).

PR #688, opened 2026-08-29
feat(otel): add durable sampling coordination

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/674

*Description of changes:*
- Second PR in the stack for reparenting durable OTel spans onto one shared execution trace. 
- Adds the sampling machinery that lets a single decision be resolved once per invocation and applied to every durable span. Not yet wired into the plugins, so behavior is unchanged.
- Next in stack: wire both plugins onto the shared execution trace and sampling.

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/durable_sampling.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/durable_sampling.py
@@ -1,0 +1,212 @@
+"""Durable execution sampling support."""
+
+from __future__ import annotations
+
+import functools
+import inspect
+from dataclasses import dataclass
+from typing import Any
+
+from opentelemetry import context as otel_context
+from opentelemetry.context import Context
+from opentelemetry.sdk.trace import Tracer as SdkTracer
+from opentelemetry.sdk.trace.sampling import Decision, Sampler, SamplingResult
+from opentelemetry.trace import Span, SpanContext, SpanKind, TraceFlags
+
+from aws_durable_execution_sdk_python_otel.context_extractors import (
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/durable_sampling.py
```diff
@@ -0,0 +1,212 @@
...
+    sampling = extracted.sampling if extracted is not None else Sampling.UNDECIDED
+    if sampling is Sampling.SAMPLED:
+        return SamplingResult(
+            Decision.RECORD_AND_SAMPLE,
+            trace_state=ambient_trace_state,
+        )
+    if sampling is Sampling.NOT_SAMPLED:
+        return SamplingResult(Decision.DROP, trace_state=ambient_trace_state)
+
+    if on_canonical_trace:
```
zhongkechen, line 121: Is this still true in case of synthetic span?

ayushiahjolia: Yes, when the backend provides no Root, sampling is effectively based on the synthetic root's trace, not the ambient span. `on_canonical_trace` evaluates to False (the ambient span is on a different trace than the ARN-derived canonical_trace_id), so the ambient-mirroring branch is skipped and its tracestate isn't captured; the decision falls to the configured sampler, evaluated against the synthetic canonical_trace_id on a fresh root context. That single decision is then applied to the synthetic root's SAMPLED flag and to every durable span, so they're all consistent.

zhongkechen (approved):

Issue #641, opened 2026-08-14
[Feature]: Reject new durable operations under orphaned map/parallel branches


zhongkechen:
## What would you like?

Reject newly created durable operations whose parent context has already been
marked orphaned after an early-completing `map` or `parallel` operation.

Today, `_mark_orphans()` marks the descendants known when the parent context
completes. If an in-flight orphaned branch subsequently creates a new durable
operation, `ExecutionState.create_checkpoint()` adds the new operation to
`_parent_to_children` and checks only whether the new operation's own ID is in
`_parent_done`.

Because the new operation did not exist when `_mark_orphans()` took its
snapshot, its ID is not in `_parent_done`, so its checkpoint can be accepted
even though its `parent_id` is already orphaned.

The orphaned branch's own terminal result is normally rejected later, and the
parent `BatchResult` remains consistent across replay:

- Normal payloads replay the parent's serialized result.
- `ReplayChildren` summaries preserve the branch as `STARTED` through
  `startedIndexes`.

This is therefore an improvement to lifecycle enforcement rather than a
customer-visible replay-result correctness bug. It would still prevent
unnecessary durable work, side effects, and history entries from being created
inside a branch whose parent has already completed.

This issue is separate from #640, which covers the race where an already-known
branch passes orphan validation before parent completion and enqueues its own
terminal result afterward.

## Possible Implementation

When admitting an operation checkpoint under `_parent_done_lock`, reject the
operation when either:

```python
operation_update.operation_id in self._parent_done
```

or:

```python
operation_update.parent_id in self._parent_done
```

If a parent can be indirectly orphaned without appearing directly in
`_parent_done`, propagate orphan state when registering the new child or walk
the known parent chain under the same lock.

The check should also be repeated in the atomic validation-and-enqueue section
proposed by #640 so parent completion cannot race queue insertion.

Suggested tests:

- A new step created under an already orphaned nested map iteration.
- A new invoke/callback/wait created under an orphaned parallel branch.
- The equivalent behavior for `NestingType.FLAT`.
- The new operation never reaches the checkpoint service.
- Parent results remain identical on first execution and replay for normal and
  `ReplayChildren` payloads.

## Is this a breaking change?

No. The operation is already outside the lifetime of its completed parent
context and its enclosing branch cannot contribute a terminal result to the
completed batch.

## Does this require an RFC?

No.

## Additional Context

A deterministic state-level reproduction shows:

```text
branch orphaned: True
[...]

PR #690, opened 2026-08-30
fix: reject orphan checkpoint race

### LINKED ISSUE #640: [Bug]: Orphaned map/parallel branch can checkpoint after parent completion
## Expected Behavior

When a `map` or `parallel` operation completes early because its completion
criteria are met, an in-flight branch marked as orphaned must not checkpoint
its terminal `SUCCEED` or `FAIL` result after the parent context completes.

Checkpoint admission should have a deterministic order:

- If the branch checkpoint is admitted first, it may be sent before the parent
  completion.
- If the parent completion marks the branch orphaned first, the branch
  checkpoint must be rejected and must not reach the checkpoint service.

## Actual Behavior

`ExecutionState.create_checkpoint()` validates `_parent_done` while holding
`_parent_done_lock`, releases that lock, invokes the plugin hook, and only later
enqueues the update while holding `_completion_lock`.

This creates a check-to-enqueue race:

1. An in-flight branch calls `create_checkpoint()` with its terminal context
   `SUCCEED` update and passes the `_parent_done` check.
2. Before the branch update is enqueued, the map/parallel parent checkpoints
   `SUCCEED` and `_mark_orphans()` marks the branch operation ID as orphaned.
3. The branch resumes and enqueues its terminal `SUCCEED` update because orphan
   status is not checked again.
4. `checkpoint_batches_forever()` sends every queued update without
   revalidating orphan status.

A deterministic reproduction recorded two successful checkpoint-service calls:

```text
service call 1: parent SUCCEED with BatchResult branch status STARTED
service call 2: branch SUCCEED with payload "branch-result"
```

Both synchronous checkpoint callers were released after the mock service
returned a new checkpoint token.

The customer-visible `BatchResult` remains replay-consistent in the current
implementation because normal payloads replay the parent's serialized result
and `ReplayChildren` summaries record `startedIndexes`. However, the durable
operation hierarchy can contain a terminal branch result checkpointed after
its parent context completed.

Relevant code:

- `packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py`,
  `ExecutionState.create_checkpoint()` around the `_parent_done_lock`
  validation and later `_checkpoint_queue.put()`
- `packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py`,
  `checkpoint_batches_forever()` where queued updates are sent without orphan
  revalidation

## Steps to Reproduce

1. Create an `ExecutionState` with a branch operation registered as a child of
   a map/parallel parent.
2. Start a branch terminal `CONTEXT/SUCCEED` checkpoint on a worker thread.
3. Pause that thread after its initial orphan validation but before queue
   insertion, for example using a blocking synchronous plugin hook.
4. Checkpoint the parent `CONTEXT/SUCCEED`, causing `_mark_orphans()` to mark
   the branch operation ID.
5. Allow the parent update to be sent successfully.
6. Release the branch thread.
7. Observe that the branch terminal update is sent in a subsequent checkpoint
   API call and the synchronous branch caller completes successfully.

## SDK Version

`1.8.0`, repository HEAD `7ac7acc6a7dae231f2abbb8e37f9780cc9b89af0`

## Python Version

3.14

## Is this a regression?

No known regression.

## Possible Fix

Make orphan validation and queue insertion atomic relative to parent
completion:

1. Inside the final `_completion_lock` block, acquire `_parent_done_lock` using
   a consistent global lock order.
2. Recheck whether `operation_update.operation_id` is in `_parent_done`
   immediately before queue insertion.
3. Optionally also reject when `operation_update.parent_id` is in
   `_parent_done`, preventing new durable descendants from being created under
   an already orphaned branch.
4. Insert the queued operation while still holding `_parent_done_lock`.

This produces deterministic ordering: either the child is queued first, or
parent completion wins and the child is rejected.

Add deterministic unit tests for:

- A terminal branch checkpoint paused between initial validation and enqueue.
- A new operation created under an already orphaned branch.
- The valid child-first ordering.
- Replay consistency for normal and `ReplayChildren` map/parallel results.

## Additional Context

The existing core test suite passes (`1504 passed`, plus 5 subtests), but it
does not currently cover this interleaving.

warren830:
## Issue

Fixes #640

## Description

A child checkpoint could pass the initial parent-done validation, pause while plugin hooks ran, then enqueue after its parent completed and marked it orphaned. The final enqueue path rechecked execution completion and checkpoint failure, but not the child's `_parent_done` state.

This change:

- extracts the repeated orphan rejection into `_reject_if_parent_done()`;
- rechecks the child's orphan state in the final validation-and-enqueue region;
- holds `_completion_lock` and then `_parent_done_lock` through queue insertion, establishing a single linearization point;
- documents the global `_completion_lock` → `_parent_done_lock` order;
- keeps plugin callbacks outside both locks and avoids the parent-done lock for empty checkpoints.

## Scope

This PR fixes only the already-known-child race in #640. It deliberately does not reject late-created descendants whose parent was already orphaned; that separate behavior remains tracked by #641.

## Red → green

- Unchanged upstream production code: the deterministic regression failed with `DID NOT RAISE OrphanedChildException`.
- Patched focused test: passed.
- Repeated fresh executions: 25/25 passed.

The regression uses two `threading.Event` barriers and no sleeps: the child pauses in the plugin hook after initial validation, the parent completes and marks it orphaned, and the child then resumes into the final enqueue attempt.

## Testing

- Repository tests: 3351 passed, 2 skipped, 5 subtests.
- Core tests: 1627 passed, 5 subtests.
- Core coverage: 98.09% (98% required).
- OpenTelemetry tests: 149 passed, 92% coverage.
- Examples: 87 passed, 2 skipped.
- Type checks: 77 core source files and 23 OpenTelemetry source files, no errors.
- Ruff format/check: core (77 files), OpenTelemetry (23 files), and examples (155 files), all passed.
- Commit lint and `git diff --check`: passed.

All validation used the repository-pinned Hatch 1.16.5 and Python 3.12. No AWS call or cloud resource was used.

## Fork CI note

Some repository cloud/conformance workflows require AWS secrets, SAM, and live CloudFormation resources that are unavailable to fork PRs. The authoritative local unit, integration, coverage, type-check, and formatting gates above are green; any credential-gated non-required fork check should be treated as an infrastructure limitation rather than a product regression.

## License

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/state.py
@@ -308,7 +308,8 @@ class ExecutionState:
         # Operations whose parent has completed
         self._parent_done: set[str] = set()
 
-        # Protects parent_to_children and parent_done
+        # Protects parent_to_children and parent_done. When both state locks are
+        # required, acquire _completion_lock before _parent_done_lock.
         self._parent_done_lock: Lock = Lock()
 
         # Branch thread pools created by concurrency coordinators. A pool
@@ -538,6 +539,23 @@ class ExecutionState:
             operation_id=operation_id,
         )
 
+    def _reject_if_parent_done(self, operation_update: OperationUpdate) -> None:
+        """Raise OrphanedChildException when the operation is orphaned.
+
+        Must be called while holding _parent_done_lock so the check can be
+        linearized with parent completion and, at the final call site, enqueue.
+        """
+        if operation_update.operation_id not in self._parent_done:
+            return
+        logger.debug(
+            "Rejecting checkpoint for operation %s - parent is done",
+            operation_update.operation_id,
+        )
+        raise OrphanedChildException(
+            "Parent context completed, child operation cannot checkpoint",
+            operation_id=operation_update.operation_id,
+        )
+
     def create_checkpoint(
         self,
         operation_update: OperationUpdate | None = None,
@@ -631,18 +649,7 @@ class ExecutionState:
                     self._mark_orphans(operation_update.operation_id)
 
                 # Check if this operation's parent is done
-                if operation_update.operation_id in self._parent_done:
-                    logger.debug(
-                        "Rejecting checkpoint for operation %s - parent is done",
-                        operation_update.operation_id,
-                    )
-                    error_msg = (
-                        "Parent context completed, child operation cannot checkpoint"
-                    )
-                    raise OrphanedChildException(
-                        error_msg,
-                        operation_id=operation_update.operation_id,
-                    )
+                self._reject_if_parent_done(operation_update)
 
         # Check if background checkpointing has failed
         if self._checkpointing_failed.is_set():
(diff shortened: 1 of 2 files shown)


zhongkechen (approved):
Thank you for the contribution! The fix cleanly closes the checkpoint admission race, establishes a clear lock order, and adds deterministic regression coverage. I also verified the full core suite passes. Approved.

PR #693, opened 2026-08-31
feat(otel): parent durable spans to shared trace

ayushiahjolia:
*Issue #, if available:* https://github.com/aws/aws-durable-execution-sdk-python/issues/674

*Description of changes:*
- Third PR in the stack for reparenting durable OTel spans onto one shared execution trace.
- Wires both plugins (InvocationOtelPlugin and ExecutionOtelPlugin) onto the shared execution trace and sampling. Workflow and Invocation spans now share one trace, anchored to a propagated backend parent when present or a deterministic synthetic execution root otherwise; an ambient Lambda span is used as parent only when it already belongs to the execution trace. 
- Sampling is resolved once per invocation and applied to every durable span. Also completes the context_extractors rewrite deferred from the first PR: extractors now return a structured ExtractedContext instead of an opaque OTel Context.
- Next in stack: docs (README) update for the shared-trace model.

### DIFF
--- a/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
+++ b/packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
@@ -5,11 +5,11 @@ The :class:`ExecutionOtelPlugin` produces the deterministic span hierarchy
     Workflow -> Operation -> Attempt
 
 that stitches a single trace across every Lambda invocation of one durable
-execution. The Workflow span is the root (created in an empty context so it
-never has a parent) and is exported exactly once, when the execution reaches a
-terminal status. Operations are parented under the Workflow span (or their
-parent operation) and *linked* to the current Invocation span. The Invocation
-span belongs to the ambient Lambda trace instead of the Workflow trace.
+execution. Workflow and Invocation spans parent onto the same execution
+ancestor: a propagated backend parent when present, otherwise a deterministic
+synthetic root. The Workflow span is exported exactly once, when the execution
+reaches a terminal status. Operations are parented under the Workflow span (or
+their parent operation) and *linked* to the current Invocation span.
 
 This is the Python adaptation of the JS ``ExecutionOtelPlugin`` from
 aws-durable-execution-sdk-js#729. Because the Python plugin interface differs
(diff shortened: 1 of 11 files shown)


packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
```diff
@@ -298,14 +324,52 @@ def _resolve_parent_span(self, parent_id: str | None = None) -> Span:
...
+        execution_trace_context = self._execution_trace_context
+        if execution_trace_context is None:
+            return self._with_sampling(Context())
+
+        ambient_span = trace.get_current_span()
+        ambient_context = ambient_span.get_span_context()
+        if (
+            ambient_context.is_valid
+            and ambient_context.trace_id == execution_trace_context.trace_id
+        ):
```
zhongkechen, line 337: Would it still use ambient span in case of synthetic span? Is this expected?

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/execution_plugin.py
```diff
@@ -274,14 +293,26 @@ def _resolve_parent(self, parent_id: str | None) -> Span | None:
...
+        """Return same-trace ambient context, else execution ancestor context."""
+        execution_trace_context = self._execution_trace_context
+        if execution_trace_context is None:
+            return self._with_sampling(Context())
+
+        ambient_span = trace.get_current_span()
+        ambient_context = ambient_span.get_span_context()
+        if (
+            ambient_context.is_valid
+            and ambient_context.trace_id == execution_trace_context.trace_id
```
zhongkechen, line 305: Same question here: Would it still use ambient span in case of synthetic span? Is this expected?

packages/aws-durable-execution-sdk-python-otel/src/aws_durable_execution_sdk_python_otel/invocation_plugin.py
ayushiahjolia: No - a synthetic root's trace id is ARN-derived, so an unrelated ambient span won't match trace_id and the branch is skipped; it parents to the synthetic ancestor instead. The ambient span is used only when it's already on the execution trace (backend-propagated Root). 

Covered by test_*_ignores_different_trace_ambient_span (synthetic → ignored) and test_invocation_span_parents_to_same_trace_ambient_span (backend → used).

zhongkechen (dismissed):

zhongkechen (approved):

PR #717, opened 2026-09-10
fix(core): detect bundled SDK via version stamp

nvasiu:
*Issue #, if available:*

N/A

*Description of changes:*

The managed runtime now stamps a `+bundled` local version label onto the installed durable SDK. This update reads the stamp to determine if the SDK is installed in the managed runtime (replacing the old method of checking the installation folder, which was unreliable).


packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/lambda_service.py
```diff
@@ -1218,7 +1214,7 @@ def initialize_client(cls) -> LambdaClient:
                 config=Config(
                     connect_timeout=5,
                     read_timeout=50,
-                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__}{'-bundled' if _is_in_var_dir() else ''}",
+                    user_agent_extra=f"aws-durable-execution-sdk-python/{__version__.removesuffix('+bundled')}{'-bundled' if _is_bundled(__version__) else ''}",
```
yaythomas: the literal `"+bundled"` now appears twice (here and in `_is_bundled`), and this f-string does three things on one line.

I'd pull the mapping into one small helper next to `_is_bundled`, so there's one place that owns the label and one place to test the mapping directly:

```
_BUNDLED_LOCAL_LABEL: str = "+bundled"


def _is_bundled(version: str) -> bool:
    """True if the managed Lambda runtime appended the ``+bundled`` local label to this install."""
    return version.endswith(_BUNDLED_LOCAL_LABEL)


def _user_agent_version(version: str) -> str:
    """Return the version segment reported in the boto3 user agent.

    The managed runtime appends ``+bundled`` to ``__version__`` at image build
    time. The user agent reports that install as ``<public version>-bundled``.
    The ``-bundled`` form predates the label and matches the JS SDK, so existing
    telemetry queries keep working. Any other local label is reported unchanged.
    """
    if _is_bundled(version):
        return f"{version.removesuffix(_BUNDLED_LOCAL_LABEL)}-bundled"
    return version
```

and here:

```
    user_agent_extra=f"aws-durable-execution-sdk-python/{_user_agent_version(__version__)}",
```

Reason: the two `__version__`-patched tests then become a plain parametrized test on `_user_agent_version`, with no boto3 mock needed. Not blocking.

packages/aws-durable-execution-sdk-python/tests/lambda_service_test.py
```diff
@@ -2226,14 +2226,14 @@ def test_lambda_client_initialize_client_no_endpoint(
 
 
 @patch(
-    "aws_durable_execution_sdk_python.lambda_service._is_in_var_dir",
+    "aws_durable_execution_sdk_python.lambda_service._is_bundled",
```
yaythomas: these two patch `_is_bundled` while the real `__version__` (no label) flows through `removesuffix`, and the two new tests below patch `__version__` and exercise the real `_is_bundled`. The new pair covers everything the old pair does. I'd keep the new pair and drop these two, or keep one of these purely as a "boto3 wiring" check. Also `__version__.split('+')[0]` in the assertions doesn't match the source's `removesuffix` behaviour; harmless today because `__about__.py` never carries a `+`, but it would diverge the day it does.

nvasiu: Updated, kept one test to check the boto3 wiring (and simplified it).

Issue #716, opened 2026-09-09
Improve the monorepo release process


yaythomas:
### Problem

Python has no native tooling for releasing multiple packages from one monorepo. There is no equivalent of npm workspaces with per-package publish orchestration. This repository works around the gap with a GitHub Actions workflow, `pypi-publish.yml`. The workflow builds and publishes every package on every GitHub release. This design has three problems.

**1. The release tag does not control what gets published.**

The workflow triggers on `release: published`. It uses the tag only as a checkout ref. No filter connects the tag name to the package matrix. So a release tagged `sdk-v2.0.0` builds and uploads all four packages: `aws-durable-execution-sdk-python`, `-otel`, `-testing`, and `-insight`.

The real selector is each package's `__about__.py` version. A package publishes when its version is absent from PyPI. A package fails when its version is already there. So the set of published packages depends on which version bumps sit on the tagged commit. It does not depend on what the tag says. A maintainer who intends to release only the SDK can unintentionally publish other packages.

One workflow does read the tag. `lambda-layer-publish.yml` gates on `contains(tag_name, 'otel-v')`. That gate forced the combined tag convention `sdk-v1.7.0,otel-v0.3.0`. The convention packs several package names into one git tag with commas.

**2. Already-published versions fail the workflow instead of being skipped.**

The publish step does not check PyPI before uploading. It does not set `skip-existing`. A package with an unchanged version is rebuilt and re-uploaded. PyPI rejects the upload with `400: Uploading new files to releases older than 14 days is not allowed`. In practice the `-testing` package hits this on most releases. So every release since the monorepo conversion has finished with a failed workflow run. This holds even when the intended packages published successfully. A release workflow that is always red trains maintainers to ignore failures. Ignored failures hide real ones.

**3. There is no pre-flight plan and no post-publish verification.**

Nothing reports which packages will publish before uploads begin. Nothing confirms afterward that PyPI serves the new versions. The first signal of a wrong or missing publish is a user report, or a red job found by reading logs.

### How the JS repository avoids this

[aws-durable-execution-sdk-js](https://github.com/aws/aws-durable-execution-sdk-js) has the same outer shape. A release event triggers a workflow that iterates all publishable packages. The difference is its publish script, `iterate-publish-npm.sh`. The script runs in two phases.

[...]

PR #718, opened 2026-09-10
ci: add gate, preflight and verification to testing image release

nvasiu:
*Issue #, if available:*

Related to https://github.com/aws/aws-durable-execution-sdk-python/issues/716

*Description of changes:*

The `ecr-release.yml` action would previously run for every monorepo release, regardless if the release included a new version for the testing package or not. So it was possible for this action to publish the latest changes of the testing package before they were released. This PR updates the action to be safer.

`.github/workflows/ecr-release.yml`
- Add a preflight job to parse the release tag, verify the tag matches the source, check if the version already exists in public ECR, and emit a plan.
- Gates the build on the above preflight checks.
- Add a post-publish verification that polls public ECR (polls 10 times with 15s wait = 150s total) and confirms that the new version was published.

`.github/scripts/parse_testing_version.py`
- Separate script for parsing the testing version from a release tag.

`.github/scripts/tests/test_parse_testing_version.py`
- Unit testing for the parsing script.

`.github/workflows/test-parser.yml`
- Wire the new script and test into the script-test workflow.

### DIFF
--- a/.github/workflows/ecr-release.yml
+++ b/.github/workflows/ecr-release.yml
@@ -14,6 +14,8 @@ env:
 
 jobs:
   build-and-upload-image-to-ecr:
+    # Only publish when the release includes a new testing package version.
+    if: contains(github.event.release.tag_name, 'testing-v')
     runs-on: ubuntu-latest
     permissions:
       contents: read


yaythomas (commented):
The preflight and verification confirm that an image tag exists. They do not confirm the image contains what the release names. The Dockerfile installs `aws-durable-execution-sdk-python>=1.0.0`, resolved at build time and unbounded, so two builds of one testing version can produce different images, and skip-if-exists assumes a version identifies one artifact.

(The other three points are on the relevant lines.)

.github/workflows/ecr-release.yml
```diff
@@ -13,7 +13,106 @@ env:
...
+          ECR_REPOSITORY: ${{ env.ecr_repository_name }}
+        run: |
+          repo_name="${ECR_REPOSITORY##*/}"
+          if aws ecr-public describe-images \
+              --region "${{ env.aws_region }}" \
+              --repository-name "$repo_name" \
+              --image-ids imageTag="v${VERSION}" >/dev/null 2>&1; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
```
yaythomas: Every nonzero exit becomes `exists=false`, which means publish. A missing IAM permission, a throttle, a network error, and a genuinely absent tag all result in publish here?

.github/scripts/parse_testing_version.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+
+import os
+import re
+
+# Matches a testing-v<x.y.z> entry in the release tag, stopping at a comma
+# so combined tags like "sdk-v2.0.0,testing-v2.0.0" resolve to the testing part.
+_PATTERN = re.compile(r"testing-v[0-9]+\.[0-9]+\.[0-9]+([^,]*)")
```
yaythomas: `resolve_layer_sdk_version.py` already parses release tags, using `split(",")` plus `startswith("sdk-v")`. `parse_testing_version.py` uses an unanchored regex with a `([^,]*)` suffix capture. The logic is also not equivalent.

.github/workflows/ecr-release.yml
```diff
@@ -132,3 +224,53 @@ jobs:
...
+            exit 1
+          fi
+
+          latest_digest="$(aws ecr-public describe-images \
+              --region "${{ env.aws_region }}" \
+              --repository-name "$repo_name" \
+              --image-ids imageTag="latest" \
+              --query 'imageDetails[0].imageDigest' --output text 2>/dev/null || true)"
+          if [[ "$latest_digest" != "$version_digest" ]]; then
+            echo "::error::latest on public ECR points at $latest_digest but v$VERSION is $version_digest."
```
yaythomas: `verify-publish` only confirms `latest` points at the version just pushed, not that it points at the newest version, so a backport release like `testing-v1.1.3` cut after 1.2.1 would move `latest` backwards and still pass.

nvasiu:
@yaythomas 
Re: https://github.com/aws/aws-durable-execution-sdk-python/pull/718#pullrequestreview-5182537123

Yes, the preflight confirms that a tag exists, but it's not guaranteed that the image matches the release. And `skip-if-exists` assumes a version maps to one image, which is not true while we have an unbounded SDK dependency.

But this workflow's current policy is that if a version is already published to ECR, we won't ever overwrite it. So it shouldn't ever be a concern that the user will get different SDK versions from the same image version.

But for the sake of reproducing the image on the ECR in the future / visiblity into what SDK version the image is using, we could:
- Pin the SDK dependency in the Dockerfile.
- Or label images with which SDK version they are using (doesn't prevent different images per version, but lets us detect them).

I think either of these option would need some more discussion, and are out of scope for this PR. But if we want to implement either of those, they would be compatible with this PR.

PR #723, opened 2026-09-13
fix: match invalid checkpoint token casing

vyrnsynx:
*Issue #, if available:*: #721

*Description of changes:*

The service emits `Invalid checkpoint token` (lowercase). The SDK matched a Title Case prefix, so a stale checkpoint token was classified as a non-retryable execution failure instead of a retryable invocation failure.

- Compare the message prefix case-insensitively
- Cover both the service casing and the previous Title Case string in unit tests

*Testing:*

- `hatch run dev-core:test`: 1670 passed
- `hatch fmt --check` in the core package: passed

### DIFF
--- a/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
+++ b/packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
@@ -13,7 +13,7 @@ BAD_REQUEST_ERROR: int = 400
 TOO_MANY_REQUESTS_ERROR: int = 429
 SERVICE_ERROR: int = 500
 INVALID_PARAMETER_VALUE_EXCEPTION: str = "InvalidParameterValueException"
-INVALID_CHECKPOINT_TOKEN_PREFIX: str = "Invalid Checkpoint Token"
+INVALID_CHECKPOINT_TOKEN_PREFIX: str = "invalid checkpoint token"
 
 # Non-retryable customer error codes that arrive as non-4xx (e.g. HTTP 502) from Lambda.
 # Unlike typical 5xx errors, these require customer intervention (e.g., fixing
@@ -162,7 +162,7 @@ class BotoClientError(InvocationError):
           These arrive as HTTP 502 but require customer intervention to fix.
         - 4xx errors → EXECUTION, except:
           - 429 (TooManyRequests) → INVOCATION (throttling is transient)
-          - InvalidParameterValueException with "Invalid Checkpoint Token" → INVOCATION
+          - InvalidParameterValueException with "Invalid checkpoint token" (case-insensitive) → INVOCATION
             (stale token from a concurrent checkpoint; next invocation gets a fresh token)
         - 5xx, network errors → INVOCATION
         """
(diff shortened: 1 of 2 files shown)


packages/aws-durable-execution-sdk-python/tests/exceptions_test.py
```diff
@@ -65,12 +65,19 @@ def test_checkpoint_error():
...
-    """Test 4xx InvalidParameterValueException with Invalid Checkpoint Token is invocation error."""
+@pytest.mark.parametrize(
+    "message",
+    [
+        "Invalid checkpoint token: token expired",
+        "Invalid Checkpoint Token: token expired",
+    ],
+)
+def test_checkpoint_error_classification_invalid_token_invocation(message: str):
+    """Service emits lowercase 'checkpoint token'; match case-insensitively as invocation."""
```
hln33: I would prefer the original description for this test. 

I feel that it more clearly conveys what the purpose of the test is. The original also follows the format that most of the test descriptions in this file use.

Maybe something like this is better:
```suggestion
    """Test 4xx InvalidParameterValueException with "Invalid checkpoint token" error message is an invocation error."""
```

hln33:
Thank you for the contribution! 

Overall approach looks good to me - just some small requested changes. 

Also, I ran the CI tests for this PR and it looks like some checks are failing. Can you please have a look and address that as well? [Example failure](https://github.com/aws/aws-durable-execution-sdk-python/actions/runs/34727451056/job/104165613827?pr=723).

hln33 (changes requested):

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
```diff
@@ -180,9 +180,9 @@ def _classify_error_category(
             and error
             and not (
                 (error.get("Code") or "") == INVALID_PARAMETER_VALUE_EXCEPTION
-                and (error.get("Message") or "").startswith(
-                    INVALID_CHECKPOINT_TOKEN_PREFIX
-                )
+                and (error.get("Message") or "")
+                .casefold()
```
hln33: let's perform an exact string match instead of doing a lowercase/casefold match. The main purpose is to match the backend exactly. If the error message does not exactly match we wouldn't want this condition to be true.

Can you please make this comparison case-sensitive and amend the other changes in this PR to reflect that?

hln33 (changes requested):

vyrnsynx:
Updated per review: switched to an exact case-sensitive match on `Invalid checkpoint token`, and restored the original-style test description for the single backend-casing case.

packages/aws-durable-execution-sdk-python/src/aws_durable_execution_sdk_python/exceptions.py
vyrnsynx: Switched to an exact case-sensitive `.startswith("Invalid checkpoint token")` match (no casefold) so we only treat the backend’s exact message as invocation.

packages/aws-durable-execution-sdk-python/tests/exceptions_test.py
vyrnsynx: Restored the original-style test description and dropped the parametrize so we only cover the backend casing.

hln33 (approved):

hln33:
LGTM!

Integration/Conformance tests are failing due to missing credentials, but that is expected for external PRs as we only grant `secrets` to internal contributors. All other CI checks are passing.

Thanks once again for your contribution!

PR #734, opened 2026-09-17
fix(insight): key export scheduling by execution ARN

ParidelPooya:
## Problem

One plugin instance serves every execution its environment hosts, and Lambda Managed Instances makes concurrent executions in one environment routine.

`_ExportScheduler` held one `_pending` record for the whole plugin and overwrote it regardless of which execution the record belonged to. Each record is a complete snapshot of *one* execution, so a newer record for the same execution supersedes the older one safely — but a record for a *different* execution supersedes nothing.

Measured before the change, 20 trials per row:

| concurrent executions | terminal records lost per trial |
|---|---|
| 2 | 0.8 of 2 (16 of 20 trials lost one) |
| 3 | 1.9 of 3 |
| 5 | 3.9 of 5 |
| 10 | 8.5 of 10 |

With 5 concurrent executions and a 200 ms exporter, 1 of 5 terminal records was exported and every `drain()` still returned without error, so the loss was silent.

## What changed

Pending records are keyed by execution ARN, with a per-execution lane holding that execution's sequence numbers and waiter count. Coalescing happens only within one execution, and `drain(execution_arn)` returns only once that execution's own record has reached every exporter *and* a flush covering it has completed. One worker thread and one `export()` at a time are unchanged, so an exporter never sees concurrent calls.

Five further defects, all found while reviewing that change:

1. A `BaseException` from a customer exporter killed the worker between consuming a record and publishing its bookkeeping. The worker slot stayed occupied by a dying thread, `_ensure_worker_locked` refused to start a replacement, and the parked drain hung the invocation thread permanently. `asyncio.CancelledError` inherits from `BaseException`, so an exporter that touches asyncio can trigger this without writing `raise`.
2. A drain could request a second flush while one was already in flight, so a `flush()` ran after `drain()` had returned — an exporter called after the invocation went back to Lambda — and made the suite flaky in about 4% of runs.
3. The per-execution lock was not reentrant while a displaced record was released under it, so a record whose finalizer re-entered a hook for the same execution self-deadlocked the invocation thread.
4. The `closed` gate was a check-then-act. Customer code running under the now-reentrant lock (an input or output transform, a result override, or `__del__` on an object the record carries) could complete `on_invocation_end` on the same thread, after which the outer frame still scheduled its RUNNING record behind the terminal one. Reproduced deterministically from a single hook call.
5. A flush happened only when a record was emitted, where JS and Java flush once per sampled-in invocation end. A buffering exporter now sees the same rhythm in all three languages.

`InsightExporter.flush` also had no docstring; it now states the cadence, the exclusivity guarantee, that a flush may cover other executions' records, what happens when an exporter omits the method, and how failures are handled. The same wording is going into the JS and Java exporter interfaces.

## Verification

- Suite green with new permanent tests for every item above, run 20 consecutive times with no flaky runs. Mutation coverage of the scheduler and gate behaviour went from 4 of 12 targeted single-line mutations caught to 11 of 17.
- Terminal-record loss after the change: 0 in every configuration, up to 10 concurrent executions, at both exporter latencies, in both scheduler-level and plugin-level probes.
- `BaseException` from `export()` and from `flush()`: all eight type-and-site combinations now return instead of hanging.
- The emitted record surface is byte-identical to `main`: same fields, same emit-mode behaviour, same sampling, same truncation.
- `ruff check`, `ruff format --check` and `mypy` clean on every changed file.

## Known follow-ups, not in this PR

- A `BaseException` out of `exporter.flush()` still ends the worker mid-flush. The waiting invocation is released by a replacement worker, so it no longer hangs, but the remaining exporters in that flush are skipped.
- On a non-terminal invocation end in on-change mode this plugin omits the execution's `endTime`, `durationMs`, `output` and `error`, where JS and Java include them. Cross-language alignment is a separate change; the retry `error` in particular is information this plugin currently drops.
- The `_flush_barrier` is latency policy rather than a correctness guarantee and no test pins it.


ParidelPooya:
Superseded by #737, which carries this fix as its first commit plus the plugin-contract change that removes the reason the bug was possible. Closing in favour of one PR per language.

